PR #40, opened 2020-02-27
Fix xarray loading problem

mapazarr:
xarray was not updating the array loaded from file if the file was overwritten on disk and reloaded into memory.
The details are explained in this issue [https://github.com/pydata/xarray/issues/3785](https://github.com/pydata/xarray/issues/3785).

This PR updates the version o xarray used and changes the method used to read data from file from `open_dataarray` to `load_dataarray`. The former was keeping the file open, while the latter closes the file after loading the contents to memory.


ngreenwald:
Great, can you change this is the utils helper functions as well?

mapazarr:
Done, I also updated the evaluation file using the same method.

Issue #44, opened 2020-02-29
add support for MIBI-TIFF files


ngreenwald:
create a new load function that reads in a folder of MIBI TIFFs and produces an xarray with the same characteristics as existing load_tiffs_from_points_dir function

mapazarr:
This issue is addressed in PR #75.

PR #75, opened 2020-04-20
Add compatibility with MIBItiff input files.

mapazarr:
Add load function that take MIBItiffs and converts them to xarrays. Add smaple MIBItiff data and a notebook that shows how to perform the preprocessing using MIBItiff data.
This PR addresses issue #44.


mapazarr:
@ngreenwald I am aware of the failed tests in travis-ci. I will work on fixing them sometime this week.

mapazarr:
@ngreenwald I fixed the tests and added the new notebook to the list for testing. BTW, the training notebook is not on the list, is this intentional?

This PR is ready for a formal review.

ngreenwald:
Okay great, I'll take a look!

What training notebook?

ngreenwald:
Heh, it's been so long since I looked at that notebook I forgot it was
there. I'll remove it, we're switching to a new setup anyway.

segmentation/utils/data_utils.py
```diff
@@ -7,14 +7,60 @@
...
+        img_xr: xarray with shape [fovs, tifs, x_dim, y_dim]
+    """
+
+    # sanity check
+    if not np.isin(load_axis, ["fovs", "stacks"]):
+        raise ValueError("Invalid value for load_axis, must be one of [fovs, stacks]")
+
+    # if no channels specified, get them from first MIBItiff file
+    if channels is None:
+        channels = img_data[0].channels
```
ngreenwald, line 40: I don't think img_data has been defined yet, right?

segmentation/utils/data_utils_test.py
```diff
@@ -13,7 +14,7 @@
 
 
 def _create_img_dir(temp_dir, fovs, imgs, img_sub_folder="TIFs"):
-    tif = np.random.randint(0, 100, 1024**2).reshape((1024, 1024)).astype('int8')
+    tif = np.random.randint(0, 100, 1024**2).reshape((1024, 1024)).astype("int8")
```
ngreenwald, line 17: You don't like single quotes I see :)

segmentation/utils/data_utils_test.py
```diff
@@ -22,6 +23,22 @@ def _create_img_dir(temp_dir, fovs, imgs, img_sub_folder="TIFs"):
             io.imsave(os.path.join(fov_path, img), tif)
 
 
+def test_load_imgs_from_mibitiff():
+    mibitiff_files = [os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                                   "..", "..", "data", "example_dataset",
+                                   "input_data", "input_data_TIFF",
+                                   "Point8_RowNumber0_Depth_Profile0-MassCorrected-Filtered.tiff")]
+    channels = ["HH3", "Membrane"]
```
ngreenwald, line 31: Adding a channels=None test case should catch the error in the main function

segmentation/utils/data_utils.py
mapazarr: You're right. I fixed this and added a test for this case.

segmentation/utils/data_utils_test.py
mapazarr: I actually prefer single quotes to double quotes, but I saw that most of the quotes in the notebooks were double quotes, so I just turned all single quotes to double quotes for consistency. Let me know if you would like me to revert to the originally used quotes (a mix of single and double ones).

segmentation/utils/data_utils_test.py
mapazarr: Done.

mapazarr:
@ngreenwald I addressed your comments. The PR is ready for another review.

mapazarr:
> One final change: could you change the behavior to take in a directory and a list of file names, rather than a list of full paths?

The reason I didn't do that is because of the way MIBI/O organizes the files: it stores all MIBItiff files in the same folder, including:
- All tiffs for all points.
- For each point: the tiff after extraction and background subtraction, the tiff after isobaric corrections and the tiff after denoising and aggregate removal.
Normally you wouldn't want to use all those files for the segmentation, just the final files after denoising.
Having said that, it's true that the user can reorganize the files afterwards in a different manner and the option you mention can be useful. HOwever, it can be confusing for an unexperienced user.

Can you please let me know if you changed your mind after this explanation?

ngreenwald:
Got it, that makes sense. In that case, I think it makes the most sense to just wait and see how we end up organizing our files before deciding.

mapazarr:
@ngreenwald I had to fix a few conflicts, could you please review the PR again please?

PR #90, opened 2020-05-06
Spatial Analysis Test

Jaiveers21:


segmentation/utils/SpatialAnalysis_Test.py
```diff
@@ -0,0 +1,14 @@
...
+
+    test_mat = np.zeros((512, 512), dtype="int")
+    test_mat[0, 20] = 1
+    test_mat[4, 17] = 2
+
+    dist_matrix = Spatial_Analysis.calc_dist_matrix(test_mat)
+    real_mat = np.array([[0, 5], [5, 0]])
+    assert np.array_equal(dist_matrix, real_mat)
+
+test_calc_dist_matrix()
```
ngreenwald, line 14: You don't need to actually call the test function, pytest will automatically call it

segmentation/utils/Spatial_Analysis.py
```diff
@@ -0,0 +1,24 @@
...
+
+def calc_dist_matrix(label_map):
+    """Generate matrix of distances between center of pairs of cells
+        Args
+            label_map: numpy array with unique cells given unique pixel labels
+        Returns
+            dist_matrix: cells x cells matrix with the euclidian
+            distance between centers of corresponding cells"""
+
+    if len(np.unique(label_map)) < 3:
```
ngreenwald, line 15: This wasn't a very useful error check in the original function, and it fails if there's only two cells, so we can remove it and think of some better error checking later on

segmentation/utils/Spatial_Analysis.py
```diff
@@ -0,0 +1,24 @@
...
+        raise ValueError("Array must be provided in labeled format")
+
+    props = skimage.measure.regionprops(label_map)
+    a = [None] * len(props)
```
ngreenwald, line 19: It would be more pythonic to create this with a [list comprehension](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) instead of creating a list full of Nones.

ngreenwald:
Okay, looks good. Can you make all of the filenames lowercase to follow the naming scheme with the rest of the repo? Also, add _utils to the filenames so that it follows the conventions of the other util scripts

PR #93, opened 2020-05-16
New test edits

Jaiveers21:


segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -10,3 +13,78 @@ def test_calc_dist_matrix():
     dist_matrix = spatialanalysis_utils.calc_dist_matrix(test_mat)
     real_mat = np.array([[0, 5], [5, 0]])
     assert np.array_equal(dist_matrix, real_mat)
+
+
+def test_distmat():
```
ngreenwald, line 18: This is going to fail because these files aren't available on github. Keep it here, but comment it out. When we come back in a month and refactor it, we may want to have this code here to double check it runs

ngreenwald: Also, all the test functions should be named like the one above, with _test at the end of the function that's being tested. You can just put this inside the test function above

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -10,3 +13,78 @@ def test_calc_dist_matrix():
...
+    # test z and p values
+    marker_thresholds = pd.read_csv(
+        "/Users/jaiveersingh/Downloads/SpatialEnrichment/markerThresholds.csv")
+    # positive enrichment
+    cellArrayP = spatialanalysis_utils.make_expression_matrix("positive")
+    distMatP = spatialanalysis_utils.make_distance_matrix("positive")
+    closeNum, closeNumRand, z, muhat, sigmahat, p, h, adj_p, markerTitles = \
+        spatialorgmarkerexpression_utils.spatial_analysis(
+            distMatP, marker_thresholds, cellArrayP)
+    assert p[0, 1, 0] < .05
```
ngreenwald, line 73: check that the opposite p-values and z values are not significant

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -10,3 +13,78 @@ def test_calc_dist_matrix():
...
+    assert z[0, 1] > 0
+    # negative enrichment
+    cellArrayN = spatialanalysis_utils.make_expression_matrix("negative")
+    distMatN = spatialanalysis_utils.make_distance_matrix("negative")
+    closeNum, closeNumRand, z, muhat, sigmahat, p, h, adj_p, markerTitles = \
+        spatialorgmarkerexpression_utils.spatial_analysis(
+            distMatN, marker_thresholds, cellArrayN)
+    assert p[0, 1, 1] < .05
```
ngreenwald, line 81: Same here, check that the positive p and z values are not significant

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -10,3 +13,78 @@ def test_calc_dist_matrix():
...
+    assert z[0, 1] < 0
+    # no enrichment
+    cellArray = spatialanalysis_utils.make_expression_matrix("none")
+    distMat = spatialanalysis_utils.make_distance_matrix("none")
+    closeNum, closeNumRand, z, muhat, sigmahat, p, h, adj_p, markerTitles = \
+        spatialorgmarkerexpression_utils.spatial_analysis(
+            distMat, marker_thresholds, cellArray)
+    assert p[0, 1, 0] > .05
+    assert abs(z[0, 1]) < 1
```
ngreenwald, line 90: same here, check both

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,131 @@
...
+
+# randMat = np.random.randint(0,200,size=(60,60))
+# np.fill_diagonal(randMat, 0)
+
+# pv_cellarrayR = pd.read_csv("/Users/jaiveers
+# ingh/Desktop/tests/pvcellarrayR.csv")
+
+
+def spatial_analysis(dist_matrix, marker_thresholds, cell_array):
+    # Setup input and parameters
```
ngreenwald, line 36: add a docstring here that follows the same formats as the other functions in the repo

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    for j in range(0, markerNum):
+        for k in range(0, markerNum):
+            if z[j, k] > 0:
+                p_summary[j, k] = p[j, k, 0]
+            else:
+                p_summary[j, k] = p[j, k, 1]
+    (h, adj_p, aS, aB) = statsmodels.stats.multitest.multipletests(
+        p_summary, alpha=.05)
+
+    return \
```
ngreenwald, line 129: better to break this after half of the return values, rather than at the beginning

ngreenwald:
Okay, it failed on the non-significant test. The problem with that test is that 5% of the time, there's going to be a false positive just by the definition of the cutoffs we're using. We can work on that next. 

For now, let's work on breaking this function up into smaller pieces. 

Here's what I'd like it to look like.

calculate_channel_spatial_enrichment():
This function has the same arguments as the current spatial_analysis(). 

helper_function(): this function will be the triple nested for-loop that computes closenum and closenumrand. We're going to break this function up into smaller pieces as well, but just for this first pass we'll leave it as is. It should return closenum and closenumrand. 

calculate_enrichment_stats(): this function should take closenum and closenumrand return p, z, muhat, etc

Then calculate_channel_spatial_enrichment will use those two functions, instead of having the code directly in the main function. The output should be the same, and the test functions should work as is.

ngreenwald:
Also, please modify the test input data to only have 2 channels. It currently takes like 6 minutes to run this test

Jaiveers21:
I have pushed the code with spatial_analysis() broken up into the multiple functions

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -10,3 +14,138 @@ def test_calc_dist_matrix():
...
     real_mat = np.array([[0, 5], [5, 0]])
     assert np.array_equal(dist_matrix, real_mat)
+
+
+"""
+def test_distmat():
+    testtiff = io.imread(
+        "/Users/jaiveersingh/Documents/MATLAB/SpatialAnalysis/newLmod.tiff")
+    distmat = np.asarray(pd.read_csv(
+        "/Users/jaiveersingh/Documents/MATLAB/distancesMat5.csv", header=None))
```
ngreenwald, line 24: triple quotes are only used for docstrings: comment these out with # comment character instead. 
You can comment/uncomment multiple lines by highlighting then pressing cmd + /

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,181 @@
...
+randMat = np.random.randint(0,200,size=(60,60))
+np.fill_diagonal(randMat, 0)
+
+pv_cellarrayr = pd.read_csv("/Users/jaiveers"
+ingh/Desktop/tests/pvcellarrayR.csv")
+"""
+
+
+def helper_function(
+        patient_data, patient_data_markers, thresh_vec, dist_mat, marker_num):
```
ngreenwald, line 41: This is hard to read, keep it on one line. 

Change your pycharm settings to have 99 characters per line instead of 80, that's what we use

Jaiveers21:
I have pushed the new closenum and closenumrand functions

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,189 @@
...
+# pv_cellarrayr = pd.read_csv("/Users/jaiveers"
+# ingh/Desktop/tests/pvcellarrayR.csv")
+
+
+def helper_function_closenum(patient_data, patient_data_markers, thresh_vec, dist_mat, marker_num):
+
+    dist_lim = 100
+    close_num = np.zeros((marker_num, marker_num), dtype='int')
+    cell_label_idx = 1
+    marker1_num = 0
```
ngreenwald, line 44: This should be a list so that it works for arbitrary numbers of markers

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,189 @@
...
+# ingh/Desktop/tests/pvdistmatN.csv", header = None))
+#
+# randMat = np.random.randint(0,200,size=(60,60))
+# np.fill_diagonal(randMat, 0)
+#
+# pv_cellarrayr = pd.read_csv("/Users/jaiveers"
+# ingh/Desktop/tests/pvcellarrayR.csv")
+
+
+def helper_function_closenum(patient_data, patient_data_markers, thresh_vec, dist_mat, marker_num):
```
ngreenwald, line 39: Add a docstring here

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,189 @@
...
+
+    dist_lim = 100
```
ngreenwald, line 41: These values should be defined in the main function and passed in as variables

Jaiveers21:
I have pushed the new test functions for the helper functions and addressed the comments by checking with Erin's inputs again.

ngreenwald:
Okay, I think the next step is to clean the code up. Right now it's pretty hard to figure out what's going on.

The first step is just adding in more comments throughout to explain why you're doing certain actions. It would also be good to group lines of code together that accomplish similar things, and separate lines of code that accomplish different things with a comment. Not every line needs a comment, but each idea should have a comment

 You also have some places where you define a variable, then set that variable equal to another variable, then subset that variable to create a third variable, and only the third variable is actually used for the function. 

For example, test_thresholds is defined on line 134
Then on line 146, you set marker_thresh equal to test_thresholds
Then on line 147, you define thresh_vec by subsetting marker_thresh. 

A lot of this is a result of us refactoring it. Now that we have a better idea of what the flow of the functions look like, it's more obvious what variables are no longer needed anymore.

Jaiveers21:
I have pushed new commented code that works with the new inputs

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,263 @@
...
+    data_markers = data_all.drop(excluded_colnames, axis=1)
+    # list of all markers
+    # marker_titles = data_all.columns[marker_inds]
+    marker_titles = data_markers.columns
+    # length of marker list
+    marker_num = len(marker_titles)
+
+    # subsetting threshold matrix to only include column with threshold values
+    # thresh_vec = marker_thresholds.iloc[1:38, 1]
+    thresh_vec = marker_thresholds.iloc[0:20, 1]
```
ngreenwald, line 247: This index should be determined by the data. If you set it to be 0:20, it's not going to work if there are different values.

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,263 @@
...
+                    distance between centers of corresponding cells
+                marker_num: number of markers in expresion data
+                dist_lim: threshold for spatial enrichment distance proximity
+            Returns
+                close_num: marker x marker matrix with counts for cells
+                    positive for corresponding markers
+                marker1_num: list of number of cell labels for marker 1
+                marker2_num: list of number of cell labels for marker 2"""
+    # identifies column in expression matrix with cell labels
+    cell_label_idx = 24
```
ngreenwald, line 57: This is hardcoded

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,263 @@
...
+                # binarize the truncated distance matrix to only include cells within distance limit
+                rand_trunc_dist_mat_bin = np.zeros(rand_trunc_dist_mat.shape, dtype='int')
+                rand_trunc_dist_mat_bin[rand_trunc_dist_mat < dist_lim] = 1
+                # record the number of interactions and store in close_num_rand in index \
+                # corresponding to both markers, for every permutation
+                close_num_rand[j, k, r] = np.sum(np.sum(rand_trunc_dist_mat_bin))
+    return close_num_rand
+
+
+def calculate_enrichment_stats(close_num, close_num_rand):
```
ngreenwald, line 136: This function needs a test function as well

Jaiveers21:
I have pushed code including the error checking with a print statement and the points for loop. I tried it with Erin's data also so it should be flexible in terms of inputs and no hard coded sections.

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,265 @@
...
+    return z, muhat, sigmahat, p, h, adj_p
+
+
+def calculate_channel_spatial_enrichment(dist_matrix, marker_thresholds, cell_array,
+                                         excluded_colnames=None, points=None,
+                                         patient_idx=30, dist_lim=100, cell_label_idx=24, bootstrap_num=100):
+    """Spatial enrichment analysis to find significant interactions between cells expressing different markers.
+    Uses bootstrapping to permute cell labels randomly.
+
+    Args:
```
ngreenwald, line 200: update the docstring to indicate what the default is if None is the argument

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,265 @@
...
+    stats = []
+
+    if excluded_colnames is None:
+        excluded_colnames = ["cell_size", "Background", "HH3",
+                             "summed_channel", "label", "area",
+                             "eccentricity", "major_axis_length", "minor_axis_length",
+                             "perimeter", "fov"]
+
+    # Error Checking
+    if not np.isin(excluded_colnames, cell_array.columns).all():
```
ngreenwald, line 232: This doesn't actually raise an error, just prints a message. Check out the error checking in segmentation_utils where we raise a ValueError to prevent the function from continuing

Jaiveers21:
I pushed new edits and ready for next wave of changes

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -10,3 +14,284 @@ def test_calc_dist_matrix():
...
+        cellarray.iloc[20:40, 3] = 1
+        cellarray = cellarray.rename({0: 'cell_size', 1: 'Background', 14: "HH3",
+                                      23: "summed_channel", 24: "label", 25: "area", 26: "eccentricity",
+                                      27: "major_axis_length", 28: "minor_axis_length", 29: "perimeter",
+                                      30: "fov"}, axis=1)
+        return cellarray
+    elif(typeofencrichment == "positive"):
+        cellarrayp = pd.DataFrame(np.zeros((80, 31)))
+        cellarrayp[30] = "Point8"
+        cellarrayp[24] = np.arange(len(cellarrayp[1])) + 1
```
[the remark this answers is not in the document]
ngreenwald: For example: 
`# We create 8 cells positive for column index 2, and 8 cells positive for column index 3. These are within the dist_lim in positive_distance_matrix`

I think this line has a typo:
` cellarrayp.iloc[38:40, 2] = 1`
This should be column 3, right?


I think this section would be easier to follow if you grouped the iloc commands together that are creating close cells. So put these commands next to each other. Same for the rest of them
```
# comment explaining both of these lines
cellarrayp.iloc[0:8, 2] = 1    
cellarrayp.iloc[10:18, 3] = 1
```

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -10,3 +14,284 @@ def test_calc_dist_matrix():
...
+    cellarray[24] = np.arange(len(cellarray[1])) + 1
+    cellarray.iloc[0:4, 2] = 1
+    cellarray.iloc[0:4, 3] = 1
+    cellarray.iloc[4:9, 5] = 1
+    cellarray.iloc[4:9, 6] = 1
+    cellarray.iloc[9, 7] = 1
+    cellarray.iloc[9, 8] = 1
+
+    distmat = np.zeros((10, 10))
+    np.fill_diagonal(distmat, 0)
```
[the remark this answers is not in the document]
ngreenwald: Same thing. Explain the logic for these distmat values are being set. Someone should be able to read your comment, and undersatnd why all of these things are happening here

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -10,3 +15,232 @@ def test_calc_dist_matrix():
...
+    distmat[5, 6:9] = 50
+    distmat[7:9, 6] = 50
+    distmat[6, 7:9] = 50
+    distmat[8, 7] = 50
+    distmat[7, 8] = 50
+
+    return cellarray, distmat
+
+
+def test_closenum():
```
ngreenwald, line 150: This is great, much easier to follow

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -10,3 +15,232 @@ def test_calc_dist_matrix():
...
+    marker_num = 20
+
+    test_closenumrand = spatialorgmarkerexpression_utils.helper_function_closenumrand(
+        marker1_num, marker2_num, test_cellarray, test_distmat, marker_num, dist_lim=100,
+        cell_label_idx=24, bootstrap_num=100)
+
+    assert test_closenumrand.shape == (20, 20, 100)
+
+
+def test_calculate_enrichment_stats():
```
ngreenwald, line 202: This looks nice! Now let's do the same thing for the negative z score, and for non-significant.

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -10,3 +15,232 @@ def test_calc_dist_matrix():
...
+    stats_cnr = np.random.randint(1, 40, (20, 20, 100))
+
+    test_z, muhat, sigmahat, test_p, h, adj_p = spatialorgmarkerexpression_utils.calculate_enrichment_stats(
+        stats_cn, stats_cnr)
+
+    assert test_z[0, 0] > 0
+    assert test_p[0, 0, 0] < .05
+
+
+def test_spatial_analysis():
```
ngreenwald, line 217: This looks great!

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,267 @@
...
+# ingh/Desktop/tests/pvdistmatN.csv", header = None))
+#
+# randMat = np.random.randint(0,200,size=(60,60))
+# np.fill_diagonal(randMat, 0)
+#
+# pv_cellarrayr = pd.read_csv("/Users/jaiveers"
+# ingh/Desktop/tests/pvcellarrayR.csv")
+
+
+def helper_function_closenum(patient_data, patient_data_markers, thresh_vec,
```
ngreenwald, line 39: Let's rename this function to 'compute_close_cell_num'

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,267 @@
...
+                    marker2poslabels - 1))]
+            # binarize the truncated distance matrix to only include cells within distance limit
+            trunc_dist_mat_bin = np.zeros(trunc_dist_mat.shape, dtype='int')
+            trunc_dist_mat_bin[trunc_dist_mat < dist_lim] = 1
+            # record the number of interactions and store in close_num in index corresponding to both markers
+            close_num[j, k] = np.sum(np.sum(trunc_dist_mat_bin))
+    return close_num, marker1_num, marker2_num
+
+
+def helper_function_closenumrand(marker1_num, marker2_num, patient_data,
```
ngreenwald, line 90: Let's change this function name to 'compute_close_cell_num_random'

Jaiveers21:
I have pushed code with the new edits

segmentation/utils/spatialanalysis_utils.py
```diff
@@ -1,16 +1,17 @@
...
 
 
 def calc_dist_matrix(label_map):
     """Generate matrix of distances between center of pairs of cells
-        Args
-            label_map: numpy array with unique cells given unique pixel labels
-        Returns
-            dist_matrix: cells x cells matrix with the euclidian
-            distance between centers of corresponding cells"""
+
```
ngreenwald, line 9: change this file name to be spatial_analysis_utils

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -1,12 +1,310 @@
 import numpy as np
+import pandas as pd
+import random
 from segmentation.utils import spatialanalysis_utils
+from segmentation.utils import spatialorgmarkerexpression_utils
+import importlib
```
ngreenwald, line 6: change file name to be spatial_analysis_utils_test

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -1,12 +1,310 @@
...
+
+    test_z, muhat, sigmahat, test_p, h, adj_p = spatialorgmarkerexpression_utils.calculate_enrichment_stats(
+        stats_cn, stats_cnr)
+
+    assert abs(test_z[0, 0]) < 1
+    assert test_p[0, 0, 1] > .05
+    assert test_p[0, 0, 0] > .05
+
+
+def test_spatial_analysis():
```
ngreenwald, line 279: change name to test_full_function_name

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,246 @@
+import pandas as pd
+import numpy as np
```
ngreenwald, line 2: change file name to spatial_analysis.py

segmentation/utils/spatialorgmarkerexpression_utils.py
```diff
@@ -0,0 +1,246 @@
...
+            positive for corresponding markers
+        marker1_num: list of number of cell labels for marker 1
+        marker2_num: list of number of cell labels for marker 2"""
+
+    # create close_num, marker1_num, and marker2_num
+    close_num = np.zeros((marker_num, marker_num), dtype='int')
+    marker1_num = []
+    marker2_num = []
+
+    # later implement outside for loop for dif points
```
ngreenwald, line 44: delete this comment

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -1,12 +1,310 @@
...
+    # test z and p values
+
+    marker_thresholds = make_threshold_mat()
+
+    # positive enrichment
+    cellarrayp = make_expression_matrix("positive")
+    distmatp = make_distance_matrix("positive")
+    values, stats = \
```
ngreenwald, line 287: This is good. I would then unpack the tuple so it's more intuitive what's being tested here

```
values, stats = function_output()
p_values, z_scores, _, _, _ = stats
assert (p_values[0, 0] == )

```

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -1,12 +1,310 @@
...
+    stats_cn = np.zeros((20, 20))
+    stats_cn[:, :] = 80
+
+    # generate random closenumrand matrix, ensuring no enrichment
+    stats_cnr = np.random.randint(78, 82, (20, 20, 100))
+
+    test_z, muhat, sigmahat, test_p, h, adj_p = spatialorgmarkerexpression_utils.calculate_enrichment_stats(
+        stats_cn, stats_cnr)
+
+    assert abs(test_z[0, 0]) < 1
```
ngreenwald, line 274: Let's change this to <2 rather than <1. 

We can get random Z-scores that are larger than 1, but aren't significant, and we don't want the test to fail because of that

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -1,12 +1,310 @@
...
+        # assigning values to the patient label and cell label columns
+        cellarrayp[30] = "Point8"
+        cellarrayp[24] = np.arange(len(cellarrayp[1])) + 1
+        # We create 8 cells positive for column index 2, and 8 cells positive for column index 3.
+        # These are within the dist_lim in distmatp (positive enrichment distance matrix).
+        cellarrayp.iloc[0:8, 2] = 1
+        cellarrayp.iloc[10:18, 3] = 1
+        # We create 4 cells in column index 2 and column index 3 that are also positive
+        # for their respective markers.
+        cellarrayp.iloc[28:30, 2] = 1
```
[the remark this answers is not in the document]
Jaiveers21: there is 36:38 and two other places (27 and 30) in column 3. I just wanted 4 in both column 2 and 3 without the same cells being positive for both markers.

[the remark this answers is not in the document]
ngreenwald: Got it. In that case, let's consolidate the four positive cells for each marker into consecutive values, that way a single comment can describe what's going on.

Jaiveers21:
I pushed the new edits. Should I change bootstrap num to 1000 in the code?

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,245 @@
...
+# marker_thresholds = pd.read_csv("/Users/jaiveersingh/Downloads/Sp"
+#                                 "atialEnrichment/markerThresholds.csv")
+# dist_matrix = np.asarray(pd.read_csv("/Users/jaiveersingh/Documen"
+#                                      "ts/MATLAB/distancesMat5.csv",
+#                                      header=None))
+
+
+def compute_close_cell_num(patient_data, patient_data_markers, thresh_vec,
+                           dist_mat, marker_num, dist_lim, cell_label_idx):
+    """Finds positive cell labels and creates matrix with counts for cells positive for corresponding markers.
```
ngreenwald, line 20: Sometimes functions are hard to explain in just a single line. In that case, it's good to have a single line description, followed by a more in-depth description separated by a full space. I think this function needs a bit more description so people understand what it does

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,245 @@
...
+    marker2_num = []
+
+    for j in range(0, marker_num):
+        # Identify cell labels that are positive for respective markers
+        marker1_thresh = thresh_vec.iloc[j]
+        marker1posinds = patient_data_markers[patient_data_markers.columns[j]] > marker1_thresh
+        marker1poslabels = patient_data.loc[marker1posinds, patient_data.columns[cell_label_idx]]
+        marker1_num.append(len(marker1poslabels))
+        for k in range(0, marker_num):
+            # Identify cell labels that are positive for above marker and all other markers
```
ngreenwald, line 51: Here you're describing what the for loop does across the entire loop (all other markers). What what you want to describe is what it is doing inside a single iteration of the loop, which is what someone reading the code will be trying to figure out. 


I would change this to 'identify all cells that are positive for kth marker'

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,245 @@
...
+            marker2_thresh = thresh_vec.iloc[k]
+            marker2posinds = patient_data_markers[patient_data_markers.columns[k]] > marker2_thresh
+            marker2poslabels = patient_data.loc[marker2posinds, patient_data.columns[cell_label_idx]]
+            marker2_num.append(len(marker2poslabels))
+            # Subset the distance matrix to only include positive cell labels
```
ngreenwald, line 56: I would change this to '...only include cells positive for markers j and k'

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,245 @@
...
+            trunc_dist_mat = dist_mat[np.ix_(
+                np.asarray(marker1poslabels - 1), np.asarray(
+                    marker2poslabels - 1))]
+            # Binarize the truncated distance matrix to only include cells within distance limit
+            trunc_dist_mat_bin = np.zeros(trunc_dist_mat.shape, dtype='int')
+            trunc_dist_mat_bin[trunc_dist_mat < dist_lim] = 1
+            # Record the number of interactions and store in close_num in index corresponding to both markers
```
ngreenwald, line 63: This is great, very descriptive

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,245 @@
...
+            close_num[j, k] = np.sum(np.sum(trunc_dist_mat_bin))
+    return close_num, marker1_num, marker2_num
+
+
+def compute_close_cell_num_random(marker1_num, marker2_num, patient_data,
+                                  dist_mat, marker_num, dist_lim, cell_label_idx, bootstrap_num):
+    """Uses bootstrapping to permute cell labels randomly.
```
ngreenwald, line 70: It permutes them randomly, and also records the number of close cells in that random setup. I think that's a key point here

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,245 @@
...
+    (h, adj_p, aS, aB) = statsmodels.stats.multitest.multipletests(
+        p_summary, alpha=.05)
+
+    return z, muhat, sigmahat, p, h, adj_p
+
+
+def calculate_channel_spatial_enrichment(dist_matrix, marker_thresholds, all_patient_data,
+                                         excluded_colnames=None, points=None,
+                                         patient_idx=30, dist_lim=100, cell_label_idx=24, bootstrap_num=100):
+    """Spatial enrichment analysis to find significant interactions between cells expressing different markers.
```
ngreenwald, line 173: Good. Because this is more than one line, the best practice is to have a single line that describes at a very high level what the function does, then any additional lines separated below that give more specifics

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,245 @@
...
+            and negative enrichment
+        h: matrix indicating whether
+            corresponding marker interactions are significant
+        adj_p: fdh_br adjusted p values"""
+    # Get the number of markers and number of permutations
+    marker_num = close_num.shape[0]
+    bootstrap_num = close_num_rand.shape[2]
+
+    # Create z, muhat, sigmahat, and p
+    z = np.zeros((marker_num, marker_num))
```
ngreenwald, line 138: lets instead create a single array to hold all of these values. That way, we can save this single array at the end. 

The dimensions should be [stats x marker1 x marker2]. Let's make a separate p_pos and p_neg array, that way everything will have the same shape. 

Then, we can use xarray to label the dimensions. Take a look at https://github.com/angelolab/segmentation/blob/76fbf4682f108bd047f878262337c0be9a3c26e5/segmentation/utils/data_utils.py#L48 to see how to create an xarray

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -1,16 +1,17 @@
 import numpy as np
-import math
+import pandas as pd
 import skimage.measure
 from scipy.spatial.distance import cdist
 
 
 def calc_dist_matrix(label_map):
     """Generate matrix of distances between center of pairs of cells
```
ngreenwald, line 8: Let's modify this to instead take in an array of label maps, loop over them, and return the output as an array of dist_mats. The input to this function is the segmentation_labels.xr produced by the segmentation function: https://github.com/angelolab/segmentation/blob/76fbf4682f108bd047f878262337c0be9a3c26e5/segmentation/utils/segmentation_utils.py#L70. It has a shape of [points, rows, cols, 1]

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -0,0 +1,314 @@
...
+# def test_distmat():
+#     testtiff = io.imread(
+#         "/Users/jaiveersingh/Documents/MATLAB/SpatialAnalysis/newLmod.tiff")
+#     distmat = np.asarray(pd.read_csv(
+#         "/Users/jaiveersingh/Documents/MATLAB/distancesMat5.csv", header=None))
+#     testmat = spatialanalysis_utils.calc_dist_matrix(testtiff)
+#     assert np.allclose(distmat, testmat)
+
+
+def make_distance_matrix(typeofenfrichment):
```
ngreenwald, line 30: change this argument to be enrichment_type

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -0,0 +1,314 @@
...
+        # We create 8 cells positive for column index 2, and 8 cells positive for column index 3.
+        # These are within the dist_lim in dist_mat_pos (positive enrichment distance matrix).
+        all_patient_data_pos.iloc[0:8, 2] = 1
+        all_patient_data_pos.iloc[10:18, 3] = 1
+        # We create 4 cells in column index 2 and column index 3 that are also positive
+        # for their respective markers.
+        all_patient_data_pos.iloc[28:32, 2] = 1
+        all_patient_data_pos.iloc[32:36, 3] = 1
+        # Assign column names to columns not for markers (columns to be excluded)
+        all_patient_data_pos = all_patient_data_pos.rename({0: 'cell_size', 1: 'Background', 14: "HH3",
```
ngreenwald, line 96: You can make a dict called col_names, and define it as equal to everything in brackets. Then, for each of these pandas arrays, call .rename(col_names, axis=1). That way you don't have to retype it

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -0,0 +1,314 @@
...
+        return all_patient_data_neg
+
+
+def make_threshold_mat():
+    thresh = pd.DataFrame(np.zeros((20, 2)))
+    thresh.iloc[:, 1] = .5
+    return thresh
+
+
+def make_test_closenum():
```
ngreenwald, line 127: This function doesn't make an example closenum, it makes example data for closenum. See if you can rename it so it's more clear what the point of this function is

segmentation/utils/spatialanalysis_utils_test.py
```diff
@@ -1,12 +0,0 @@
-import numpy as np
-from segmentation.utils import spatialanalysis_utils
-
-
```
ngreenwald, line 4: We're going to be putting more stuff in here in the next round, don't delete this yet

Jaiveers21:
I pushed new code with the Xarrays and comment cleanup. Would you be able to check the Xarray code or see if any of it can be cleaned up?

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,272 @@
...
+    # List of all markers
+    marker_titles = data_markers.columns
+    # Length of marker list
+    marker_num = len(marker_titles)
+
+    # Create stats Xarray with the dimensions (points, stats variables, number of markers, number of markers)
+    stats_raw_data = np.zeros((num_points, 7, marker_num, marker_num))
+    coords = [points, ["z", "muhat", "sigmahat", "p_pos", "p_neg", "h", "p_adj"], marker_titles,
+              marker_titles]
+    dims = ["points", "stats", "marker1", "marker2"]
```
ngreenwald, line 248: yup, this is great!

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,272 @@
...
+    close_num = np.zeros((marker_num, marker_num), dtype='int')
+    marker1_num = []
+    marker2_num = []
+
+    for j in range(0, marker_num):
+        # Identify cell labels that are positive for respective markers
+        marker1_thresh = thresh_vec.iloc[j]
+        marker1posinds = patient_data_markers[patient_data_markers.columns[j]] > marker1_thresh
+        marker1poslabels = patient_data.loc[marker1posinds, patient_data.columns[cell_label_idx]]
+        marker1_num.append(len(marker1poslabels))
```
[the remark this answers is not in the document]
Jaiveers21: marker1posinds is a boolean array, so its length would still be 80 for example, whereas the length of marker1poslabels is only the labels with TRUE, so it would only be 12. Should I instead replace it with like marker1_num.append(sum(marker1posinds))

Jaiveers21:
I pushed the code. Is the test function code where I create example_all_patient_data not necessary now too since it is not needed as an argument to test close_num and close_num_rand

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,259 @@
...
+# cell_array = pd.read_csv("/Users/jaiveersingh/Downloads/SpatialEn"
+#                          "richment/granA_cellpheno_CS-asinh-norm_revised.csv")
+# marker_thresholds = pd.read_csv("/Users/jaiveersingh/Downloads/Sp"
+#                                 "atialEnrichment/markerThresholds.csv")
+# dist_matrix = np.asarray(pd.read_csv("/Users/jaiveersingh/Documen"
+#                                      "ts/MATLAB/distancesMat5.csv",
+#                                      header=None))
+
+
+def compute_close_cell_num(patient_data_markers, thresh_vec,
```
ngreenwald, line 19: This isn't a main function, so now that we're happy with it lets move this to spatial_analysis_utils

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,259 @@
...
+            trunc_dist_mat = dist_mat[np.ix_(np.asarray(marker1posinds), np.asarray(marker2posinds))]
+            # Binarize the truncated distance matrix to only include cells within distance limit
+            trunc_dist_mat_bin = np.zeros(trunc_dist_mat.shape, dtype='int')
+            trunc_dist_mat_bin[trunc_dist_mat < dist_lim] = 1
+            # Record the number of interactions and store in close_num in index corresponding to both markers
+            close_num[j, k] = np.sum(np.sum(trunc_dist_mat_bin))
+    return close_num, marker1_num, marker2_num
+
+
+def compute_close_cell_num_random(marker1_num, marker2_num,
```
ngreenwald, line 69: Let's move this one to utils as well

segmentation/utils/spatial_analysis.py
```diff
@@ -0,0 +1,259 @@
...
+                # Binarize the truncated distance matrix to only include cells within distance limit
+                rand_trunc_dist_mat_bin = np.zeros(rand_trunc_dist_mat.shape, dtype='int')
+                rand_trunc_dist_mat_bin[rand_trunc_dist_mat < dist_lim] = 1
+                # Record the number of interactions and store in close_num_rand in the index
+                # corresponding to both markers, for every permutation
+                close_num_rand[j, k, r] = np.sum(np.sum(rand_trunc_dist_mat_bin))
+    return close_num_rand
+
+
+def calculate_enrichment_stats(close_num, close_num_rand):
```
ngreenwald, line 109: Let's move this one to utils as well

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -0,0 +1,308 @@
...
+
+
+def make_example_data_closenum():
+    # Creates example data for the creation of the closenum matrix in the below test function
+
+    # Create example all_patient_data cell expression matrix
+    all_patient_data = pd.DataFrame(np.zeros((10, 31)))
+    # Assigning values to the patient label and cell label columns
+    all_patient_data[30] = "Point8"
+    all_patient_data[24] = np.arange(len(all_patient_data[1])) + 1
```
ngreenwald, line 130: Yes, we can change this to just be patient_data_markers

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -0,0 +1,308 @@
...
+    stats_cnr = np.random.randint(78, 82, (20, 20, 100))
+
+    stats_xr = spatial_analysis.calculate_enrichment_stats(stats_cn, stats_cnr)
+
+    assert abs(stats_xr.loc["z", 0, 0]) < 1
+    assert stats_xr.loc["p_neg", 0, 0] > .05
+    assert stats_xr.loc["p_pos", 0, 0] > .05
+
+
+def test_calculate_channel_spatial_enrichment():
```
ngreenwald, line 274: Let's move this to function to a new file, called spatial_analysis_test, since it's testing a function in spatial_analysis, not spatial_analysis_utils

Jaiveers21:
I pushed the new files

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -0,0 +1,182 @@
...
+import random
+from segmentation.utils import spatial_analysis_utils
+from segmentation.utils import spatial_analysis
+import importlib
+importlib.reload(spatial_analysis)
+importlib.reload(spatial_analysis_utils)
+
+
+def test_calc_dist_matrix():
+    test_mat = np.zeros((1, 512, 512), dtype="int")
```
ngreenwald, line 12: Let's make a different easy pythagorean triple in a different slice of this array, just to make sure its preserving the information correctly, now that we're passing in multiple different labeled arrays

Jaiveers21:
I have pushed code with the labels

PR #108, opened 2020-06-26
Synthetic data generation (Stage 1: distance-based)

alex-l-kong:
Now that we've got a good foundation for conducting spatial analysis, it will help to develop a set of robust functions that can automatically generate test data that we can simply call instead of hard coding. The first stage is to assume that distances between cell centroids are the only parameter we care about.


alex-l-kong:
The distance matrix tests are passing, but I realize only now that it may not be formatted in a way such that it is compatible with other spatial_analysis code.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+# Constants for random centroid matrix generation
+A_CENTROID_FACTOR = 0.5
+B_CENTROID_FACTOR = 0.9
+C_CENTROID_FACTOR = 0.4
+
+A_CENTROID_COV = [[1, 0], [0, 1]]
+B_CENTROID_COV = [[1, 0], [0, 1]]
+C_CENTROID_COV = [[1, 0], [0, 1]]
```
ackagel, line 25: Since these are all default values of function arguments, they can just be declared in the function args.  For example, in `direct_init_dist_matrix`:

```
def direct_init_dist_matrix(num_A=100, num_B=100, num_C=100,
                            distr_AB={'mean' : 100, 'var' : 1}, 
                            dist_AC={'mean' : 20, 'var' : 1},
                            seed=None):
```

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def direct_init_dist_matrix(num_A=100, num_B=100, num_C=100, distr_AB=None, distr_AC=None, seed=None):
```
ackagel, line 28: Depending on whether this is used for development-testing or by lab users, it might be good to set the default seed to 0.  If it's for development/testing, we'll probably want to default to a set seed, and set `seed=None` manually.  If it's for lab users, the current default is fine, since they probably want 'non-seeded' randomization.

Honestly, it depends on which behavior is used more often; I don't know the full context of this function's intended usage

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,204 @@
...
+    else:
+        mean_ab = distr_AB['mean']
+        var_ab = distr_AB['var']
+
+    if not distr_AC:
+        mean_ac = AC_DIST_MEAN
+        var_ac = AC_DIST_VAR
+    else:
+        mean_ac = distr_AC['mean']
+        var_ac = distr_AC['var']
```
ackagel, line 65: referencing the comment on lines 11-25, no need to set new variables; just call `distr_AB` when necessary

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: Yeah definitely! Forgot to take those globals out, will absolutely fix.

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: I'll have to talk to Noah about this, thanks for pointing it out.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,204 @@
...
+    second_row = np.concatenate((random_ab.T, (random_bb + random_bb.T) / 2, random_bc), axis=1)
+    third_row = np.concatenate((random_ac.T, random_bc.T, (random_cc + random_cc.T) / 2), axis=1)
+
+    # then concatenate them together
+    dist_mat = np.concatenate((first_row, second_row, third_row), axis=0)
+
+    # assert that the created submatrix is symmetric
+    assert np.allclose(dist_mat, dist_mat.T, rtol=1e-05, atol=1e-08)
+
+    return dist_mat
```
ackagel, line 94: This method of generating a random distance matrix can create non-physical results.  Consider the case of just 3 cell-A's in a 2d plane, with distances a12, a23, and a13.  There are two geometric cases for this:

1. The cells are co-linear, so some reordering of a12 + a23 = a13 must be true
2. The cells form a triangle, and hence follow the all of the inequalities:
      - a12 + a23 < a13
      - a13 + a23 < a12
      - a12 + a13 < a23

This method of generating distance matrices doesn't guarantee these criteria and creates un-physical situations.  Rather than fix this, it's probably best remove this function and use `get_random_centroid_centers` into `point_init_dist_matrix` instead, that defaults to having mean_A closer to mean_B than mean_C

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def get_random_centroid_centers(size_img=(1024, 1024), num_A=100, num_B=100, num_C=100, distr_A=None, distr_B=None, distr_C=None, seed=None):
```
ackagel, line 97: instead of inputting a matrix for the covariance, you could use a tuple (sigx, sigy, theta) and generate the covariance matrix with something like:
```
# generate scaling
scale = np.array([[sigx**2, 0], [0, sigy**2]])

# generate rotation matrix
s, c = np.sin(theta), np.cos(theta)
rot = np.array([[c, -s], [s, c]])

# compute covariance matrix
cov = rot * scale * rot.T
```
this guarantees that the final matrix is symmetric positive definite, and covers all possible cases; could maybe go into a separate function too

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,204 @@
...
+    # if specified, set the random seed
+    if seed:
+        np.random.seed(seed)
+
+    # use the multivariate_normal distribution to generate the points
+    a_points = np.random.multivariate_normal(a_mean, a_cov, num_A)
+    b_points = np.random.multivariate_normal(b_mean, b_cov, num_B)
+    c_points = np.random.multivariate_normal(c_mean, c_cov, num_C)
+
+    return a_points, b_points, c_points
```
ackagel, line 155: doesn't really matter but it's probably better to return these as a tuple or a dictionary.  i'd go tuple personally, but i have nothin against dictionaries

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,204 @@
...
+    # create each matrix row
+    # we need to correct aa, bb, and cc to ensure symmetry
+    first_row = np.concatenate(((a_a_dist + a_a_dist.T) / 2, a_b_dist, a_c_dist), axis=1)
+    second_row = np.concatenate((a_b_dist.T, (b_b_dist + b_b_dist.T) / 2, b_c_dist), axis=1)
+    third_row = np.concatenate((a_c_dist.T, b_c_dist.T, (c_c_dist + c_c_dist) / 2), axis=1)
+
+    # and then the entire matrix
+    dist_mat = np.concatenate((first_row, second_row, third_row), axis=0)
+
+    return dist_mat
```
ackagel, line 204: I believe the current format for `dist_mat` is a dictionary with fovs as keys and matrices as values; just wrapping the matrix in a dictionary with something like 'generated' as the key would probably be fine.

building on that, you could internally loop the function over multiple conditions and store the outputs in the same dictionary together, labelled with numbered strings; could be useful if you need to generate multiple different synthetic distributions.  Would need to have a conditions array/dict as an argument to loop through tho

segmentation/utils/synthetic_spatial_datagen_test.py
```diff
@@ -0,0 +1,60 @@
...
+import scipy
+import os
+from scipy.spatial.distance import cdist
+from segmentation.utils import synthetic_spatial_datagen
+import importlib
+importlib.reload(synthetic_spatial_datagen)
+
+
+# random seed
+RSEED = 42
```
ackagel, line 14: found the answer

segmentation/utils/synthetic_spatial_datagen_test.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_direct_init_dist_matrix():
```
ackagel, line 17: un-needed if I didn't make an oopsie in the `direct_init_dist_matrix` comment

segmentation/utils/synthetic_spatial_datagen_test.py
```diff
@@ -0,0 +1,60 @@
...
+    assert np.allclose(sample_dist_mat, sample_dist_mat.T, rtol=1e-05, atol=1e-08)
+
+    # assert the average of the distance between A and B is greater
+    # than the average of the distance between A and C.
+    # this may not eliminate the possibility that the null is proved true
+    # but it's definitely a great check that can ensure greater success
+    assert sample_dist_mat[:100, 100:200].mean() > sample_dist_mat[:100, 200:].mean()
+
+
+def test_point_init_dist_matrix():
```
ackagel, line 40: looks good; not sure what else is needed testing-wise at the moment.

ackagel: on second thought, maybe a test over some different distribution constants would be good, especially to cover the `get_random_centroid_centers` function; although that could maybe have its own test function

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,204 @@
...
+    a_c_dist = cdist(a_points, c_points)
+    b_b_dist = cdist(b_points, b_points)
+    b_c_dist = cdist(b_points, c_points)
+    c_c_dist = cdist(c_points, c_points)
+
+    # create each matrix row
+    # we need to correct aa, bb, and cc to ensure symmetry
+    first_row = np.concatenate(((a_a_dist + a_a_dist.T) / 2, a_b_dist, a_c_dist), axis=1)
+    second_row = np.concatenate((a_b_dist.T, (b_b_dist + b_b_dist.T) / 2, b_c_dist), axis=1)
+    third_row = np.concatenate((a_c_dist.T, b_c_dist.T, (c_c_dist + c_c_dist) / 2), axis=1)
```
ackagel, line 199: I'm pretty sure that `a_a_dist` is guaranteed to be symmetric.  no real harm in ensuring symmetry anyways, but for big matrices, this might incur a performance drop.  Consider this comment as negligible as the performance difference.

ngreenwald:
Okay, here's my high-level comments. 

Overall, it looks super good! To help guide you for which of Adam's comments to focus on, I think it's useful to remember what the purpose of these functions are. 

For the pure distance matrix generation, we want to use that to test the functions Jay put together that take a distance matrix as an input, and identify groups of cells that are closer to one another than expected by chance. 
For example, this function: https://github.com/angelolab/segmentation/blob/ec733863f0f9186759abc2ebe8281dd0b2ac0473/segmentation/utils/spatial_analysis.py#L24

I think the format you have will already slot very easily into that testing framework. In line with that, it doesn't really matter if we have non-pythagorean relationships between triplets of cells, because we're only testing that given a distance matrix with certain values, the relative spatial enrichment of those populations is accurately computed. 


The second set of functions produces cell centroids with those same specifications. This would be useful for testing the same thing, but at the level of the original cell locations, rather than the already computed distance matrix. However, I think the only reason that something like this would be necessary is for when we want to do integration testing of the entire pipeline. i.e., we start with cell centroids, and want to propagate through the calculate_distance_matrices function Jay wrote, and then use that same distance matrix as input to one of the spatial functions to make sure the output is correct. 

If that's the use case, then we want the output of the centroid generating function to be inputs to the calc_dist_matrix function that the pipeline uses, and we don't need to create a second test function that makes a distance matrix from these centroids. 

Let me know if I misunderstood something about the purpose of the second and third function.

alex-l-kong:
Thanks, Adam and Noah for the code review!

@ngreenwald I can work on fixing the second test function to be more in line with how the pipeline works. I had originally thought that we would need to test that the generated set of points produced a compatible matrix, but from what I've heard, it doesn't seem like that will be necessary.

In that case, if I understand correctly, the return value of get_random_centroid_centers should actually be the label_map passed into the calc_dist_matrix function (which means I'll need to work on reformatting the output so that calc_dist_matrix can process it). That eliminates the need for the second test function because the resulting return value of calc_dist_matrix then can be verified with test_direct_init_dist_matrix (or a different test function).

ngreenwald:
Exactly. If you create a binary matrix with true values anywhere there is a centroid, you can pass that into skimage.label to generate a label map.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,167 @@
...
+import xarray as xr
+import copy
+
+from segmentation.utils import spatial_analysis_utils as sau
+from scipy.spatial.distance import cdist
+from skimage.measure import label
+
+
+def direct_init_dist_matrix(num_A=100, num_B=100, num_C=100,
+                            ab_dist_mean=100, ab_dist_var=1, ac_dist_mean=20, ac_dist_var=1,
```
ngreenwald, line 16: Can remove arguments that aren't being used anymore in new version

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,167 @@
...
+    random_aa = np.random.normal(0, 1, (num_A, num_A))
+    random_ab = np.random.normal(mean_ab, var_ab, (num_A, num_B))
+    random_ac = np.random.normal(mean_ac, var_ac, (num_A, num_C))
+    random_bb = np.random.normal(0, 1, (num_B, num_B))
+    random_bc = np.random.normal(0, 1, (num_B, num_C))
+    random_cc = np.random.normal(0, 1, (num_C, num_C))
+
+    # create each row one-by-one first
+    # we need to correct each aa, bb, and cc matrix to ensure symmetry
+    first_row = np.concatenate(((random_aa + random_aa.T) / 2, random_ab, random_ac), axis=1)
```
ngreenwald, line 67: Perhaps we can find a different name besides "row", since it's not a single row? This was confusing me when I tried to figure out what's going on here.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,167 @@
...
+    second_row = np.concatenate((random_ab.T, (random_bb + random_bb.T) / 2, random_bc), axis=1)
+    third_row = np.concatenate((random_ac.T, random_bc.T, (random_cc + random_cc.T) / 2), axis=1)
+
+    # then concatenate them together
+    dist_mat = np.concatenate((first_row, second_row, third_row), axis=0)
+
+    # assert that the created submatrix is symmetric
+    assert np.allclose(dist_mat, dist_mat.T, rtol=1e-05, atol=1e-08)
```
ngreenwald, line 75: You guys probably know better than me, but my understanding is that best practice is to only have asserts in actual test functions?

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    This function will return a random dist matrix such that the distance between cells
+    of types A and B are overall larger than the distance between cells of types A and C
+
+    Each row and column representing a cell.
+    We generate the points using Gaussian distributions
+    Ideally, the parameters for A to B distances will be set such that they produce a lower range of values
+    than A to C distances.
+
+    Will return a random distance matrix as well as the dictionary of associated cell: label IDs
```
ngreenwald, line 28: Lets update this with Args, Returns format, and then update the Returns to match what the function is now returning

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -0,0 +1,167 @@
...
+    rows, cols = zip(*non_dup_points)
+
+    # generate the binary matrix to pass into label_map
+    binary_mat = np.zeros(size_img)
+    binary_mat[rows, cols] = True
+
+    # generate the label matrix for the image now
+    label_mat = label(binary_mat)
+
+    # TODO: check and see if we need to return just the label matrix or if we need to convert it to a distance matrix first
```
ngreenwald, line 155: I think we'll want this to output data ready to be run through calc_dist_matrix

alex-l-kong:
The return values are in the right format, and the function descriptions have been updated to reflect this, so the last major check is if the fovs and segmentation_label dimensions have been initialized correctly or if more work needs to be done on that. I would assume because we're interested in just the distance_matrix, that these fields don't matter as much.

ngreenwald:
Yeah that looks fine to me. We may decide we want to change those values, but that would be after we start integrating this into the spatial analysis tests.

Issue #109, opened 2020-06-26
Add option to load in pre-segmented label maps


ngreenwald:
Currently, the segmentation masks are produced by this [watershed function](https://github.com/angelolab/segmentation/blob/ec733863f0f9186759abc2ebe8281dd0b2ac0473/segmentation/utils/segmentation_utils.py#L19). 

However, we're going to move towards deepcell itself producing the masks. That means the output people will get is most likely a single folder with the masks for all of their FOVs. We should add the option to load a folder of masks, convert it to an xarray, and then pass it the subsequent function in the pipeline, which is [here](https://github.com/angelolab/segmentation/blob/ec733863f0f9186759abc2ebe8281dd0b2ac0473/segmentation/utils/segmentation_utils.py#L326)

We should also update the example post-processing jupyter notebook.

PR #110, opened 2020-07-01
Loading TIFF segmentation labels

ackagel:
Primary purpose of PR is to address issue #109.  Main changes include:

 - Adding functionality to `data_utils.load_imgs_from_dir` to load images from a single directory, and match them to their respective fov's via a prefixed naming convention
- Adding `data_utils.tiffs_to_xr_labels` which uses the new functionality described above to load a directory of tiffs as an xarray of segmentation_labels, and rename the xarray coords to match conventions established in `segmentation_utils.watershed_transform`.


ackagel:
just an unrelated heads up, `data_utils_test.test_crop_image_stack()` was using >4GB of memory through pytest, so I changed it to use int16's in its arrays instead of floats

segmentation/utils/data_utils.py
```diff
@@ -142,20 +156,53 @@ def load_imgs_from_dir(data_dir, img_sub_folder=None, fovs=None, imgs=None,
...
-
     if variable_sizes:
         row_coords, col_coords = range(1024), range(1024)
     else:
         row_coords, col_coords = range(test_img.shape[0]), range(test_img.shape[1])
 
+    if use_filenames:
+        fovs = [fov.split(delimiter)[0] for fov in fovs]
+        img_names = [image_name]
+
```
alex-l-kong, line 167: This sort of goes along with what Noah discussed in the meeting. The logic looks good, but it just might be a bit much to incorporate several different cases into the same function which could become confusing if it becomes a habit. Again, it might still be a good idea to condense all this into one function for now, but as more options become available, it might be best to split into separate functions.

segmentation/utils/data_utils.py
```diff
@@ -109,6 +115,9 @@ def load_imgs_from_dir(data_dir, img_sub_folder=None, fovs=None, imgs=None,
     if len(imgs) == 0:
         raise ValueError("No imgs found in designated folder")
 
+    # remove .tif or .tiff from image name
+    img_names = [os.path.splitext(img)[0] for img in imgs]
+
```
alex-l-kong, line 120: Is there a reason these lines were moved from 145-147 to here?

segmentation/utils/data_utils.py
```diff
@@ -142,20 +156,53 @@ def load_imgs_from_dir(data_dir, img_sub_folder=None, fovs=None, imgs=None,
     if np.min(img_data) < 0:
         raise ValueError("Integer overflow from loading TIF image, try a larger dtype")
 
-    # remove .tif or .tiff from image name
-    img_names = [os.path.splitext(img)[0] for img in imgs]
-
```
alex-l-kong, line 147: Following up on 118-120, if it's possible, it might be nice to add an else statement after lines 164-166 to handle this case of img_name extraction.

segmentation/utils/data_utils.py
ackagel: pickybacking on the already-written code of the function was made easier by essentially making `fovs` act like `imgs`, instead of having `fovs = [""]` and using imgs the same.  So theres a point where I move `imgs` into `fovs` and redefine `imgs`, so `img_names` needed to be set earlier.  In hindsight, I think I could've changed it to:
```
img_names = [os.path.splitext(fov)[0] for fov in fovs]
```
and left it where it was, but there's not much of a difference.  No matter now tho, since I'll probably move this jazz into a separate function.

tl;dr makes piggy backing on already written code easier if it's declared earlier

segmentation/utils/data_utils.py
ackagel: It's no trouble to move this into a separate function, so that's what I'll do.  I'll probably rename this current function tho and make the appropriate reference changes in other files+nb's.  While I'm at it, I'll add the separate file-checking function and fully fledge it out when I make a PR for 'in-dockerland path checking'

ngreenwald:
Okay cool, while you're at, I think you can remove the load_axis keyword argument, and the logic that uses it. This is another leftover relic, we don't use that form of loading.

segmentation/utils/data_utils.py
alex-l-kong: Ah OK, I see.

ackagel:
Thanks for your feedback, Noah+Alex!  I migrated the previous functionality of `load_imgs_from_dirs` into `load_imgs_from_tree` and changed all references to the function accordingly.  Now `load_imgs_from_dirs` loads all images in a single directory into an xarray, setting the FOV coordinate equal to a delimit-able prefix of each respective file.

While I was at it, I also added a skeleton path-validator, which is functioning, but could definitely be fleshed out in a later pr.  

Also, I removed the `load_axis` argument from all `load_imgs_from_x` type functions and updated the descriptions accordingly.

Look forward to y'alls feedback on this one

segmentation/utils/data_utils_test.py
```diff
@@ -19,11 +20,36 @@ def _create_img_dir(temp_dir, fovs, imgs, img_sub_folder="TIFs"):
...
+
+    starts_out_of_scope = os.path.join(*valid_parts)
+
+    entirely_out_of_scope = '/not_a_real_directory/somewhere_else_not_here'
+
+    nonexistant = './not_a_real_directory'
+
+    data_utils.validate_paths(valid_path)
+
+    with pytest.raises(ValueError, match=r".*starts*"):
```
ngreenwald, line 43: Woah, I didn't know this was possible. Super cool!

segmentation/utils/data_utils.py
```diff
@@ -142,20 +170,113 @@ def load_imgs_from_dir(data_dir, img_sub_folder=None, fovs=None, imgs=None,
...
+
+    Inputs:
+        tiff_dir: path to directory containing segmentation labels as TIFs
+        output_dir: path to directory where the output will be saved. If None, no xr is saved
+        delimiter: character which separates fov-id prefix from the rest of the filename
+    Outputs:
+        Returns xarray and optionally saves it to output directory
+    """
+    im_xr = \
+        load_imgs_from_dir(data_dir=tiff_dir,
```
ngreenwald, line 265: We have the max line-length set to 99 for easier readability

ngreenwald:
Everything looks good. However, I noticed that in the deepcell_postprocessing script, there's a leftover call to load_imgs_from_dir

`image_data = data_utils.load_imgs_from_dir(data_dir=points_folder, img_sub_folder=img_sub_folder, 
                                                      fovs=current_points)`

It looks like this didn't throw an error for some reason? Can you figure out why?

ngreenwald:
If you could also add a brief comment in the pre-processing notebook telling people they have to put their files in the data folder within segmentation in order to work with docker that would be great

segmentation/utils/data_utils.py
ackagel: oops, didn't realize I went over there

segmentation/utils/data_utils_test.py
ackagel: oh yeah! pytest can catch/match exceptions with standard regex.  I was about to wrap an assert statement in a try/except structure, but then I found this

segmentation/utils/data_utils.py
ngreenwald: No, you went under. I think you can move it to one line and get rid of the return. The defaults are 60 chars I think, but I like 99 better.

ackagel: ohhh, okay cool.  It looks like it doesn't fit in one, but it'll fit in two now (there used to be more arguments so I think that's why I did it that way originally)

ackagel:
It looks like the postprocesing test passed because the test function was accidentally defined twice, with the intended test being overwritten by a test for the spatial analysis script.  I'll rename the latter to `test_example_spatial_analysis`.

https://github.com/angelolab/segmentation/blob/491a3d573dfee3f55f755af2dc6dfaf35c572750/segmentation/utils/notebooks_test.py#L24-L29

alex-l-kong:
Looks good Adam!

Issue #101, opened 2020-06-14
Improved testing for segmentation functions


ngreenwald:
The current segmentation core utils have only modest testing, and likely can also be further refactored. We should add correctness tests for extract_single_cell_data() at a minimum. 

This part of the pipeline is going to be expanded significantly in the near future, would be good to think about any rewriting we want to do now in preparation

PR #111, opened 2020-07-02
Add more segmentation testing functions

alex-l-kong:
Addressing issue #101 about adding a more complete set of testing functions to segmentation_utils_test.py. Currently, there are only three functions tested. It would be useful to add more to test the correctness of other functions such as combine_segmentation_masks.


ngreenwald:
Let me know before you start working on this: some of these functions will be going away in the near future so it won't be worth writing tests, some of them are going to stay with us forever so it's definitely worth extensively testing

PR #112, opened 2020-07-04
neighborhood_analysis

Jaiveers21:


segmentation/utils/spatial_analysis.py
```diff
@@ -192,3 +197,62 @@ def calculate_cluster_spatial_enrichment(all_data, dist_mats, fovs=None,
...
+
+    # initiate empty matrices for cell neighborhood data
+    cell_neighbor_counts = pd.DataFrame(np.zeros((all_data.shape[0], pheno_num + 2)))
+    cell_neighbor_freqs = pd.DataFrame(np.zeros((all_data.shape[0], pheno_num + 2)))
+
+    # Replace the first and second columns of cell_neighbor_counts with the fovs and cell labels respectively
+    cell_neighbor_counts[0] = all_data[fov_col]
+    cell_neighbor_counts[1] = all_data[cell_label_col]
+    cell_neighbor_freqs[0] = all_data[fov_col]
+    cell_neighbor_freqs[1] = all_data[cell_label_col]
```
alex-l-kong, line 244: This is admittedly nitpicky to the extreme, but if you wanted, you could shorten this to something like:

cell_neighbor_counts[[0, 1]] = all_data[[fov_col, cell_label_col]]
cell_neighbor_freqs[[0, 1]] = all_data[[fov_col, cell_label_col]]

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -250,3 +250,55 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+        cell_neighbor_freqs: matrix with phenotype frequencies of
+            counts per phenotype/total phenotypes for each cell
+        cell_count: current cell in analysis
+    Returns:
+        cell_neighbor_counts: matrix with phenotype counts per cell
+        cell_neighbor_freqs: matrix with phenotype frequencies of
+            counts per phenotype/total phenotypes for each cell
+        cell_count: current cell in analysis"""
+
+    cell_label_col = "cellLabelInImage"
```
alex-l-kong, line 275: Might be convenient to pass this as a parameter instead of redefining it here since you're using it in create_neighborhood_matrix already.

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -250,3 +250,55 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+
+    for j in list(fov_data.index):
+        # get specific cell
+        cell = fov_data.loc[j, cell_label_col]
+        # get all cell neighbors within threshold distance
+        cell_dist_mat = dist_matrix[int(cell - 1), :]
```
alex-l-kong, line 281: I'm pretty sure the variable "cell" is meant to always be at least 1, but it never hurts to have a check to make sure we're not accidentally passing in negative values which could lead to unintended behavior.

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -250,3 +250,55 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+        neighbor_labels = np.argwhere(cell_dist_mat_bin > 0) + 1
+        # filter out non-cellular objects (with no corresponding label in fov data)
+        neighbor_labels_cells = np.intersect1d(neighbor_labels, fov_data[cell_label_col])
+
+        # count phenotypes in cell neighbors
+        count_vec = np.zeros((1, pheno_num))
+        # Only include the neighbor labels that are cell labels
+        neighbor_inds = np.isin(fov_data[cell_label_col], neighbor_labels_cells)
+        # Get the phenotypes of the neighbor labels by index
+        pheno_vec = fov_data.iloc[neighbor_inds, 2]
```
alex-l-kong, line 296: Again, kind of nitpicky, but might be helpful to pass flowsom_col ("FlowSOM_ID") from create_neighborhood_matrix and use that to index the column rather than arbitrarily use the number "2".

segmentation/utils/spatial_analysis_test.py
```diff
@@ -14,7 +14,7 @@ def make_threshold_mat():
     return thresh
 
 
-def make_distance_matrix(enrichment_type):
+def make_distance_matrix(enrichment_type, analysis_type="spatial"):
```
alex-l-kong, line 17: Noticed that the analysis_type argument isn't used in the function make_distance_matrix nor make_expression_matrix. Just curious if you meant to add functionality for this later or if it was a parameter you ended up not needing.

segmentation/utils/spatial_analysis_test.py
```diff
@@ -63,7 +63,7 @@ def make_distance_matrix(enrichment_type):
         return dist_mat_neg
 
 
-def make_expression_matrix(enrichment_type):
+def make_expression_matrix(enrichment_type, analysis_type="spatial"):
```
alex-l-kong, line 66: See comment above about make_distance_matrix.

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -250,3 +250,55 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+        for k in range(1, pheno_num):
+            count_vec[0, k - 1] = sum(pheno_vec == k)
```
alex-l-kong, line 298: Is it possible to broadcast this operation? Using a list comprehension (not broadcasting), I was able to do: 

count_vec[0, 0:(pheno_num - 1)] = [sum(pheno_vec == k) for k in range(1, pheno_num).

List comprehensions can often produce marginal speed improvements, and on average, I found my code running 0.01s faster per iteration using a comprehension.

It's not the most significant time improvement I can find but it could be a start because when running your code on the test example_expression_matrix.csv and segmentation_labels.xr, create_neighborhood_matrix takes a long time to run.

alex-l-kong:
Overall, a great first pass Jay! Outside of some minor changes here and there, I think the major thing to look at right now is speed. Granted, creating a neighborhood matrix may just inherently be a slow operation due to the number of iterations involved, but I still think it's worth looking into what can be done to speed up the code. My concern is that for much larger .csv files with more phenotype codes, the process could potentially take a much longer time to run. Of course, if there's really nothing else that can be done, don't kill yourself over it. Just something to think about.

For the record, it might be worth looking into a library like tqdm which is a progress bar library which allows the user to know how far along a process is to completion. This would especially be helpful when we get a script converted into a Jupyter notebook. If this is an inherently slow process, then it would be nice to let the user know how far along the process is to completion without having them guess how much longer it will take if creating the neighborhood matrix is really that slow of a process.

segmentation/utils/spatial_analysis.py
```diff
@@ -192,3 +197,62 @@ def calculate_cluster_spatial_enrichment(all_data, dist_mats, fovs=None,
         stats_xr = spatial_analysis_utils.calculate_enrichment_stats(close_num, close_num_rand)
         stats.loc[fovs[i], :, :] = stats_xr.values
     return values, stats
+
+
+def create_neighborhood_matrix(all_data, dist_matrices, fovs=None, distlim=50):
+    """Calculates the number of neighbor phenotypes for each cell
+
+        Args:
+            all_data: data for the all fov
```
ackagel, line 206: it'd be great if you could describe the `all_data` argument a bit more.  is it an xarray?  if so, does it have an assumed shape/structure?  those kinds of things

segmentation/utils/spatial_analysis.py
```diff
@@ -192,3 +197,62 @@ def calculate_cluster_spatial_enrichment(all_data, dist_mats, fovs=None,
...
+        Returns:
+            cell_neighbor_counts: matrix with phenotype counts per cell
+            cell_neighbor_freqs: matrix with phenotype frequencies of
+                counts per phenotype/total phenotypes for each cell
+            cell_count: current cell in analysis"""
+
+    # Setup input and parameters
+    fov_col = "SampleID"
+    flowsom_col = "FlowSOM_ID"
+    cell_label_col = "cellLabelInImage"
```
ackagel, line 220: I don't have the full context for this functions use-cases, but this feels like hard-coding.  If it's possible for `all_data` to have different index names, maybe add some argument(s) to the function which takes these as defaults?  Could be condensed into one argument as a dictionary, or could be separate args (basically the same thing considering the splay(*) operator exists)

segmentation/utils/spatial_analysis.py
ackagel: also, you don't have to initiate `cell_neighbor_freqs` on line 238.  after you fill `cell_neighbor_counts` you can just copy it into a new `cell_neighbor_freqs`:
```
cell_neighbor_freqs = cell_neighbor_counts.copy()
```

sidenote, dataframe copies are deep by default, but if you want to be extra-extra sure, or make it so that people can see that in the code, you can call it with `.copy(deep=True)`

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -250,3 +250,55 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+        pheno_num: number of different cell phenotypes
+        cell_neighbor_counts: matrix with phenotype counts per cell
+        cell_neighbor_freqs: matrix with phenotype frequencies of
+            counts per phenotype/total phenotypes for each cell
+        cell_count: current cell in analysis
+    Returns:
+        cell_neighbor_counts: matrix with phenotype counts per cell
+        cell_neighbor_freqs: matrix with phenotype frequencies of
+            counts per phenotype/total phenotypes for each cell
+        cell_count: current cell in analysis"""
```
ackagel, line 273: `cell_neighbor_counts` and `cell_neighbor_freqs` are modified in place, which means they shouldn't need to be returned.

however, `cell_count` is an immutable int (no in-place modification), so its current implementation can stay the same

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -250,3 +250,55 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+        neighbor_inds = np.isin(fov_data[cell_label_col], neighbor_labels_cells)
+        # Get the phenotypes of the neighbor labels by index
+        pheno_vec = fov_data.iloc[neighbor_inds, 2]
+        for k in range(1, pheno_num):
+            count_vec[0, k - 1] = sum(pheno_vec == k)
+        # add to neighborhood matrices
+        cell_neighbor_counts.loc[cell_count, 2:] = count_vec[0, :]
+        cell_neighbor_freqs.loc[cell_count, 2:] = (count_vec[0, :] / len(neighbor_labels_cells))
+        # update cell counts
+        cell_count += 1
```
ackagel, line 303: You can actually do this with no for loop.
```
# remove non-cell2cell distances
cell_dist_mat = np.take(dist_matrix, fov_data[cell_label_col] - 1, 0)
cell_dist_mat = np.take(cell_dist_mat, fov_data[cell_label_col] - 1, 1)

# binarize distance matrix
cell_dist_mat_bin = np.zeros(cell_dist_mat.shape)
cell_dist_mat_bin[cell_dist_mat < dist_lime] = 1

# get num_neighbors for freqs
num_neighbors = np.sum(cell_dist_mat_bin, axis=0)

# create the 'phenotype has cell?' matrix, excluding non cell-label rows (should match up w/ dist_matrix)
pheno_has_cell = pd.get_dummies(fov_data.iloc[fov_data[cell_label_col], 2]).to_numpy().T

# dot binarized 'is neighbor?' matrix with pheno_has_cell to get counts
counts = pheno_has_cell.dot(dist_mat).T

# compute freqs with num_neighbors (one of these or both might need to be transposed)
freqs = counts/num_neighbors

# add to neighbor counts + freqs
cell_neighbor_counts.loc[cell_count:(cell_count+cell_dist_mat.shape[0]), 2:] = counts
cell_neighbor_freqs.loc[cell_count:(cell_count+cell_dist_mat.shape[0]), 2:] freqs
```

there are probably some array-shape related errors in the above sample/snippit (particularly with transposes, etc), but this approach, in principle, should work and should significantly speed the function up.  

basically, it uses `fov_data` to construct a binarized, "does cell k, have phenotype j?" matrix, P_jk, for all labelled-cells.  just taking one cell's binarized "is cell k my neighbor?" vector, d, it's clear that the matrix (dot in np) product P_jk*d_k = c_j is the vector containing the number of neighbors of each phenotype, j.  luckily, numpy's `dot` automatically repeats this over for each column d_k of `cell_dist_mat_bin`, which yields the counts for each cell.  
The rest is just filing away the results into `cell_neighbor_counts` and doing a little extra hoop-jumping for the frequencies.
Feel free to slack me if any of that wasn't clear, or if I'm mondo wrong

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -208,3 +214,30 @@ def test_calculate_enrichment_stats():
...
+    fov_data.iloc[:, 30] = 8
+    fov_data = fov_data[[fov_col, cell_label_col, flowsom_col]]
+    pheno_num = len(fov_data[flowsom_col].drop_duplicates())
+
+    cell_neighbor_counts = pd.DataFrame(np.zeros((fov_data.shape[0], pheno_num + 2)))
+    cell_neighbor_freqs = pd.DataFrame(np.zeros((fov_data.shape[0], pheno_num + 2)))
+    cell_neighbor_counts[0] = fov_data[fov_col]
+    cell_neighbor_counts[1] = fov_data[cell_label_col]
+    cell_neighbor_freqs[0] = fov_data[fov_col]
+    cell_neighbor_freqs[1] = fov_data[cell_label_col]
```
ackagel, line 236: See Alex and I's comments from earlier

Jaiveers21:
Can you guys check this out? Other than this I think all that is left is to add touch up comments and the implementation in the jupyter notebook

segmentation/utils/spatial_analysis.py
```diff
@@ -192,3 +197,72 @@ def calculate_cluster_spatial_enrichment(all_data, dist_mats, fovs=None,
...
+    for i in range(len(fovs)):
+        # Subsetting expression matrix to only include patients with correct label
+        patient_ids = all_data.iloc[:, 0] == fovs[i]
+        fov_data = all_data[patient_ids]
+
+        # Subset the distance matrix dictionary to only include the distance matrix for the correct point
+        dist_matrix = dist_matrices[str(fovs[i])]
+
+        # Get cell_neighbor_counts and cell_neighbor_freqs for points
+        spatial_analysis_utils.compute_neighbor_counts(
```
alex-l-kong, line 265: This might just be a pet peeve of mine, but I personally feel cell_neighbor_counts and cell_neighbor_freqs shouldn't be updated like this (as in, passing them into a function, and using the mutability properties of the parameters to update them in a separate function, in this case, compute_neighbor_counts). Let's see what Adam and Noah have to say, my suggestion is to initialize cell_neighbor_counts and cell_neighbor_freqs inside compute_neighbor_counts and having that function return those computed values.

segmentation/utils/spatial_analysis_test.py
```diff
@@ -287,3 +296,18 @@ def test_calculate_cluster_spatial_enrichment():
...
+    dist_mat_pos = make_distance_matrix("positive")
+
+    counts, freqs = spatial_analysis.create_neighborhood_matrix(all_data_pos, dist_mat_pos, distlim=51)
+
+    # Test the counts values for both fovs
+    assert (counts.loc[:19, "Pheno3":"Pheno2"] == 8).all().all()
+    assert (counts.loc[:19, "Pheno1"] == 4).all()
+
+    assert (counts.loc[80:99, "Pheno1":"Pheno3"] == 8).all().all()
+    assert (counts.loc[80:99, "Pheno2"] == 4).all()
```
alex-l-kong, line 313: Probably a total noob for asking this, but is there a reason we're not testing for negative and no enrichment matrices?

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -208,3 +217,35 @@ def test_calculate_enrichment_stats():
...
+
+    cell_neighbor_counts[[0, 1]] = fov_data[[fov_col, cell_label_col]]
+    cell_neighbor_freqs[[0, 1]] = fov_data[[fov_col, cell_label_col]]
+
+    # Rename the columns to match cell phenotypes
+    cols = [fov_col, cell_label_col] + list(pheno_titles)
+    cell_neighbor_counts.columns = cols
+    cell_neighbor_freqs.columns = cols
+
+    spatial_analysis_utils.compute_neighbor_counts(
```
alex-l-kong, line 246: This test may need to be slightly altered depending on what Adam and Noah think about my earlier suggestion about changing the way cell_neighbor_counts and cell_neighbor_freqs are computed in create_neighborhood_matrix.

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -250,3 +250,57 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+    pheno_titles = fov_data[cell_type_col].drop_duplicates()
+
+    # remove non-cell2cell distances
+    cell_dist_mat = np.take(dist_matrix, fov_data[cell_label_col] - 1, 0)
+    cell_dist_mat = np.take(cell_dist_mat, fov_data[cell_label_col] - 1, 1)
+
+    # binarize distance matrix
+    cell_dist_mat_bin = np.zeros(cell_dist_mat.shape)
+    cell_dist_mat_bin[cell_dist_mat < distlim] = 1
+    # cell_dist_mat_bin[cell_dist_mat == 0] = 0 <- cell counts itself as neighbor if this line is not included
```
alex-l-kong, line 286: Remove this commented line if not needed.

segmentation/utils/spatial_analysis.py
```diff
@@ -20,6 +20,11 @@
 # pheno = pd.read_csv("/Users/jaiveersingh/Downloads/CellType_GlobalSpatialEnrichment/cellpheno_numkey.csv")
 # dist_mat = np.asarray(pd.read_csv("/Users/jaiveersingh/Documents/MATLAB/distancesMat5.csv", header=None))
 
+# Erin's Data Inputs-Neighborhood
+# dist_matrix = np.asarray(pd.read_csv("/Users/jaiveersingh/Documents/MATLAB/distancesMat5.csv", header=None))
+# all_data = pd.read_csv("/Users/jaiveersingh/Desktop/granA_cellpheno_CS-asinh-norm_matlab_revised.csv")
+# pheno_key = pd.read_csv("/Users/jaiveersingh/Downloads/CreateNeighborMatrix/cellpheno_numkey.csv")
+
```
alex-l-kong, line 27: Recommend removing these comments prior to merging with master unless it's absolutely necessary to keep them around for reference sake.

segmentation/utils/spatial_analysis.py
ackagel: Updating mutables this way is generally considered good practice for a few reasons.  First, it guarantees that any changes to `cell_neighbor_counts` inside `compute_neighbor_counts` are never 'unnoticed' (e.g not overwritten on return, but used later).  Second, it should be a bit faster to directly update than to copy (shallowly or deeply) and then update with return values.  On the other hand, returning mutable objects should be done when the original input may be needed afterwards, and when you can guarantee that no changes to the original were made within the function.  (For a C/C++ analogy, think of it like passing by reference, versus passing by constant reference)

An examples of this in standard python is `random.shuffle(a)` which doesn't return a shuffled copy of `a` but directly updates it.

I recommend either leaving it as is, or (super optional) changing the order of arguments so that the mutables (`counts` and `freqs`) are the first two (could make it more clear that the other args are more like parameters, not true inputs).

segmentation/utils/spatial_analysis_utils.py
ackagel: I think it's unclear if it's needed atm.  I'd leave it until it's decided, or possibly have an argument `count_self=True` which uses it.  Although, if the former, reformatting it so that the arrow'd comment is another comment above it seems more appropriate:
```
# cell counts itself as neighbor if this line is not included
# cell_dist_mat_bin[cell_dist_mat == 0] = 0
```

segmentation/utils/spatial_analysis.py
ackagel: Should be removed once comparing with old output is no longer needed

segmentation/utils/spatial_analysis.py
```diff
@@ -192,3 +197,72 @@ def calculate_cluster_spatial_enrichment(all_data, dist_mats, fovs=None,
...
+    # Replace the first and second columns of cell_neighbor_counts with the fovs and cell labels respectively
+    cell_neighbor_counts[[0, 1]] = all_data[[fov_col, cell_label_col]]
+    cell_neighbor_freqs[[0, 1]] = all_data[[fov_col, cell_label_col]]
+
+    # Rename the columns to match cell phenotypes
+    cols = [fov_col, cell_label_col] + list(pheno_titles)
+    cell_neighbor_counts.columns = cols
+    cell_neighbor_freqs.columns = cols
+
```
ackagel, line 255: The `cell_neighbor_freqs = cell_neighbor_counts.copy(deep=True)` can be moved under this code.  That way, lines 249 and 254 can be removed.

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -250,3 +250,57 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+    # compute freqs with num_neighbors
+    freqs = counts.T / num_neighbors
+
+    # convert to pandas df and assign columns to phenos
+    countsdf = pd.DataFrame(counts, columns=pheno_titles)
+    freqsdf = pd.DataFrame(freqs.T, columns=pheno_titles)
+
+    # add to neighbor counts + freqs for only the matching phenotypes between the fov and the whole dataset
+    cell_neighbor_counts.loc[fov_data.index, countsdf.columns] = counts
+    cell_neighbor_freqs.loc[fov_data.index, freqsdf.columns] = freqs.T
```
ackagel, line 306: I think we can bypass making new df's by indexing with `.loc[fov_data.index, pheno_titles]` directly.  I could be wrong tho, I don't work with df's too often.

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -208,3 +217,35 @@ def test_calculate_enrichment_stats():
...
+    cell_neighbor_freqs = cell_neighbor_counts.copy(deep=True)
+
+    cell_neighbor_counts[[0, 1]] = fov_data[[fov_col, cell_label_col]]
+    cell_neighbor_freqs[[0, 1]] = fov_data[[fov_col, cell_label_col]]
+
+    # Rename the columns to match cell phenotypes
+    cols = [fov_col, cell_label_col] + list(pheno_titles)
+    cell_neighbor_counts.columns = cols
+    cell_neighbor_freqs.columns = cols
+
```
ackagel, line 245: See earlier comment about moving `cell_neighbor_freqs` copy under this code block.

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -250,3 +250,57 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
     return stats_xr
+
+
+def compute_neighbor_counts(fov_data, dist_matrix, distlim,
+                            cell_neighbor_counts, cell_neighbor_freqs,
+                            cell_label_col="cellLabelInImage", cell_type_col="cell_type"):
+    """Calculates the number of neighbor phenotypes for each cell. The cell counts itself as a neighbor.
+
+    Args:
+        fov_data: data for the current fov
```
ngreenwald, line 261: This description should be expanded to explain what the data actually is

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -250,3 +250,57 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+    pheno_has_cell = pd.get_dummies(fov_data.iloc[:, 2]).to_numpy().T
+
+    # dot binarized 'is neighbor?' matrix with pheno_has_cell to get counts
+    counts = pheno_has_cell.dot(cell_dist_mat_bin).T
+
+    # compute freqs with num_neighbors
+    freqs = counts.T / num_neighbors
+
+    # convert to pandas df and assign columns to phenos
+    countsdf = pd.DataFrame(counts, columns=pheno_titles)
```
ngreenwald, line 301: Given that we're generating an FOV-specific list of values to update, I think it makes more sense to return these values and update them outside of the function than to do it inside the function. There doesn't appear to be any efficiency gains from doing it this way, and I think it will be much easier to understand if we return and update

segmentation/utils/spatial_analysis.py
ngreenwald: My main problem with this is that it requires you to go inside of compute_neighbor_counts to figure out that cell_neighbor_counts and cell_neighbor_freqs are secretly being updated.

Adam, thanks for the description, that's good to know. My understanding then is that because we know nothing else is going to be changes these mutables, it would be acceptable to return and update? 

Based on the way the function is currently doing the updating, it will require almost no changes to make this happen (see my comment inside of the helper function)

ackagel: when you put it like that, I think the best way to do this is to return the `counts` and `freqs` matrices and then update the `cell_neighbor_***` df's outside the function.  This way, those df's don't even need to be passed into the function, best practices aren't offended, and the code is mondo legible.

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -208,3 +202,38 @@ def test_calculate_enrichment_stats():
...
+    cell_neighbor_counts.columns = cols
+
+    cell_neighbor_freqs = cell_neighbor_counts.copy(deep=True)
+
+    counts, freqs = spatial_analysis_utils.compute_neighbor_counts(
+        fov_data, dist_matrix, distlim)
+
+    # add to neighbor counts + freqs for only the matching phenotypes between the fov and the whole dataset
+    cell_neighbor_counts.loc[fov_data.index, pheno_titles] = counts
+    cell_neighbor_freqs.loc[fov_data.index, pheno_titles] = freqs.T
```
ackagel, line 235: see earlier comment

segmentation/utils/spatial_analysis.py
```diff
@@ -195,3 +180,78 @@ def calculate_cluster_spatial_enrichment(all_data, dist_mats, fovs=None,
...
+            cell_label_col: column with the cell labels (Default is cellLabelInImage)
+            cell_type_col: column with the cell types (Default is cell_type)
+        Returns:
+            cell_neighbor_counts: matrix with phenotype counts per cell
+            cell_neighbor_freqs: matrix with phenotype frequencies of
+                counts per phenotype/total phenotypes for each cell
+            cell_count: current cell in analysis"""
+
+    # Setup input and parameters
+
```
alex-l-kong, line 208: Remove blank line here between comment and code.

Jaiveers21:
This is the final pass with the #ToDo for the cell2cell until our own inputs are generated

segmentation/utils/spatial_analysis.py
```diff
@@ -195,3 +180,77 @@ def calculate_cluster_spatial_enrichment(all_data, dist_mats, fovs=None,
...
+    # Rename the columns to match cell phenotypes
+    cols = [fov_col, cell_label_col] + list(pheno_titles)
+    cell_neighbor_counts.columns = cols
+
+    cell_neighbor_freqs = cell_neighbor_counts.copy(deep=True)
+
+    for i in range(len(fov_list)):
+        # Subsetting expression matrix to only include patients with correct label
+        patient_idx = all_data.iloc[:, 0] == fov_list[i]
```
ngreenwald, line 239: Correct me if I'm wrong, but I believe patient and FOV are actually referring to the same thing here? If so, let's change all references to patients in variable names to be FOVs so that we have consistency and the code is easier to follow

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -253,3 +253,48 @@ def calculate_enrichment_stats(close_num, close_num_rand):
     dims = ["stats", "rows", "cols"]
     stats_xr = xr.DataArray(stats_data, coords=coords, dims=dims)
     return stats_xr
+
+
+def compute_neighbor_counts(fov_data, dist_matrix, distlim, self_neighbor=True,
+                            cell_label_col="cellLabelInImage"):
+    """Calculates the number of neighbor phenotypes for each cell. The cell counts itself as a neighbor.
```
ngreenwald, line 260: We don't need to state that the cell counts itself as a neighbor in the main summary docstring if it's specified with self_neighbor argument description

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -253,3 +253,48 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+    Args:
+        fov_data: data for the current fov, including the cell labels, cell phenotypes, and cell phenotype ID
+        dist_matrix: cells x cells matrix with the euclidian
+            distance between centers of corresponding cells
+        distlim: threshold for spatial enrichment distance proximity
+        self_neighbor: If true, cell counts itself as a neighbor in the analysis.
+        cell_label_col: Column name with the cell labels
+    Returns:
+        counts: data array with phenotype counts per cell
+        freqs: data array with phenotype frequencies of
```
alex-l-kong, line 271: Change this to freqs.T, since we're returning the transpose of freqs.

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -253,3 +253,48 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+            counts per phenotype/total phenotypes for each cell
+        cell_count: current cell in analysis"""
```
alex-l-kong, line 273: Remove, because we don't return cell_count anymore.

segmentation/utils/spatial_analysis.py
```diff
@@ -195,3 +180,76 @@ def calculate_cluster_spatial_enrichment(all_data, dist_mats, fovs=None,
...
+            distlim: cell proximity threshold. Default is 50.
+            fov_col: column with the cell fovs (Default is SampleID)
+            flowsom_col: column with the cell phenotype IDs (Default is FlowSOM_ID)
+            cell_label_col: column with the cell labels (Default is cellLabelInImage)
+            cell_type_col: column with the cell types (Default is cell_type)
+        Returns:
+            cell_neighbor_counts: matrix with phenotype counts per cell
+            cell_neighbor_freqs: matrix with phenotype frequencies of
+                counts per phenotype/total phenotypes for each cell
+            cell_count: current cell in analysis"""
```
alex-l-kong, line 204: Remove, because we no longer return cell_count.

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -253,3 +253,48 @@ def calculate_enrichment_stats(close_num, close_num_rand):
...
+
+    # TODO remove non-cell2cell lines after our own inputs for functions are created
+    # remove non-cell2cell distances
```
alex-l-kong, line 276: I think this comment should be updated because as @ngreenwald and @ackagel discussed below, it's not entirely clear what non-cell2cell distances are being removed and how fov_data[cell_label_col] - 1 specifically indexes rows and columns that are cell2cell distances. This is especially the case when you look at calc_dist_matrix in spatial_analysis_utils where dist_matrix is created. It doesn't seem like that function includes any non-cell2cell distances, at least how I understood it.

In summary, I think it would be helpful to clarify further what non-cell2cell distances means in the context of this function and the function that calls it (create_neighborhood_matrix).

segmentation/utils/spatial_analysis_utils.py
ackagel: we are actually just returning freqs.  It looks like we're returning `freqs.T` but it's easier to broadcast the division for counts -> freqs within the transposed space and returned the transpose of that result (although maybe `num_neighbors` could be transposed instead?).  Regardless, saying that we're returning `freqs.T` makes it seem like the frequency output has to be transposed outside of the function.  I think it's best to leave it as is imo.

segmentation/utils/spatial_analysis_utils.py
ackagel: Changing the comment to something like 'refine distance matrix to only cover cell ids in fov_data' could be better.  Regardless, I agree that non-cell2cell is pretty non-specific.  idk if that needs to be changed right now, though.  Thoughts @ngreenwald ?

ngreenwald: I believe what's happening is the following. The distance matrix has the first cell ID with index 0, and doesn't include background as a row. However, the cell IDs start at  1, because background is assigned the value of 0. So we have to subtract 1 to convert from cell_IDs to row indices. 

There are certain Cell_IDs which are filtered out, because they aren't cells. This is an artifact of the previous segmentation which included non-cell objects that were identified by heuristic rules. However, it appears that maybe the distance matrix was created with all objects, not just cell objects? Not 100% sure about this part, but this isn't relevant moving forward because we won't have this issue. However, we're going to wait to remove this until we have our own fleshed out test functions, since right now we want to be able to check against the matlab output

segmentation/utils/spatial_analysis_utils.py
alex-l-kong: Ah OK, thanks for clarifying.

Jaiveers21:
@alex-l-kong @ngreenwald @ackagel 

I have added the suggestions above

ngreenwald:
Why do you hate hitting the request review button. What did it ever do to you? 
If you don't click it, I just have to click it myself in order to review the changes

Jaiveers21:
Haha oops, I will use the button

> Why do you hate hitting the request review button. What did it ever do to you?
> If you don't click it, I just have to click it myself in order to review the changes

Issue #113, opened 2020-07-04
Bad Path Explainer


ngreenwald:
Given that people have an incredible propensity to generate bad paths, I think it will be worthwhile putting together a function to give interpretable outputs when this happens. 

Adam's idea about a docker-specific warning when people specify a path outside of the mounted volume got me thinking about what such a helper function would look like. 

Currently, when a path doesn't exist, we (usually) at least tell them the reason the function failed was because of a bad path. However, figuring out why the path was bad involves careful staring at the string. 

I think a nice feature would be to add a function that identifies the the most recent common ancestor of the path the user supplied that actually exists. This will help to identify where the typo was. This can be part of a function that does a range of sanity checking for paths, including the docker volume check.

ackagel:
Just added an early version of this to PR #110.  I imagine it could be fleshed out more and added to more instances of general file-loading, which I can do in a separate PR specifically for this issue

ngreenwald:
Here's what the more fleshed out version of this might look like. 

If path is not valid:
    if path does not start with /data:
         Raise an error about non-docker paths
    if path does start with /data:
         loop through parts of the path, and identify the first part that does not exist
         print out that part, as a guide to the user for where the invalid path originated. 


I'm not sure I understand the logic of using pathlib.cwd() in the current implementation. What does that tell us about whether a path is valid or not?

ackagel:
I'm just using `pathlib.Path.cwd()` to get the top-level path to the segmentation folder, since the cwd should never be outside of it.  Currently, when a path doesn't exist, it indexes from its highest folder until it (possibly) reaches 'segmentation', awaiting a mismatch between the path's head and the cwd's head.  If none is found, then the path is inscope, but the file/directory doesn't exist (this could be fleshed out as above regardless).  If a mismatch occurs, there are two cases.  Either 'segmentation' is in the path structure somewhere, meaning this is most likely an absolute path error (path starts out of scope), or 'segmentation' is not in the path structure, meaning the directory/file is totally outside of docker's scope.

That being said, just checking that the path starts with `../data` (or maybe `./data` ?) like your pseudo-code above is probably better, since I think we're encouraging users to put things **only** in that folder, and to only use the codebase from a notebook within `scripts`.  I used the cwd approach for the case where the codebase is used in a notebook or a script with unknown working directory, since the relative path to `data` (or any folder) is ambiguous.

ngreenwald:
Got it. Inside of the docker, the top level path starts where the docker is located:
[...]

PR #116, opened 2020-07-05
Mount code to docker image

### LINKED ISSUE #115: Mount code in docker
We should update the docker starting script to mount the code in addition to data so that the image doesn't need to be rebuilt to get latest changes

ngreenwald:
An issue that has come up repeatedly is people not realizing they have to rebuild the docker image when they git pull. 

To address this, I modified the start_docker.sh script to mount the code inside the container, so any updates from git are propagated automatically inside. 

I also modified the installation process so that the repo is installed via pip. This means we can directly import the package, rather than having to append the relative path to the repo onto the system path in the example scripts, which was our previous hack. 

Closes #115


start_docker.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 docker run -it \
   -p 8888:8888 \
+  -v $PWD/segmentation:/usr/local/lib/python3.6/site-packages/segmentation \
   -v $PWD/scripts:/scripts \
   -v $PWD/data:/data \
-  $USER/segmentation:latest
+  segmentation:latest
```
alex-l-kong, line 7: I had someone come to me with a use case where they needed to upload a local file, which is not supported by this .sh script (needed an additional -v flag for that). In the future, do we expect more use cases like this, or can we just safely assume that this is all the user needs for the most part?

scripts/Deepcell_Postprocessing.ipynb
```diff
@@ -13,17 +13,16 @@
...
-    "sys.path.append(\"../\")\n",
+    "import matplotlib.pyplot as plt\n",
+    "import numpy as np\n",
+    "import pandas as pd\n",
+    "import skimage.io as io\n",
+    "import xarray as xr\n",
     "\n",
-    "from segmentation.utils import data_utils, segmentation_utils\n",
-    "import xarray as xr"
+    "from segmentation.utils import data_utils, segmentation_utils"
```
alex-l-kong, line 25: Any particular reason to change the order of these imports, just code cleanup I assume?

scripts/Deepcell_Preprocessing_MIBItiff.ipynb
```diff
@@ -87,7 +84,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.3"
+   "version": "3.6.10"
```
alex-l-kong, line 87: Out of curiosity, are we planning on sticking with Python 3.6 as outlined in the Dockerfile for the near future?

start_docker.sh
ngreenwald: I think for now it makes sense to have all of the notebooks that someone wants to run be located within the scope of the docker, but if it turns into a repeated issue we can certainly address that.

scripts/Deepcell_Postprocessing.ipynb
ngreenwald: Yup, just alphabetizing and separating based on import type

scripts/Deepcell_Preprocessing_MIBItiff.ipynb
ngreenwald: Yes, I think we'll stick with 3.6 until we find a package that requires 3.7. But we can add 3.7 to Travis

start_docker.sh
ackagel: I think adding a blurb to the readme to the effect that all relevant data/files/etc should be placed in the `data` folder should be sufficient.

Dockerfile
```diff
@@ -6,13 +6,15 @@ RUN apt-get update && apt-get install -y gcc
...
+RUN pip install -r /opt/segmentation/requirements.txt
 
 # copy the scripts over
-COPY segmentation /segmentation
+COPY segmentation /opt/segmentation/segmentation
 COPY scripts /scripts
 
+# Install the package via setup.py
+RUN pip install /opt/segmentation
+
```
ackagel, line 18: I think that this might affect the "in docker's scope" checking being added in PR #110 (see issue #113), but I'm not 100% sure of that

ngreenwald: I don't think this will affect that. Changing the folder that segmentation is located in within the docker shouldn't affect the logic about relative paths, right? Your code just looks at the cwd, which looks like it's relative to the docker container 
image

ngreenwald:
Oooh fancy, Github now automatically removes approvals if you push new commits

PR #119, opened 2020-07-07
Plotting functions

### LINKED ISSUE #117: Improvements for plotting cohort composition
Hey guys, here's the summary of what Saeed, Mike and I thought would be some next steps for the plotting code:

- [x] Bars should be sorted either by total counts, counts of a specified subpopulation, or based on hierarchical clustering
- [x] The order of individual stacks should be most common to least common from the bottom of the stack to the top
- [x] Color map for individual stacks should be able to be manually specified by the user
- [x] Function should be able to handle any categorical column in CSV for either the “patient” column or the “population” column

ngreenwald:
Reopening PR from new branch. Closes #117


segmentation/utils/visualize.py
```diff
@@ -18,3 +18,84 @@ def visualize_z_scores(z, pheno_titles):
...
     sns.set(font_scale=.7)
     sns.clustermap(zplot, figsize=(8, 8), cmap="vlag")
+    
+def newDict(names):
+    result = {}
+    for name in names:
+        result[name] = [0]
+    return result
+
+def visualize_distribution_of_cell_count(df, id_col_name, cell_col_name):
```
ackagel, line 28: Please add multiline comment/description of function.  It should contain a blurb about what the function does, its inputs + their format, and its outputs (returns/write out to disk) + their formats.  Examples can be found in most of the other scripts in the `utils` folder.

Also, you should change `df` to a more descriptive name for better readability.

segmentation/utils/visualize.py
```diff
@@ -18,3 +18,84 @@ def visualize_z_scores(z, pheno_titles):
...
+    print(ids)
+    cell_names = df[cell_col_name].value_counts().index
+    cell_names = cell_names.tolist()  
+
+    unique_ids = []
+    for item in ids:
+        if type(item) is float or type(item) is int:
+            unique_ids.append(item)
+
+    unique_ids = [x for x in unique_ids if x == x]
```
ackagel, line 39: I believe a python set automatically only store unique entries.  I don't have too much context for this but aren't id's always floats/ints?  Also, I'm pretty sure line 39 does nothing.  Long story short, I think these lines are unneeded

segmentation/utils/visualize.py
```diff
@@ -18,3 +18,84 @@ def visualize_z_scores(z, pheno_titles):
...
+
+    print(unique_ids)
```
ackagel, line 41: this is good for debugging but generally shouldn't be in pushed code

segmentation/utils/visualize.py
```diff
@@ -18,3 +18,84 @@ def visualize_z_scores(z, pheno_titles):
     zplot = pd.DataFrame(z, columns=pheno_titles, index=pheno_titles)
     sns.set(font_scale=.7)
     sns.clustermap(zplot, figsize=(8, 8), cmap="vlag")
+    
+def newDict(names):
+    result = {}
+    for name in names:
+        result[name] = [0]
+    return result
```
alex-l-kong, line 26: It's good that you're defining functions for specific tasks, but this might be a bit overkill. This is a pretty standard operation in Python, and it can be simplified into one line as well: result = {name: [0] for name in names}.

It's also good to choose a better name than "name" because it's too similar to "names" which you are iterating over.

segmentation/utils/visualize.py
```diff
@@ -18,3 +18,84 @@ def visualize_z_scores(z, pheno_titles):
...
+
+def visualize_distribution_of_cell_count(df, id_col_name, cell_col_name):
+    ids = set(df[id_col_name])
+    print(ids)
```
alex-l-kong, line 30: See Adams comment about removing print statements from pushed code.

segmentation/utils/visualize.py
alex-l-kong: I second Adam on this. 

On a side note, line 39 doesn't work even if you actually had a list with duplicates. That's because you're comparing each element to itself, and such a statement will always be true for integer/float values. You can call set again on unique_ids to properly ensure no duplicates, but in this case, you don't have to cause you pulled items from a set (the id variable) to put into unique_ids.

segmentation/utils/visualize.py
```diff
@@ -18,3 +18,84 @@ def visualize_z_scores(z, pheno_titles):
...
+    df_stacked = pd.DataFrame(newDict(cell_names))
+
+    first = True
+    for id in unique_ids:
+
+        df_rows = df.loc[df["PatientID"] == id]
+        row_cell_types = df_rows["cell_type"].value_counts()
+
+        if first:
+            df_stacked = pd.DataFrame([row_cell_types.values], columns=row_cell_types.index)
```
alex-l-kong, line 52: This line is a bit confusing because it seems like you've already defined df_stacked to be something on line 43, but in this line, you end up redefining df_stacked to be something else which overwrites what your previous value assigned here.

segmentation/utils/visualize.py
```diff
@@ -18,3 +18,84 @@ def visualize_z_scores(z, pheno_titles):
...
+             df_stacked = pd.DataFrame([row_cell_types.values], columns=row_cell_types.index)
+            first = False
+        else:
+            df_stacked = pd.concat([df_stacked, (pd.DataFrame([row_cell_types.values], columns=row_cell_types.index))], axis=0, ignore_index=True)
+
+    df_stacked.plot.bar(stacked=True, legend = False)
+    plt.legend(loc='center left', bbox_to_anchor=(1.0, 0.5))
+    plt.xlabel("Patient ID")
+    plt.ylabel("Cell Proportion")
+    plt.title("Distribution of Cell Count Proportion in Patients")
```
alex-l-kong, line 93: This function looks very similar to visualize_distribution_of_cell_count. Unless you expect the functionality of these functions to change drastically over time, I'd recommend combining much of the shared logic into one function.

segmentation/utils/visualize_test.py
```diff
@@ -10,3 +11,17 @@ def test_visualize_z_scores():
...
+def test_visualize_cells():
+    rand_type = []
+    ids = []
+    for x in range(0,1000):
+        rand_type.append(chr(random.randint(0,25) + 97))
+        ids.append(random.randint(1,5))
+    print(rand_type, ids)
+    rand_dict = {"PatientID": ids, "cell_type": rand_type}
+    df = pd.DataFrame.from_dict(rand_dict)
+    print(df)
```
alex-l-kong, line 24: See comments about not pushing print statements to pushed code.

segmentation/utils/visualize_test.py
```diff
@@ -10,3 +11,17 @@ def test_visualize_z_scores():
     # Assign random phenotype titles
     pheno_titles = [chr(i) for i in range(ord('a'), ord('z') + 1)]
     plot = visualize.visualize_z_scores(z, pheno_titles)
+    
+def test_visualize_cells():
+    rand_type = []
+    ids = []
+    for x in range(0,1000):
```
alex-l-kong, line 18: Just nitpicking here, it's OK to leave out the 0 in range in this case and just use range(1000). If only one value is specified, it is assumed that you start from 0.

segmentation/utils/visualize_test.py
```diff
@@ -10,3 +11,17 @@ def test_visualize_z_scores():
...
+        rand_type.append(chr(random.randint(0,25) + 97))
+        ids.append(random.randint(1,5))
```
ackagel, line 20: typically, arrays/lists in python aren't initialized this way.  some one-liners for these could look like:
```
rand_type = random.choices(string.ascii_lowercase, k=1000)
ids = random.choices(range(1,6), k=1000)
```
(might have to `import string.ascii_lowercase` )

segmentation/utils/visualize.py
wang0324: We'll combine after we make the improvements to our plotting functions

segmentation/utils/visualize.py
wang0324: We originally didn't have line 39 but for some reason unique_ids would hold a bunch of NaNs, which is shown below.
image
To filter out the NaNs, we test if x==x which doesn't hold true because Nan != Nan

alex-l-kong: Ok, I see, however, the code may be somewhat confusing at first. Aside from putting a comment in to explain what this line does, there are two other possibilities I'd encourage you to look into.

The first is to filter out the NaNs in the if statement. NumPy has a function np.isnan(x) which allows you to check if x is nan or not.

The second is to call np.isnan(x) in your list comprehension.

ackagel: I second Alex's suggestions.  

But I'm curious as to where the NaN's originate.  I'm not sure if it's that meaningful, but if it's from another error earlier in the processing, then we could try to catch it now, and fix it in a later PR.

wang0324:
The reason for the build failure is in the spatial analysis, @alex-l-kong could you take a look?

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,109 @@
...
     sns.set(font_scale=.7)
     sns.clustermap(zplot, figsize=(8, 8), cmap="vlag")
+
+
+def visualize_patient_population_distribution(df, patient_col_name, population_col_name, color_map="jet"):
+    """Plots the distribution of the population given by total count, direct count, and proportion
+              Args:
+                  df: Pandas Dataframe containing columns with Patient ID and Cell Name
+                  id_col_name: Name of column containing categorical Patient data
+                  cell_col_name: Name of column in dataframe containing categorical Population data
```
alex-l-kong, line 26: Update the names in this comment to match the names of the parameters passed into the function.

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,109 @@
...
+            df_stacked = pd.concat([df_stacked, (pd.DataFrame([row_pop_types.values], columns=row_pop_types.index))],
+                                   axis=0, ignore_index=True)
+
+        # Gathering sums to sort dataframe by population count
+        sums = []
+        for index, row in df_stacked.iterrows():
+            sum = 0
+            for name in names:
+                val = row[name]
+                if (val == val):
```
alex-l-kong, line 61: If you're checking whether or not val exists, would recommend using clearer logic (e.g. `if val is None:`)

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,109 @@
...
+
+    def getSortedDf(isNormalized=False):
+        # Flipping the data so it's easily graphable
+        df_stacked = pd.DataFrame()
+        for id in unique_ids:
+            df_rows = df.loc[df[patient_col_name] == id]
+            row_pop_types = None
+
+            # Normalize it if asking for proportion, otherwise don't
+            if (isNormalized):
```
alex-l-kong, line 47: In Python, we generally don't put parentheses around the condition in the if statement.

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,109 @@
...
+
+        # Gathering sums to sort dataframe by population count
+        sums = []
+        for index, row in df_stacked.iterrows():
+            sum = 0
+            for name in names:
+                val = row[name]
+                if (val == val):
+                    sum += val
+            sums.append(sum)
```
alex-l-kong, line 63: It seems like what you're trying to do here is to sum up each row of your DataFrame. If that's the case, I'd recommend using the sum() function of Pandas. This will compartmentalize this logic into one line of code. To avoid the issue of summing up NaNs, you can pass a skipna parameter into the sum() function.

No need to reinvent the wheel.

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,109 @@
...
+        # Sort columns by total count
+        unsorted_pop_names = df_stacked.columns.tolist()
+        pop_sums = df_stacked.sum().sort_values(ascending=False)
+
+        def swap_columns(df, c1, c2):
+            df['temp'] = df[c1]
+            df[c1] = df[c2]
+            df[c2] = df['temp']
+            df.drop(columns=['temp'], inplace=True)
+            return df
```
alex-l-kong, line 79: This operation shouldn't need its own function.

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,109 @@
...
+
+    # Plot by count
+    getSortedDf().plot.bar(stacked=True, colormap=color_map)
+    plt.legend(loc='center left', bbox_to_anchor=(1.0, 0.5))
+    plt.xlabel("Patient ID")
+    plt.ylabel("Cell Count")
+    plt.title("Distribution of Population Count in Patients")
+
+    # Plot by Proportion
+    getSortedDf(isNormalized=True).plot.bar(stacked=True, legend=False, colormap=color_map)
```
alex-l-kong, line 105: See the comment above.

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,109 @@
...
+            df[c1] = df[c2]
+            df[c2] = df['temp']
+            df.drop(columns=['temp'], inplace=True)
+            return df
+
+        x = 0
+        for pop in pop_sums.index.tolist():
+            col_name = unsorted_pop_names[x]
+            if (col_name != pop):
+                df_stacked = swap_columns(df_stacked, col_name, pop)
```
alex-l-kong, line 85: I think you're trying to sort the columns to ensure the bar chart gets drawn correctly. If I'm not mistaken, I believe the bar plot function in Matplotlib has parameters that help you handle a lot of that already. If not Matplotlib, I'd recommend looking at Seaborn's barplot function. Could be wrong depending on what you're trying to accomplish, but I'd absolutely recommend looking into that.

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,109 @@
...
+
+        df_stacked = df_stacked[pop_sums.index.tolist()]
+        return df_stacked
```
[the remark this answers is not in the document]
ackagel: So I see what you're trying to do here but I think it can be accomplished in a different way.
```
df_stacked = pd.crosstab(df[patient_col_name], df[population_col_name])

id_order = df.groupby(patient_col_name).count().sort_values(by=population_col_name, ascending=False).index.values
type_order = df.groupby(population_col_name).count().sort_values(by=patient_col_name, ascending=False).index.values

df_stacked = df_stacked.reindex(id_order, axis='index')
df_stacked = df_stacked.reindex(type_order, axis='columns')
```
There might be a better way to get the sorting order, but I know that way should work.

For the normalized version, just add `normalized='index'` to the `crosstab` function.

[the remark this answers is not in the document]
wang0324: Thanks!

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,78 @@
...
+    df['temp'] = df[c1]
+    df[c1] = df[c2]
+    df[c2] = df['temp']
+    df.drop(columns=['temp'], inplace=True)
+    return df
+
+
+def getSortedDf(df, patient_col_name, population_col_name, isNormalized=False):
+    ids = df[patient_col_name]
+    names = df[population_col_name].value_counts().index.tolist()
```
alex-l-kong, line 31: It's more clear and easier do something like `names = df[population_col_name].unique().tolist()`.

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,78 @@
...
+    plt.ylabel("Cell Count")
+    plt.title("Distribution of Population Count in Patients")
+
+    # Plot by Proportion
+    getSortedDf(df, patient_col_name, population_col_name, isNormalized=True).plot.bar(stacked=True, legend=False,
+                                                                                       colormap=color_map)
+    plt.legend(loc='center left', bbox_to_anchor=(1.0, 0.5))
+    plt.xlabel("Patient ID")
+    plt.ylabel("Population Proportion")
+    plt.title("Distribution of Population Count Proportion in Patients")
```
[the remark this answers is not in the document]
ackagel: Yeah that'd be super.  Regardless, I think right now, each plot would overwrite the one prior to it, right?

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,78 @@
...
+
+def swap_columns(df, c1, c2):
+    df['temp'] = df[c1]
+    df[c1] = df[c2]
+    df[c2] = df['temp']
+    df.drop(columns=['temp'], inplace=True)
+    return df
+
+
+def getSortedDf(df, patient_col_name, population_col_name, isNormalized=False):
```
ngreenwald, line 29: Lets change this from CamelCase to snake_case naming convention, which we use in python

segmentation/utils/visualize.py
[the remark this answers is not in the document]
ngreenwald: Yes, there should be a keyword argument that controls which of these gets plotted. 
There should also be the option to save the plot, so the function should take in an optional save_dir argument that points to the correct folder.

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,78 @@
...
+    # Sorts by Kagel Method
+    id_order = df.groupby(patient_col_name).count().sort_values(by=population_col_name, ascending=False).index.values
+    pop_order = df.groupby(population_col_name).count().sort_values(by=patient_col_name, ascending=False).index.values
+    df_stacked = df_stacked.reindex(id_order, axis='index')
+    df_stacked = df_stacked.reindex(pop_order, axis='columns')
+
+    return df_stacked
+
+
+def visualize_patient_population_distribution(df, patient_col_name, population_col_name, color_map="jet"):
```
ackagel, line 49: `df` should probably be changed to something like `cell_data`.  I'm p open to other naming suggestions tho.  Just no more df pls.

segmentation/utils/visualize.py
ngreenwald: Also add a docstring

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,90 @@
...
+    id_order = cell_data.groupby(patient_col_name).count().sort_values(by=population_col_name, ascending=False).index.values
+    pop_order = cell_data.groupby(population_col_name).count().sort_values(by=patient_col_name, ascending=False).index.values
+    cell_data_stacked = cell_data_stacked.reindex(id_order, axis='index')
+    cell_data_stacked = cell_data_stacked.reindex(pop_order, axis='columns')
+
+    return cell_data_stacked
+
+
+def visualize_patient_population_distribution(cell_data, patient_col_name, population_col_name, color_map="jet",
+                                              show_total_count=True, show_distribution=True, show_proportion=True, save_dir=""):
```
ackagel, line 50: small thing.  `visualize_patient_population_distribution` will now save images to the current working directory by default.  To avoid this, set the default to `save_dir=None`, and, within each plot, wrap `savefig` as follows:
```
if save_dir is not None:
   plt.savefig(save_dir + "PlotName.png")
```

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,90 @@
...
+                  Args:
+                      cell_data: Pandas Dataframe containing columns with Patient ID and Cell Name
+                      patient_col_name: Name of column containing categorical Patient data
+                      population_col_name: Name of column in dataframe containing categorical Population data
+                      is_normalized: Boolean specifying whether to normalize cell counts or not, default is False
+                  Outputs:
+                      cell_data_stacked: DataFrame with rows and columns sorted by population"""
+    ids = cell_data[patient_col_name]
+    names = cell_data[population_col_name].unique().tolist()
+    unique_ids = cell_data[patient_col_name].unique()
```
ackagel, line 32: I don't think any of these variables are used in this function (double check me on this).  If that's the case, these lines can be removed.

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,90 @@
...
     # Replace the NA's and inf values with 0s
     z[np.isnan(z)] = 0
     z[np.isinf(z)] = 0
     # Assign numpy values respective phenotype labels
     zplot = pd.DataFrame(z, columns=pheno_titles, index=pheno_titles)
     sns.set(font_scale=.7)
     sns.clustermap(zplot, figsize=(8, 8), cmap="vlag")
+
+
+def getSortedData(cell_data, patient_col_name, population_col_name, is_normalized=False):
```
ngreenwald, line 21: We'll also want to make sure we test this function. A simple test function might create a pandas dataframe, then check that the values have appropriately sorted

segmentation/utils/visualize_test.py
```diff
@@ -10,3 +14,12 @@ def test_visualize_z_scores():
     # Assign random phenotype titles
     pheno_titles = [chr(i) for i in range(ord('a'), ord('z') + 1)]
     plot = visualize.visualize_z_scores(z, pheno_titles)
+
+
+def test_visualize_cells():
```
ngreenwald, line 19: We can add a couple assert statements to make sure the right plot is saved when the different options are provided by the user

segmentation/utils/visualize.py
```diff
@@ -1,20 +1,90 @@
...
+    """Gets the cell data and generates a new Sorted DataFrame with each row representing a patient and column representing Population categories
+                  Args:
+                      cell_data: Pandas Dataframe containing columns with Patient ID and Cell Name
+                      patient_col_name: Name of column containing categorical Patient data
+                      population_col_name: Name of column in dataframe containing categorical Population data
+                      is_normalized: Boolean specifying whether to normalize cell counts or not, default is False
+                  Outputs:
+                      cell_data_stacked: DataFrame with rows and columns sorted by population"""
```
alex-l-kong, line 29: In line with #141 , please add the types to each of these arguments in parentheses, for example:

```patient_col_name (str): ...```

File: segmentation/utils/visualize.py (hunk 25 above)
Comment (alex-l-kong) [inline segmentation/utils/visualize.py:21]: Please change function to snake case.

File: segmentation/utils/visualize_test.py [hunk 28]
```diff
@@ -10,3 +16,26 @@ def test_visualize_z_scores():
...
+
+def test_get_sort_data():
+    rand_type = random.choices(string.ascii_lowercase, k=100)
+    ids = random.choices(range(1, 10), k=100)
+    random_data = pd.DataFrame.from_dict({"PatientID": ids, "cell_type": rand_type})
+    print("Before sorting")
+    print(random_data)
+    print("----------------------------------")
+    print("After sorting")
+    print(visualize.get_sorted_data(random_data, "PatientID", "cell_type"))
```
alex-l-kong, line 29: Test functions shouldn't verify by using print statements. Instead, you should use assertions to make sure random_data is getting sorted correctly.

segmentation/utils/visualize_test.py
```diff
@@ -10,3 +16,26 @@ def test_visualize_z_scores():
...
+def test_visualize_cells():
+    rand_type = random.choices(string.ascii_lowercase, k=100)
+    ids = random.choices(range(1, 3), k=100)
+    df = pd.DataFrame.from_dict({"PatientID": ids, "cell_type": rand_type})
+    visualize.visualize_patient_population_distribution(df, "PatientID", "cell_type", save_dir="")
+
+    # Check if correct plots are saved
+    assert path.exists("PopulationDistribution.png"), "Should be showing distribution"
+    assert path.exists("TotalPopulationDistribution.png"), "Should be showing total count"
+    assert path.exists("PopulationProportion.png"), "Should be showing proportion"
```
alex-l-kong, line 41: Remove the side comments or else change to asterisk notation (ex. # should be showing distribution)

segmentation/utils/visualize_test.py
```diff
@@ -10,3 +16,25 @@ def test_visualize_z_scores():
...
+
+
+def test_get_sort_data():
+    rand_type = random.choices(string.ascii_lowercase, k=100)
+    ids = random.choices(range(1, 10), k=100)
+    random_data = pd.DataFrame.from_dict({"PatientID": ids, "cell_type": rand_type})
+    sorted_data = visualize.get_sorted_data(random_data, "PatientID", "cell_type")
+
+    row_sums = [row.sum() for index, row in sorted_data.iterrows()]
+    assert sorted(row_sums)
```
alex-l-kong, line 28: This assertion only tests for the existence of `sorted(row_sums)` aka that it's not `None`, so it will never fail. If you want to assert that `row_sums` is sorted properly, you should test for equality:

`row_sums == sorted(row_sums)`

By the way, I ran the function and row_sums actually looks like it's sorted in reverse order by default. In that case, you can do `assert list(reversed(row_sums)) == sorted(row_sums)`.

wang0324:
Are we good for merging @ngreenwald

Issue #120, opened 2020-07-07
Postprocessed CSV outputs normalized vs transformed


oliveria89:
Can these have more informative descriptions? -eg- normalized is cell-size normalized, transformed is arcsin-h transformed

ngreenwald:
We should add a description to the jupyter notebook cell that produces these for how they're computed.

PR #121, opened 2020-07-07
Subcellular segmentation update

ngreenwald:
This is a first update to the segmentation functions to allow extraction of subcellular segmentation. 

Previously, I had mocked up what I thought the inputs would look like. However, that version assumed we would get a probability mask representing nuclear signal which would be thresholded, and used to define individual compartments. However, we're instead going to have a distinct segmentation mask representing nuclei, so the inputs had to be changed. 

One thing in particular I'd like to get your feedback on is the best way to perform the indexing. 

Currently, we do the following:
```
for cell in cell_props:
        # get coords corresponding to current cell.
        # TODO: This is faster than mask-based indexing, but leads to confusing code. Is there
        # TODO a better way to use mask-based indexing to extract counts?
        cell_coords = cell.coords.T

        # calculate the total signal intensity within cell
        cell_counts = signal_extraction_utils.default_extraction(cell_coords, input_images)
```

We use regionprops to get an array of row, col coordinates representing the locations of pixels corresponding to cell i.
The more intuitive way to generate the mask would be 
`mask = segmentation_label == i`

However, using this mask to index into the image ends up being slow, since the mask is of size 1024x1024, whereas the list of coords is only 100-500 coordinates. 

The problem is that we then have a list of coords, which makes subsequent manipulations more awkward than if we just had a boolean mask we could manipulate. This is especially the case as we move towards subcellular segmentation. We might want to get only the inner ring of pixels, or the outermost ring, etc etc. This is easy with a binary mask, but not with a list of coords. 

I also added a number of TODOs, some of which I'll address in a separate PR, but didn't want to put too much in here.


segmentation/utils/segmentation_utils.py
```diff
@@ -297,103 +301,92 @@ def segment_images(input_images, segmentation_masks):
...
+                             dims=['compartments', 'cell_id', 'features'])
+
+    # get regionprops for each cell
+    cell_props = regionprops(segmentation_masks.loc[:, :, 'whole_cell'].values)
+
+    # loop through each cell in mask
+    for cell in cell_props:
+        # get coords corresponding to current cell.
+        # TODO: This is faster than mask-based indexing, but leads to confusing code. Is there
+        # TODO a better way to use mask-based indexing to extract counts?
```
ackagel, line 320: Not necessarily indexing, but if there's a good way of getting a mask as a `scipy.sparse.dok_matrix`, point-wise multiplying it by the image and calling sum on the resultant sparse matrix should be super fast and possibly more understandable at-a-glance.

scripts/Deepcell_Postprocessing.ipynb
```diff
@@ -261,7 +256,7 @@
    "source": [
     "# save output as CSV\n",
     "combined_normalized_data.to_csv(os.path.join(single_cell_dir, 'normalized_data.csv'), index=False)\n",
-    "combined_transformed_data.to_csv(os.path.join(single_cell_dir, 'transformed_data.csv'), index=False)\n"
+    "combined_transformed_data.to_csv(os.path.join(single_cell_dir, 'transformed_data.csv'), index=False)"
```
ackagel, line 259: Not particularly relevant for this review, should we make the output name configurable in an earlier cell (like where file paths are asked for)?  I ask because of issue #120.

ngreenwald:
Okay cool. I'm going to take another pass at this to add in some additional functionality, and then once it does everything want we can start optimizing.

PR #122, opened 2020-07-08
Synthetic spatial data generation: Part 2 (channel-based)

### LINKED ISSUE #124: Basic implementation of synthetic channel data
In order to start testing different approaches to compensate/correct segmentation, we need to have synthetic data of both a segmentation mask and corresponding channel data.

The most bare-bones implementation would accomplish the following:
- Generate a segmentation mask with a pair of cells that share a border. 
- Assign one cell in the pair to express marker A, and the other cell in the pair to express marker B
- For cell 1, determine whether the marker will be predominantly expressed in the nucleus, the membrane, or the entire cell
- For cell 2, determine whether the marker will be predominantly expressed in the nucleus, the membrane, or the entire cell
- Generate channel data for marker A, present in the appropriate location corresponding to the expression pattern of cell 1. 
- Generate channel data for marker B, present in the appropriate location corresponding to the expression pattern of cell 2.

alex-l-kong:
Addresses issue #124, a first pass into channel-based synthetic data generation.

With the distance-based portion of synthetic data generation out of the way, we now need to add another factor we didn't consider in the previous part for this PR.

Part 1 was completely insensitive to the various channels that appear in our data. Part 2 will attempt to resolve this by including more carefully crafted test data. This means that we can no longer settle for simply generating a random matrix or points that magically conform to our desired test conditions based on some hard-coded parameters. We now need to be concerned more with the actual shape of the cells we are generating. And this, of course, comes with more constraints.

While the shape is a concern, we don't have to try too hard to generate realistic-looking cells, we just need to make sure that we're not taking into account just cell centroids and that we factor in the boundaries of the cells in as well. And this leads to another problem; border correction.

Admittedly, I'm still in the learning stage about this portion but simply put, we don't have as much confidence toward the edges of a cell about certain properties. Techniques exist to correct this and that's another issue this branch wishes to resolve.

Closes #124.


segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +219,142 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+        center_value = prob_mask[cell_center_x, cell_center_y]
+        prob_mask = norm.pdf(prob_mask, center_value)
+    else:
+        prob_mask = np.maximum(np.abs(prob_mask_rows - cell_center_x),
+                               np.abs(prob_mask_cols - cell_center_y)) - (size_img[0] + size_img[1]) * -random()
+        edge_value = prob_mask[cell_center_x, cell_center_y - radius]
+        prob_mask = norm.pdf(prob_mask, edge_value)
+
+    # extract only the coordinates defined by the cell region and set the prob mask accordingly
+    signal_data[cell_region[0], cell_region[1]] = prob_mask[cell_region[0], cell_region[1]]
```
ngreenwald, line 281: I don't think we want to limit the signal to only occur within the boundaries of the cell. One of the reasons we're making this synthetic data is because we know that there can be signal outside of the segmentation boundaries we've drawn.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +219,142 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+    center_1 = (radius, radius)
+    center_2 = (radius, radius * 3 + 1)
+
+    # draw the coordnates covered for the two cell
+    x_coords_cell_1, y_coords_cell_1 = circle(center_1[0], center_1[1], radius + 1)
+    x_coords_cell_2, y_coords_cell_2 = circle(center_2[0], center_2[1], radius + 1)
+
+    # now store the regions generated by skimage.draw.circle into a dict with the key
+    # identifying the marker
+    cell_regions = {1: (x_coords_cell_1, y_coords_cell_1),
```
ngreenwald, line 320: I think instead of storing the actual coordinates of the cells, we should just store the value we assigned them in a segmentation label mask, and then pass the mask.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +219,142 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+    # but around an edge value for membrane
+    if expression == "nuclear":
+        prob_mask = size_img[0] + size_img[1] - np.maximum(np.abs(prob_mask_rows - cell_center_x),
+                                                           np.abs(prob_mask_cols - cell_center_y)) * random()
+        center_value = prob_mask[cell_center_x, cell_center_y]
+        prob_mask = norm.pdf(prob_mask, center_value)
+    else:
+        prob_mask = np.maximum(np.abs(prob_mask_rows - cell_center_x),
+                               np.abs(prob_mask_cols - cell_center_y)) - (size_img[0] + size_img[1]) * -random()
+        edge_value = prob_mask[cell_center_x, cell_center_y - radius]
```
ngreenwald, line 277: Why does membrane  signal involve an offset in y but not in x?

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +219,142 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+                                       pattern, cell_region, cell_center):
+    """
+    This function generates the signal-level test data for a specified channel (nuclear and membrane)
+    based on a given pattern and the marker num we wish to generate data for.
+
+    Args:
+        size_img: the dimensions of the image
+        radius: the radius of the cell
+        marker_num: the marker num of the cell we're looking at
+        expression: whether the channel we're looking at is nuclear or membrane
```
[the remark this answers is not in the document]
alex-l-kong: Thanks for clarifying. I was overthinking a bit about other potential ways a pattern could be represented (ex. "strong nuclear" vs "weak nuclear"). But yeah, they should be the same.

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: Right, that slipped my mind. That'll be implemented in the next version to review.

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: I'm trying to center the distribution around a value extracted from the edge of the cell, so I pick one of the many possible edge values of the cell (I could have also done [cell_center_x, cell_center_y + radius] or [cell_center_x - radius, cell_center_y]. I can definitely comment on that since it seems like that's not clear enough as it stands.

Although now that I think about it, it may not be the best way to ensure a membrane-based pattern. Unlike nuclear representation, we may not want to make the assumption that we're most confident about a membrane pattern at the very edge of the cell. For that matter, it may also be unwise to make the assumption that nuclear representation is strongest at the very center. What do you think @ngreenwald ?

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +219,142 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+    prob_mask_rows, prob_mask_cols = np.ogrid[:size_img[0], :size_img[1]]
+
+    # generate an array of numbers valued depending on how close or far away the coordinate is
+    # from the center or the membrane
+    # note that we multiply by -random() for the membrane case because the values
+    # will initially be negative and it will help to make the values consistent
+    # also, we need to center the distribution around the center for nuclear
+    # but around an edge value for membrane
+    if expression == "nuclear":
+        prob_mask = size_img[0] + size_img[1] - np.maximum(np.abs(prob_mask_rows - cell_center_x),
```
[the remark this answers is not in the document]
alex-l-kong: Ah OK. I should include a comment about what is going on here for sure.

Yes, prob_mask is a placeholder for the intermediate calculations. This is because the current way I generate probability values covers the entire 1024x1024 of the image, but I'll only need to extract a certain non-square subset. That being said, I'm going to work on trying not to use a intermediate variable next version.

For nuclear expression, the best way to describe what prob_mask becomes initially is a topographical map of a mountain, with higher values indicating higher elevation. I want the highest value to be at the center and for the values to decrease the further away they get from the center. This is so when I calculate the probability of representation, it returns higher values the closer they are to the center (which I center the distribution around).

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +219,142 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+    else:
+        prob_mask = np.maximum(np.abs(prob_mask_rows - cell_center_x),
+                               np.abs(prob_mask_cols - cell_center_y)) - (size_img[0] + size_img[1]) * -random()
+        edge_value = prob_mask[cell_center_x, cell_center_y - radius]
+        prob_mask = norm.pdf(prob_mask, edge_value)
+
+    # extract only the coordinates defined by the cell region and set the prob mask accordingly
+    signal_data[cell_region[0], cell_region[1]] = prob_mask[cell_region[0], cell_region[1]]
+
+    return signal_data
```
ackagel, line 283: I think the easiest, most consistent way to do this, is to generate an ellipse w/ random values and corrupt it a little bit (something like randomly shifting every pixel value up/down/left/right).
```
# generate 1px thick ellipse coords (_perimeter for membrane, just ellipse for nuclear)
xx, yy = skimage.draw.ellipse_perimeter(cx, cy, rx, ry, rotation=angle)

# generate corruption of coordinates
xx_cor = xx + np.around(4*(np.random.rand(*xx.shape)-0.5)
yy_cor = yy + np.around(4*(np.random.rand(*yy.shape)-0.5)

# set signal at ellipse+corruption values (not necessarily 1)
signal_data = np.zeros(size_img)
signal_data[xx, yy] = 1
signal_data[xx_cor, yy_cor] = 1

return signal_data
```
Here's an example of the output of something like this:
example

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +219,142 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+def generate_two_cell_test_signal_data(size_img, radius, expression,
+                                       pattern, cell_region, cell_center):
+    """
+    This function generates the signal-level test data for a specified channel (nuclear and membrane)
+    based on a given pattern and the marker num we wish to generate data for.
+
+    Args:
+        size_img: the dimensions of the image
+        radius: the radius of the cell
+        marker_num: the marker num of the cell we're looking at
```
ackagel, line 240: doesn't look like `marker_num` is a part of this function anymore

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +219,142 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+
+    # generate the nuclear- and membrane-level channel signal for each cell
+    for i in range(2):
+        sample_channel_data[0, :, :] += generate_two_cell_test_signal_data(size_img=size_img, radius=radius,
+                                                                           expression='nuclear', pattern='nuclear',
+                                                                           cell_region=cell_regions[i + 1], cell_center=cell_centers[i + 1])
+
+        sample_channel_data[1, :, :] += generate_two_cell_test_signal_data(size_img=size_img, radius=radius,
+                                                                           expression='membrane', pattern='membrane',
+                                                                           cell_region=cell_regions[i + 1], cell_center=cell_centers[i + 1])
```
ackagel, line 357: I think you can do something like
```
for cr, cc in zip(cell_regions, cell_centers):
```
for a teenytiny speed up

segmentation/utils/synthetic_spatial_datagen.py
ngreenwald: I agree, I think this will be simpler. It also addresses the issue you brought up above @alex-l-kong, which is that we aren't necessarily more confident about values in the very center of the nucleus than at the edge. 

Here's how this might look. Generate a ellipse of radius r. This is the nucleus. Assign all pixels within the nucelus to a constant value. Then, draw three perimeters of expanding radii, each of which is subsampled or jittered slightly more than the last. 

We have a lot of information that we can leverage in terms of what the signal looks like, whereas if you construct your distribution from scratch with a PDF you lose that.

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: Yeah, I'll remove that.

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: Thanks, guys for the suggestions! I'll see to implementing something similar (and simpler!).

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: Ah yes, thanks for that suggestion.

alex-l-kong:
I still need to work on making the jitter somewhat more random the further away we get from the nucleus/membrane.

alex-l-kong:
I have not written the testing functions yet, mostly because I'll be doing further restructuring of the code. If you want to see the output, I've passed a plot parameter so matplotlib displays the distribution of signal.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,302 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+        ring_region_1_x = np.concatenate((ring_region_1_x, region_1_perim_x))
+        ring_region_1_y = np.concatenate((ring_region_1_y, region_1_perim_y))
+        ring_region_2_x = np.concatenate((ring_region_2_x, region_2_perim_x))
+        ring_region_2_y = np.concatenate((ring_region_2_y, region_2_perim_y))
+
+    # create a tuple for the x and y coordinates for each cell ring
+    ring_region_1 = (ring_region_1_x, ring_region_1_y)
+    ring_region_2 = (ring_region_2_x, ring_region_2_y)
+
+    return ring_region_1, ring_region_2
```
ackagel, line 271: I think it is/was actually better to subtract an inner radius circle from an outer radius:
```
r1x, r1y = circle(center_1[0], center_1[1], outer_rad, shape=size_img)
condition = ((r1x-center_1[0])**2 + (r1y-center_1[1])**2 > inner_rad**2)
r1x = r1x[condition]
r1y = r1y[condition]

img = np.zeros(img_size)
img[r1x, r1y] = 1
```
Idk if the condition statement automagically uses broadcasting or not, but it still should be pretty fast.

ackagel, line 271: I feel like we can just return the image; I don't see a real need to return tuples of coordinates

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,302 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+    # over trying to fine tune jitter factor and what not
+    # number of points to keep is inversely proportional to
+    # how far away the ring is from the nucleus or membrane respectively
+    ring_region_1_selection = np.random.choice(len(ring_region_1_x_jitter_coords), int(len(ring_region_1_x_jitter_coords) / jitter_multiplier))
+    ring_region_2_selection = np.random.choice(len(ring_region_2_x_jitter_coords), int(len(ring_region_2_x_jitter_coords) / jitter_multiplier))
+
+    ring_region_1_x_jitter_coords = ring_region_1_x_jitter_coords[ring_region_1_selection]
+    ring_region_1_y_jitter_coords = ring_region_1_y_jitter_coords[ring_region_1_selection]
+    ring_region_2_x_jitter_coords = ring_region_2_x_jitter_coords[ring_region_2_selection]
+    ring_region_2_y_jitter_coords = ring_region_2_y_jitter_coords[ring_region_2_selection]
```
ackagel, line 329: since `jitter_factor` wasn't included in this logic, it made `jitter_factor=0` in higher level functions not work properly.  If we remove partitioning, this can be removed as well.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,302 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+    if seed:
+        np.random.seed(seed)
+
+    # generate some jitter for each x and y coordinate
+    # the way we do this is to generate random offset based on the jitter_factor multiplier
+    # and sampling from a uniform distribution with an additional jitter_factor and jitter_multiplier multiplier
+    ring_region_1_x_jitter = (ring_region_1[0] + np.around(np.random.uniform(-jitter_factor * jitter_multiplier, jitter_factor * jitter_multiplier, len(ring_region_1[0])))).astype(np.int16)
+    ring_region_1_y_jitter = (ring_region_1[1] + np.around(np.random.uniform(-jitter_factor * jitter_multiplier, jitter_factor * jitter_multiplier, len(ring_region_1[1])))).astype(np.int16)
+    ring_region_2_x_jitter = (ring_region_2[0] + np.around(np.random.uniform(-jitter_factor * jitter_multiplier, jitter_factor * jitter_multiplier, len(ring_region_2[0])))).astype(np.int16)
+    ring_region_2_y_jitter = (ring_region_2[1] + np.around(np.random.uniform(-jitter_factor * jitter_multiplier, jitter_factor * jitter_multiplier, len(ring_region_2[1])))).astype(np.int16)
```
ackagel, line 303: You can just call `np.random.uniform` just once with size/shape set to `ring_region_1.size + ring_region_2.size` and do:
```
num_coords1 = ring_region_1.size/2
ring_region_1_x_jitter = ring_region_1[0] + jitter[:num_coords1]
ring_region_1_y_jitter = ring_region_1[1] + jitter[num_coords1:2*num_coords1]
...
```
that should be faster and resolve pep8 inspired formatting headaches

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,302 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+
+    # now ensure the jitter coordinates don't fall out of bounds
+    # this we need to do because the offsets could fall out of bounds
+    ring_region_1_jitter_indices = (ring_region_1_x_jitter >= 0) & (ring_region_1_x_jitter < size_img[0]) & \
+                                   (ring_region_1_y_jitter >= 0) & (ring_region_1_y_jitter < size_img[1])
+
+    ring_region_2_jitter_indices = (ring_region_2_x_jitter >= 0) & (ring_region_2_x_jitter < size_img[0]) & \
+                                   (ring_region_2_y_jitter >= 0) & (ring_region_2_y_jitter < size_img[1])
```
ackagel, line 311: good pre-emptive catch

alex-l-kong:
OK, the process has been simplified, like, super-simplified now. Maybe a bit too simplified, but definitely much, much easier to follow and much shorter as well. I've held off on writing the test functions for now because I may need to further restructure.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,163 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+        memb_cell_ids (list): a list of which cells we wish to generate membrane signal for, if None assume just cell 2
+
+    Returns:
+        sample_segmentation_mask (numpy): an array which contains the labeled cell regions
+        sample_nuclear_signal (TIF): defines the nuclear signal for the desired cells
+        sample_membrane_signal (TIF): defines the membrane signal for the desired cells
+    """
+
+    # default nuclear signal cell to 1 and membrane signal cell to 2 if not specified
+    if not nuc_cell_ids:
```
ngreenwald, line 370: I think this just adds clutter to the code, why not set nuc_cell_ids to 1 in the function definition as the default instead?

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,163 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+        memb_cell_ids = [2]
+
+    # generate the segmentation mask
+    sample_segmentation_mask = generate_two_cell_test_segmentation_mask(size_img, cell_radius)
+
+    # generate the nuclear and membrane-level signal
+    sample_nuclear_signal = generate_two_cell_test_nuclear_signal(sample_segmentation_mask, nuc_radius, nuc_cell_ids)
+    sample_membrane_signal = generate_two_cell_test_membrane_signal(sample_segmentation_mask, memb_diameter, memb_cell_ids)
+
+    return sample_segmentation_mask, sample_nuclear_signal, sample_membrane_signal
```
ngreenwald, line 383: Lets reformat the output to match this function, which is what it will eventually replace: https://github.com/angelolab/segmentation/blob/5081bf50fd6b9e203426cfeec2c50f207a3914d3/segmentation/utils/segmentation_utils_test.py#L42

alex-l-kong: 👍

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: Ok, will fix that.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,163 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+    Returns:
+        sample_segmentation_mask (numpy): an array of dimensions size_img with two separate labeled
+            cells that border each other
+    """
+
+    # define the segmentation mask
+    sample_segmentation_mask = np.zeros(size_img)
+
+    center_1 = (int(size_img[0] / 2), int(size_img[0] / 2))
+    center_2 = (int(size_img[0]) / 2, int(size_img[0] / 2 + cell_radius * 2))
```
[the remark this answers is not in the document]
ackagel: also, should the cell centers be parametrizable at all?

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,163 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+        cell_radius (int): the radius of each cell
+        nuc_radius (int): the radius of each nucleus
+        memb_diameter (int): the diameter of each membrane
+        nuc_cell_ids (list): a list of which cells we wish to generate nuclear signal for, if None assume just cell 1
+        memb_cell_ids (list): a list of which cells we wish to generate membrane signal for, if None assume just cell 2
+
+    Returns:
+        sample_segmentation_mask (numpy): an array which contains the labeled cell regions
+        sample_nuclear_signal (TIF): defines the nuclear signal for the desired cells
+        sample_membrane_signal (TIF): defines the membrane signal for the desired cells
```
ackagel, line 358: (TIF) is a little vague since it could be anything from a string, file pointer, numpy ndarray, mibitiff object, etc.  I'd change it to (numpy).

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,163 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+        center = cell_centers[cell]
+
+        # generate both the coordinates of the cell region and non-membrane region
+        # for proper circle subtraction to generate membrane
+        cell_region_x, cell_region_y = circle(center[0], center[1], cell_radius, shape=size_img)
+        non_memb_region_x, non_memb_region_y = circle(center[0], center[1], cell_radius, shape=size_img)
+
+        # perform circle subtraction
+        sample_membrane_signal[cell_region_x, cell_region_y] = 1
+        sample_membrane_signal[non_memb_region_x, non_memb_region_y] = 1
```
ackagel, line 334: shouldn't the id numbers (1 and 1) be different here (1 and 0(2?)) or am I missing something?

segmentation/utils/synthetic_spatial_datagen.py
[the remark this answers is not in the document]
alex-l-kong: Can definitely work on changing the data types. For now, I think it's fine to have fixed centers, as we move forward we can definitely work on parameterizing the centers so they appear in more random locations.

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: Yeah, meant to make that numpy. No idea why I put TIF there.

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: Crap, yes it should be 1 and 0.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,166 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+    # define the nuclear signal array
+    sample_nuclear_signal = np.zeros(segmentation_mask.shape)
+
+    for cell in nuc_cell_ids:
+        center = cell_centers[cell]
+
+        # generate the nuclear region in the middle of the cell with the same cell center
+        # and set signal to a uniform value
+        nuc_region_x, nuc_region_y = circle(center[0], center[1], nuc_radius, shape=size_img)
+        sample_nuclear_signal[nuc_region_x, nuc_region_y] = 1
```
ngreenwald, line 297: I find it's useful for debugging purposes if the signal isn't exactly equal to the area of the cell. Can change to 2 or 10 or whatever you want, just not 1.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,166 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+    This function generates membrane signal for the provided cells
+
+    Args:
+        segmentation_mask (numpy): an array which contains the labeled cell regions
+        cell_centers (dict): a dictionary which contains the centers associated with each cell region
+        size_img (tuple): the dimensions of the image we wish to generate
+        cell_radius (int): the radius of the entire cell, needed to do proper circle subtraction
+            for a ring-shaped membrane
+        memb_cell_ids (list): a list of cells we wish to generate nuclear signal for, if None assume just cell 2
+        memb_diameter (int): the diameter of the membrane ring of each cell
```
ngreenwald, line 318: I think this is more like mem_thickness than diameter

segmentation/utils/synthetic_spatial_datagen_test.py
```diff
@@ -102,3 +106,70 @@ def test_generate_test_label_map():
...
     assert (np.sort(label_map_flat) == np.sort(centroid_indices) + 1).all()
+
+
+def test_generate_two_cell_test_segmentation_mask():
+    # this function tests the functionality of generating the segmentation mask
+
+    sample_segmentation_mask, sample_cell_centers = synthetic_spatial_datagen.generate_two_cell_test_segmentation_mask()
+
+    # assert that our labels are just blank, 1, and 2
+    assert set(np.unique(sample_segmentation_mask)) == set([0, 1, 2])
```
ackagel, line 117: I don't think you need the np unique here

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -222,10 +221,166 @@ def generate_test_label_map(size_img=(1024, 1024), num_A=100, num_B=100, num_C=1
...
+    for cell in memb_cell_ids:
+        center = cell_centers[cell]
+
+        # generate both the coordinates of the cell region and non-membrane region
+        # for proper circle subtraction to generate membrane
+        cell_region_x, cell_region_y = circle(center[0], center[1], cell_radius, shape=size_img)
+        non_memb_region_x, non_memb_region_y = circle(center[0], center[1], cell_radius - memb_diameter, shape=size_img)
+
+        # perform circle subtraction
+        sample_membrane_signal[cell_region_x, cell_region_y] = 1
```
ngreenwald, line 337: Same thing, pick a different multiple than nuclear, but not 1

segmentation/utils/synthetic_spatial_datagen_test.py
```diff
@@ -102,3 +106,70 @@ def test_generate_test_label_map():
...
+    # we'll limit ourselves to testing the final channel data creation here
+
+    sample_segmentation_mask, sample_channel_data = synthetic_spatial_datagen.generate_two_cell_test_channel_synthetic_data()
+
+    # assert that we've created both a nuclear and membrane channels
+    assert sample_channel_data.shape[2] == 2
+
+    # assert that we've only labeled nuclear and membrane signal with 0 or 1
+    assert set(np.unique(sample_channel_data[:, :, 0])) == set([0, 1])
+    assert set(np.unique(sample_channel_data[:, :, 1])) == set([0, 1])
```
ackagel, line 175: see earlier np unique comment

segmentation/utils/synthetic_spatial_datagen_test.py
```diff
@@ -102,3 +106,70 @@ def test_generate_test_label_map():
...
+    assert set(np.unique(sample_segmentation_mask)) == set([0, 1, 2])
+
+    # assert that our centers are being labeled correctly
+    assert sample_segmentation_mask[sample_cell_centers[1][0], sample_cell_centers[1][0]] == 1
+    assert sample_segmentation_mask[sample_cell_centers[2][0], sample_cell_centers[2][1]] == 2
+
+    # using the default cell radius of 10, assert that the cells are next to each other
+    # we only include the offset columnwise because that's how the cells are generated next to each other
+    assert sample_segmentation_mask[sample_cell_centers[1][0], sample_cell_centers[1][1] + 10] == 2
+    assert sample_segmentation_mask[sample_cell_centers[2][0], sample_cell_centers[2][1] - 10] == 1
```
ackagel, line 126: just setting the cell radius to 10 in the function call (line 114) will guarantee that this test function doesn't break here (in case someone decides to change the default cell radius later).

segmentation/utils/synthetic_spatial_datagen_test.py
```diff
@@ -102,3 +106,70 @@ def test_generate_test_label_map():
...
+
+    sample_segmentation_mask, sample_cell_centers = synthetic_spatial_datagen.generate_two_cell_test_segmentation_mask()
+
+    sample_membrane_signal = synthetic_spatial_datagen.generate_two_cell_test_membrane_signal(sample_segmentation_mask,
+                                                                                              sample_cell_centers)
+
+    # assuming the default membrane diameter of 5, assert that our membrane inner edge is being labeled correctly
+    # we only include the offset columnwise by choice: it could be done rowwise as well
+    # we only care about cell 2 because that is the only membrane-level expression cell by default
+    assert sample_membrane_signal[sample_cell_centers[2][0], sample_cell_centers[2][1] - 5] == 1
```
ackagel, line 157: see earlier comment on function default dependence (honestly, I think I might have to go back into some of my data_utils test functions to fix that kinda thing...)

alex-l-kong:
> Looks good to me. Just edit the description to close #124.

Done.

ngreenwald:
Github gets confused if you have any text in between 'closes' and the issue number. Once it's linked successfully, you'll see the text get highlighted and there will be a message in this thread saying 'Alex linked an issue that may be closed...'

alex-l-kong:
> Github gets confused if you have any text in between 'closes' and the issue number. Once it's linked successfully, you'll see the text get highlighted and there will be a message in this thread saying 'Alex linked an issue that may be closed...'

OK, looks like it's fixed now.

PR #123, opened 2020-07-08
Compartment extraction refactor

ngreenwald:
Further feature additions and some refactoring for extracting imaging signal from subcellular compartments.

- Add helper function (find_nuclear_mask_id) to get the corresponding nuclear mask for the specified cell. This allows us to get the regionprops information on morphology of that nucleus and link it to the cell currently being analyzed

- Add helper function (transform_expression_matrix) to perform size normalization and arcsinh transformation. Also expand functionality to work for nuclear information as well as whole_cell

- Reshuffle logic for computing regionprops information to be inside of the helper function, rather than in main function, to allow matching with corresponding nuclear object. 

- Code cleanup

- Increase testing coverage

Any feedback you guys have on the changes above would be great. 

The main goal for this is to make the logic easier to follow, more modular, and potentially extensible to additional compartments in the future, so definitely raise any suggestions/concerns you might have in that vein.

I also added in TODOs for the testing function for where stuff can likely be consolidated and improved in the future, but I think I'll leave that to a future PR from one of you.


segmentation/utils/segmentation_utils.py
```diff
@@ -278,6 +278,86 @@ def combine_segmentation_masks(big_mask_xr, small_mask_xr, size_threshold,
...
+        cell_coords: list of coords specifying pixels that belong to a cell
+
+    Returns:
+        nuclear_mask_id: ID of the nuclear mask that overlaps most with cell. If not matches found,
+            returns None.
+    """
+
+    ids, counts = np.unique(nuc_segmentation_mask[tuple(cell_coords.T)], return_counts=True)
+
+    max_id = ids[np.argmax(counts)]
```
ackagel, line 295: Not sure if it matters, but if there are equal counts of many different id's, `np.argmax(counts)` will only return first-in-order id.

segmentation/utils/segmentation_utils.py
```diff
@@ -278,6 +278,86 @@ def combine_segmentation_masks(big_mask_xr, small_mask_xr, size_threshold,
...
+    if max_id == 0:
+        if len(ids) == 1:
+            nuclear_mask_id = None
+        else:
+            non_zero_idx = ids > 0
+            ids, counts = ids[non_zero_idx], counts[non_zero_idx]
+
+            nuclear_mask_id = ids[np.argmax(counts)]
+    else:
+        nuclear_mask_id = max_id
```
ackagel, line 306: I think this logic can be cleaned up with something like:
```
# np.unique returns in sorted order, so this works assuming all positive ids
if ids[1:]:
  nuclear_mask_id = ids[np.argmax(counts[1:]) + 1]
else:
  nuclear_mask_id = None
```
I think this could even be a one-line ternary statement, but that might look a little messy

segmentation/utils/segmentation_utils.py
```diff
@@ -278,6 +278,86 @@ def combine_segmentation_masks(big_mask_xr, small_mask_xr, size_threshold,
...
+
+    return nuclear_mask_id
+
+
+def transform_expression_matrix(cell_data, transform, transform_kwargs={}):
+    """Transform an xarray of marker counts with supplied transformation
+
+    Args:
+        cell_data: xarray containing marker expression values
+        transform: the type of transform to apply. Must be one of ['size_norm', 'arcsinh']
```
ackagel, line 316: I think a short list of acceptable `transform_kwargs` inputs would be good

segmentation/utils/segmentation_utils.py
```diff
@@ -278,6 +278,86 @@ def combine_segmentation_masks(big_mask_xr, small_mask_xr, size_threshold,
...
+
+        # first linearly scale the data
+        cell_data_transformed.values[:, :, channel_start:channel_end] *= linear_factor
+
+        # arcsinh transformation
+        cell_data_transformed.values[:, :, channel_start:channel_end] = \
+            np.arcsinh(cell_data_transformed[:, :, channel_start:channel_end].values)
+
+    return cell_data_transformed
+
```
ackagel, line 359: Should this be split into two different functions?  I know the preprocessing for both functionalities is the same, but that could be put in an internal utility function.  That way, if more types of transformations are wanted later, we don't have to keep adding `transform` cases into this function

Another option in this line of thinking is having the `transform` argument be a function instead of a string, but that would probably be a little confusing to lab-users

segmentation/utils/segmentation_utils.py
```diff
@@ -287,59 +367,87 @@ def compute_marker_counts(input_images, segmentation_masks, nuclear_counts=False
...
+                nuc_features = np.concatenate((nuc_counts, current_nuc_props), axis=None)
 
-    return xr_counts
+                # add counts of each marker to appropriate column
+                marker_counts.loc['nuclear', nuc_id, marker_counts.features[1]:] = nuc_features
+
+                # add cell size to first column
+                marker_counts.loc['nuclear', nuc_id, marker_counts.features[0]] = \
+                    nuc_coords.shape[0]
+
```
ackagel, line 449: A lot of the nuclear counting logic is the shared with the whole cell.  I think a lot of this could be moved into a loop over the `segmentation_masks.compartments` (currently from 'whole_cell' to 'nuclear'), which opens the door to easily adding more types of mask counting.  

The only hiccup is associating each type of label mask (nuclear, etc) to each cell, but that functionality is already covered by `find_nuclear_mask_id`.  So, this could be a loop so that for each `c in compartments` apply `find_nuclear_mask_id` if `c is not 'whole_cell'` and then perform the rest of the shared logic, storing the result in `marker_counts.loc[c, nuc_id, markercounts.features[1]:]`

ackagel: also, sorry i only just now remembered to use the 'start a review button' and not single comment.

segmentation/utils/segmentation_utils.py
```diff
@@ -287,59 +367,87 @@ def compute_marker_counts(input_images, segmentation_masks, nuclear_counts=False
             nuclear_counts: boolean flag to determine whether nuclear counts are returned
 
         Returns:
-            xr_counts: xarray containing segmented data of cells x markers"""
-
-    if type(input_images) is not xr.DataArray:
-        raise ValueError("Incorrect data type for ground_truth, expecting xarray")
+            xr_counts: xarray containing segmented data of cells x markers
```
ackagel, line 370: now this is `marker_counts`

segmentation/utils/segmentation_utils_test.py
```diff
@@ -182,6 +182,136 @@ def test_watershed_transform():
...
+    assert np.array_equal(arcsinh_data.loc[:, :, unchanged_cols].values,
+                          cell_data.loc[:, :, unchanged_cols].values)
+
+    # TODO: In general it's bad practice for tests to call the same function as code under test
+    for cell in cell_data.cell_id:
+        arcsinh_vals = np.arcsinh(cell_data.loc[:, cell, modified_cols].values)
+        assert np.array_equal(arcsinh_data.loc[:, cell, modified_cols].values, arcsinh_vals)
+
+
+# TODO: The testing for this function can be improved. Repeated code, and lots of manual checks
```
ackagel, line 314: to avoid repeated code here, the setup/initialization could be put into an internal helper function, with a flag for how many compartments to add.  

As for the 'same function as code' bad practice, I don't see a reasonable way around it here.  Only thing that comes to mind is that if a set seed is used for the randomized input, we can pre-compute and store the desired output in a file to load during testing, but this seems like overkill for making sure that `np.arcsinh` still works as expected.

segmentation/utils/segmentation_utils.py
```diff
@@ -278,6 +278,86 @@ def combine_segmentation_masks(big_mask_xr, small_mask_xr, size_threshold,
...
+            ids, counts = ids[non_zero_idx], counts[non_zero_idx]
+
+            nuclear_mask_id = ids[np.argmax(counts)]
+    else:
+        nuclear_mask_id = max_id
+
+    return nuclear_mask_id
+
+
+def transform_expression_matrix(cell_data, transform, transform_kwargs={}):
```
ackagel, line 311: `transform_kwargs = None` is safer/better practices

segmentation/utils/segmentation_utils.py
ngreenwald: For now we're not going to deal with cells that have multiple nuclei, or are poorly segmented and could potentially have multiple. But something to think about for the future.

segmentation/utils/segmentation_utils.py
ngreenwald: This doesn't work for me: 
image

segmentation/utils/segmentation_utils.py
ngreenwald: Or maybe we could pull out these transforms into their own function, and then just call them here, rather than defining the code here? Or do you think it's better to not have a high-level function at all, and instead pull our the redundant code into a low-level helper function that gets called at the beginning of each

segmentation/utils/segmentation_utils.py
ngreenwald: Right, the issue is that find_nuclear_mask_id requires the cell coords for that cell to calculate the right nuclear ID. 

You're totally right that we'll be incorporating additional compartments. Gonna punt on this for now, whichever one of you gets to this first can take a stab at it.

segmentation/utils/segmentation_utils_test.py
ngreenwald: Agreed. I think maybe switching to doing it line-by-line rather than the whole thing at once is enough of a difference? Just want to catch any dumb mistakes

segmentation/utils/segmentation_utils.py
ackagel: oh that should be something like `if x[1:] is not None`, sorry; all it means to do is check to see if the non-zero part of the array is zero

segmentation/utils/segmentation_utils.py
ackagel: I like the idea of the high-level transform function, but if there are more transforms, we basically end up with a switch-case over what transform the user is asking for.  If we plan on having only these two transforms, then the first option you mentioned is probably best.  If we plan on having significantly more, then something along the lines of the second option.

ngreenwald: Got it. I'm not sure TBH, I think we'll have to see. In that case I'll leave it as is and we can decide which makes more sense as we continue to build it out

segmentation/utils/segmentation_utils.py
ngreenwald: Oooh, got it. That's much neater. However, I don't think it works in the case that there's only a single ID and it's not zero, right?

ackagel: possibly not; idk if it's bad practice, but you could maybe wrap it in some sort of try/catch instead?  Some thing like:
```
try:
  nuclear_mask_id = ids[np.argmax(counts[1:])+1]
except:
  nuclear_mask_id = None
```

ackagel: practice-wise, it looks like this is fine as long as except isn't blanket-ed.  So using `except ValueError as e:` is best.  Pretty sure it'll always be a `ValueError` from calling `np.argmax` on an empty list; there's a possibility it'll be an `IndexError` if the size of `counts` is somehow not the same as `ids`, but I don't think we want that behavior anyways

ngreenwald: Okay, in that case I'm going to leave it as is. We want to keep the functionality of a single non-zero value being accurately reported as the nuclear_id, and not raise an exception.

ngreenwald:
@alex-l-kong any comments?

segmentation/utils/segmentation_utils.py
ackagel: oooh, nvm.  I misunderstood what you were saying, sorry.  I think the fix is:
```
if ids[ids != 0] is not None:
  nuclear_mask_id = ids[np.argmax(counts[ids != 0])]
else:
  nuclear_mask_id = None
```
this could also be replaced with the try/except structure as well.  But this is probably clearer syntax.

segmentation/utils/segmentation_utils.py
```diff
@@ -278,6 +278,86 @@ def combine_segmentation_masks(big_mask_xr, small_mask_xr, size_threshold,
...
+        # Only calculate where cell_size > 0
+        cell_data_transformed.values[:, :, channel_start:channel_end] = \
+            np.divide(cell_data_transformed.values[:, :, channel_start:channel_end],
+                      cell_size_large, where=cell_size_large > 0)
+
+    elif transform == 'arcsinh':
+        linear_factor = transform_kwargs.get('linear_factor', 100)
+
+        # first linearly scale the data
+        cell_data_transformed.values[:, :, channel_start:channel_end] *= linear_factor
```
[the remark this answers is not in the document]
ngreenwald: Totally. I think this got changed while your review was pending, it now defaults to an empty dict if None is specified so .get will return 100 in that case.

segmentation/utils/segmentation_utils.py
```diff
@@ -287,59 +367,87 @@ def compute_marker_counts(input_images, segmentation_masks, nuclear_counts=False
             nuclear_counts: boolean flag to determine whether nuclear counts are returned
 
         Returns:
-            xr_counts: xarray containing segmented data of cells x markers"""
-
-    if type(input_images) is not xr.DataArray:
-        raise ValueError("Incorrect data type for ground_truth, expecting xarray")
```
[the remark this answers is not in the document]
ngreenwald: That was the plan, looks like I never actually moved them over.

PR #126, opened 2020-07-10
Integrating synthetic data generation with spatial testing functions

alex-l-kong:
We now have a solid foundation for distance-based random generation of distance matrices, initialized directly or from centroids. Now, we can integrate this into the spatial analysis testing code.

This might require a few changes to be made to the synthetic_spaitial_datagen code. This is because it needs to be formatted so it returns a compatible type that the functions in spatial_analysis_test require. This PR will address those changes as well as general integration.


alex-l-kong:
In addition to integrating synthetic_spatial_datagen with spatial_analysis_test, I've also added random seed initialization to test functions as well as added a few extra checks to synthetic_spatial_datagen's generate_random_centroids to ensure the data falls within acceptable parameters.

ackagel:
@ngreenwald, I agree in general.  In hindsight, having a set seed is not good practice for test functions, mostly because it opens the door to 'passing-seed hunting'.  But it's important to know the two main use cases for randomness in test functions.

If you're testing a function which should be guaranteed to work, no matter what the random input is, then unseeded is definitely better, if only for disallowing the 'oops, only this seed passes' problem.

If you're testing a monte-carlo method (e.g computing pi from the ratio of simulated dart-throws inside/outside a circle inscribed in a square), you would generally repeat an unseeded test several times, and assert something about the resulting statistics (e.g result has error < 0.01 at least 90% of the time).

From what I've heard about our test functions, a few of them sometimes pass, and sometimes fail (causing some agitation when pushing finished code).  If this is expected behavior, then setting their seed does still verify that the function logically works (syntactically, etc.), but it doesn't verify that it works as expected.  So to verify behavior, treating the function like a monte-carlo method is much more appropriate, using repeated tests for a statistic-based assertion.

TL;DR adam almost sold the cow for magic seeds.  If the function sometimes failing its individual test is ok/expected, then loop the test and just assert a statistic.  Otherwise, no seeds pls.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -138,35 +139,158 @@ def point_init_dist_matrix(size_img=(1024, 1024), num_A=100, num_B=100, num_C=10
...
+           len(total_points[:, 0][total_points[:, 0] < 0]) == 0 and \
+           len(total_points[:, 1][total_points[:, 1] > size_img[1]]) == 0 and \
+           len(total_points[:, 1][total_points[:, 1] < 0]) == 0:
 
-    # generate the label matrix for the image now
-    label_mat = label(binary_mat)
+            non_dup_array = True
+        else:
+            if seed:
+                np.random.seed(seed + 1)
```
ackagel, line 168: I don't really see the need for the big while loop here.  If we don't really need exactly `num_X` cells of type X, then just removing duplicates (already done with `np.unique`) and removing out of bounds points is fine.  Also, using `np.abs` here makes the distributions kinda wonky (reflections off of x & y axis), so just remove any negative points.  

If exactly `num_X` cells of type X are needed, this is extra extra arduous, and currently isn't happening (i.e `np.unique` is called when initing `a_points`, which removes duplicates).  

If the former is the case, `num_X` could be renamed to `max_num_X` or something like that

alex-l-kong:
@ngreenwald @vacuousplanet yeah I agree with your comments about random seeding now. I'll remove those in my next commit.

@ngreenwald I can remove the point-wise distance matrix tests as well. I thought to include them at first because if in certain cases people need to generate the centroids first, then pass those centroids into a distance matrix generating function, it seemed reasonable to include a test to verify that this method of generating distance matrices was indeed correct in the first place. But I can see what you're saying about just testing on the distance matrix itself. After all, the end result is going to be a distance matrix anyway. So all we need to do is to just verify that the final distance matrix is correct and not whatever algorithms are used to generate the distance matrix if I understand you correctly.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -74,22 +75,20 @@ def direct_init_dist_matrix(num_A=100, num_B=100, num_C=100,
...
-def point_init_dist_matrix(size_img=(1024, 1024), num_A=100, num_B=100, num_C=100,
-                           distr_A={'centroid_factor': (0.5, 0.5), 'cov': [[200, 0], [0, 200]]},
-                           distr_B={'centroid_factor': (0.9, 0.9), 'cov': [[200, 0], [0, 200]]},
-                           distr_C={'centroid_factor': (0.4, 0.4), 'cov': [[200, 0], [0, 200]]},
-                           seed=None):
+def generate_random_centroids(size_img=(1024, 1024), num_A=100, num_B=100, num_C=100,
+                              distr_A=((0.5, 0.5), [[200, 0], [0, 200]]),
+                              distr_B=((0.6, 0.6), [[200, 0], [0, 200]]),
+                              distr_C=((0.1, 0.1), [[200, 0], [0, 200]]),
+                              float_type=False, seed=None):
```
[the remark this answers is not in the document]
alex-l-kong: Yeah, seems like I never found a use for np.float64 in the end. So I'll remove this arg and just default to np.int16.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -74,22 +75,20 @@ def direct_init_dist_matrix(num_A=100, num_B=100, num_C=100,
...
 
-def point_init_dist_matrix(size_img=(1024, 1024), num_A=100, num_B=100, num_C=100,
-                           distr_A={'centroid_factor': (0.5, 0.5), 'cov': [[200, 0], [0, 200]]},
-                           distr_B={'centroid_factor': (0.9, 0.9), 'cov': [[200, 0], [0, 200]]},
-                           distr_C={'centroid_factor': (0.4, 0.4), 'cov': [[200, 0], [0, 200]]},
-                           seed=None):
+def generate_random_centroids(size_img=(1024, 1024), num_A=100, num_B=100, num_C=100,
+                              distr_A=((0.5, 0.5), [[200, 0], [0, 200]]),
+                              distr_B=((0.6, 0.6), [[200, 0], [0, 200]]),
+                              distr_C=((0.1, 0.1), [[200, 0], [0, 200]]),
```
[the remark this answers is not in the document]
alex-l-kong: Oof, thanks for checking that! In hindsight, it looks bad to compartmentalize these complicated arguments for multivariate_normal into one tuple. Probably better to split up as you said, or just set cov to None in the tuple.

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: Looking back, I realize the main motivation for the while loop was to avoid the hassle of having to redo Jay's testing function, specifically make_expression_matrix. If you leave points out, then Jay's tests are no longer valid because you don't have the same number of cell centroids in the distance matrix as you do in the expression matrix.

However, with Noah's point about not needing to test this function, it's probably safe to leave points out. Definitely makes life easier!

All that hard work to make things just right and the simple solution ended up working best in the end 😂.

ngreenwald:
You're totally right. It is important to be able to generate centroids with a given distribution, so that people can use them for testing. However, we want to try and keep the tests decoupled as much as possible. 

If we want to know if a given function accurately extracts the relationships present in a distance matrix, we should just generate a distance matrix. 

If we want to know if a function accurately extracts the relationships present in a set of centroids, then we should generate the centroids. 

If we want to know if our function to generate centroids is correct in the first place, we should create a separate test specifically for the synthetic data generation function itself. 

But having all of them linked together I think adds too much complexity.

segmentation/utils/synthetic_spatial_datagen.py
ngreenwald: The other point of this PR is to see if there are any changes that need to be made to the way Jay implemented these functions. If it would make our life easier to change the assumptions about data format, indexing, etc, those are all on the table for this. Don't feel like you have to code around whatever initial choices were made, nothing is set in stone here, especially if it will reduce the complexity of the stuff we have to write on top.

alex-l-kong:
The changes I've made include:

- Removing the point-wise matrix tests inside spatial_analysis_test.

- Parameter passing fix: prevent mutable objects from being passed.

- Simpler code: because we don't need to test the point-wise matrix function inside spatial_analysis_test, I removed a lot of the complexity and now it's pretty simple.

- Removing random seeding: the code now generates a completely random matrix on each run.

ngreenwald:
Okay cool, can you tag us for re-review once you're ready?

segmentation/utils/spatial_analysis_test.py
```diff
@@ -183,17 +175,18 @@ def test_calculate_channel_spatial_enrichment():
...
 
     # Test z and p values
-
     marker_thresholds = make_threshold_mat()
 
-    # Positive enrichment
-    all_data_pos = make_expression_matrix("positive")
-    dist_mat_pos = make_distance_matrix("positive")
+    # Positive enrichment with direct matrix initialization
+    all_data_pos = make_expression_matrix(enrichment_type="positive")
```
ngreenwald, line 181: We can remove the now-redundant lines here

segmentation/utils/spatial_analysis_test.py
```diff
@@ -202,14 +195,17 @@ def test_calculate_channel_spatial_enrichment():
     assert stats.loc["Point9", "p_pos", 3, 2] < .05
     assert stats.loc["Point9", "p_neg", 3, 2] > .05
     assert stats.loc["Point9", "z", 3, 2] > 0
-    # Negative enrichment
+
+    # Negative enrichment with direct matrix initialization
     all_data_neg = make_expression_matrix("negative")
-    dist_mat_neg = make_distance_matrix("negative")
+    dist_mat_neg_direct = make_distance_matrix("negative")
+    dist_mat_neg_point = make_distance_matrix("negative")
```
ngreenwald, line 202: Can remove redundant line here

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -111,7 +111,8 @@ def test_compute_close_cell_num():
 
     example_closenum, m1, m2 = spatial_analysis_utils.compute_close_cell_num(
         dist_mat=example_dist_mat, dist_lim=100, num=marker_num, analysis_type="Channel",
-        fov_data=all_data, fov_channel_data=fov_channel_data, thresh_vec=thresh_vec)
+        fov_data=all_data, fov_channel_data=fov_channel_data, thresh_vec=thresh_vec,
+        seed=42)
```
ngreenwald, line 115: Let's remove the seed unless it is causing problems

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -130,7 +131,8 @@ def test_compute_close_cell_num():
     thresh_vec = example_thresholds.iloc[0:20, 1]
     example_closenum, m1, m2 = spatial_analysis_utils.compute_close_cell_num(
         dist_mat=example_dist_mat, dist_lim=100, num=marker_num, analysis_type="Channel",
-        fov_data=all_data, fov_channel_data=fov_channel_data, thresh_vec=thresh_vec)
+        fov_data=all_data, fov_channel_data=fov_channel_data, thresh_vec=thresh_vec,
+        seed=42)
```
ngreenwald, line 135: Same here

segmentation/utils/spatial_analysis_utils_test.py
```diff
@@ -161,7 +163,7 @@ def test_compute_close_cell_num_random():
 
     example_closenumrand = spatial_analysis_utils.compute_close_cell_num_random(
         marker1_num, marker2_num, example_distmat, marker_num, dist_lim=100,
-        bootstrap_num=100)
+        bootstrap_num=100, seed=42)
```
ngreenwald, line 166: and here

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -7,33 +7,34 @@
...
 
 
 def direct_init_dist_matrix(num_A=100, num_B=100, num_C=100,
-                            distr_AB={'mean': 100, 'var': 1}, distr_AC={'mean': 20, 'var': 1},
+                            distr_AB=(10, 1), distr_random=(200, 1),
                             seed=None):
     """
     This function will return a random dist matrix such that the distance between cells
-    of types A and B are overall larger than the distance between cells of types A and C
+    of types A and B are overall shorter than the distance between cells of types A and C
```
ngreenwald, line 23: We can update the docstring here to say that it allows the user to specify one set of distances between cells of pops A and B, and a second set of distances between all other cell pairings.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -136,37 +132,145 @@ def point_init_dist_matrix(size_img=(1024, 1024), num_A=100, num_B=100, num_C=10
...
     sample_img_xr = xr.DataArray(sample_img,
                                  coords=[[1], range(size_img[0]), range(size_img[1]), ['segmentation_label']],
                                  dims=['fovs', 'rows', 'cols', 'channels'])
 
     # and return the xarray to pass into calc_dist_matrix
-    return sample_img_xr
+    return sample_img_xr, centroid_indices
+
+
+def generate_random_cell_shapes(size_img=(1024, 1024), num_A=100, num_B=100, num_C=100,
```
ngreenwald, line 225: This should go in a different PR.

segmentation/utils/visualize.py
```diff
@@ -18,3 +20,17 @@ def visualize_z_scores(z, pheno_titles):
     zplot = pd.DataFrame(z, columns=pheno_titles, index=pheno_titles)
     sns.set(font_scale=.7)
     sns.clustermap(zplot, figsize=(8, 8), cmap="vlag")
+
+
+def draw_ellipsoids(centroid_info):
```
ngreenwald, line 25: Same here

segmentation/utils/spatial_analysis_test.py
```diff
@@ -49,13 +45,9 @@ def make_distance_matrix(enrichment_type):
...
         # markers that are not located near each other (not within the dist_lim).
-        dist_mat_neg = np.zeros((60, 60))
-        dist_mat_neg[20:40, :20] = 300
-        dist_mat_neg[:20, 20:40] = 300
-        dist_mat_neg[40:50, :40] = 50
-        dist_mat_neg[:40, 40:50] = 50
-        dist_mat_neg[50:60, :50] = 200
-        dist_mat_neg[:50, 50:60] = 200
+
+        dist_mat_neg = synthetic_spatial_datagen.direct_init_dist_matrix(num_A=20, num_B=20, num_C=20,
```
ngreenwald, line 49: Why does this pass our tests? Doesn't this make it so that all cells are equally spaced away from one another? This shouldn't result in any negative enrichment

alex-l-kong: I must be misunderstanding what negative enrichment is. I made the assumption that it meant that the cells of type 1 and type 2 are on average farther away from each other than dist_lim according to the documentation under the `enrichment_type == 'negative'` case in make_distance_matrix. So, I didn't think that the insignificant marker cells (which I'll call type 3) mattered. However, I'm guessing more is involved.

The tests for negative enrichment, in a nutshell, seem dependent on the fact that there are on average an equal number or fewer number of cells within the dist_lim of close_num as compared to close_num_rand. In theory, and if my understanding is correct, this means that if the negative enrichment test matrices are initialized such that the mean distance between type 1 and type 2 cells is about the same as type 1 and type 3 cells, we should be seeing the desired results.

A quick fix could be changing distr_random in the negative matrix initialization to have a lower mean than that of distr_AB, but as mentioned, it's quite possible more is involved that I missed.

ngreenwald: That's close, but not quite right. The dist_lim determines whether a cell is close to another cell. In positive enrichment, we are looking for cell types A and B to have more cells that are close to one another than would be expected by chance. 

For negative enrichment, we are looking for cell types A and B to have fewer cells close to one another than would be expected by chance. 

So in Jay's original hardcoded data, cell types A and B only had 1 or two cells that were close to one other out of 20, but the overall population prevelance of cells that were close to one another was around 20%. Hence, they were negatively enriched. 

You put that for negative enrichment, the tests check that 'there are on average an equal number or fewer number of cells within the dist_lim of close_num as compared to close_num_rand.' If this is the case, we should change the tests, as what this is supposed to be testing is strict inferiority, not less than or equal to.

Does that all make sense?

alex-l-kong: Ok, I see. I'll change the test to in spatial_analysis_utils to reflect that. Also will change the distribution parameters in the negatively-enriched matrix to ensure cell types A and B have fewer cells close to one another than expected.

segmentation/utils/spatial_analysis_test.py
```diff
@@ -177,113 +172,129 @@ def make_expression_matrix(enrichment_type):
...
     # Test both Point8 and Point9
-    assert stats.loc["Point8", "p_pos", 2, 3] < .05
-    assert stats.loc["Point8", "p_neg", 2, 3] > .05
-    assert stats.loc["Point8", "z", 2, 3] > 0
-
-    assert stats.loc["Point9", "p_pos", 3, 2] < .05
-    assert stats.loc["Point9", "p_neg", 3, 2] > .05
-    assert stats.loc["Point9", "z", 3, 2] > 0
-    # Negative enrichment
+    assert stats_pos.loc["Point8", "p_pos", 2, 3] < .05
```
ngreenwald, line 196: Would be good to add some comments about what 3 and 2 are indexing to make this less cryptic to read

segmentation/utils/spatial_analysis_test.py
```diff
@@ -177,113 +172,129 @@ def make_expression_matrix(enrichment_type):
...
     # Test both Point8 and Point9
-    assert stats.loc["Point8", "p_neg", 2, 3] < .05
-    assert stats.loc["Point8", "p_pos", 2, 3] > .05
-    assert stats.loc["Point8", "z", 2, 3] < 0
+    assert stats_neg.loc["Point8", "p_neg", 2, 3] < .05
+    assert stats_neg.loc["Point8", "p_pos", 2, 3] > .05
+    assert stats_neg.loc["Point8", "z", 2, 3] < 0
+
+    assert stats_neg.loc["Point9", "p_neg", 3, 2] < .05
+    assert stats_neg.loc["Point9", "p_pos", 3, 2] > .05
```
ngreenwald, line 220: These are the same negative enrichment tests as before. That means when the negative enrichment matrix was getting created last time where it was actually a neutral matrix, the tests were still passing, right? That means we have a bug somewhere

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -7,35 +7,38 @@
...
+from random import randint
 from segmentation.utils import spatial_analysis_utils as sau
+from segmentation.utils import visualize as viz
 from scipy.spatial.distance import cdist
 from skimage.measure import label
 
 
 def direct_init_dist_matrix(num_A=100, num_B=100, num_C=100,
-                            distr_AB={'mean': 100, 'var': 1}, distr_AC={'mean': 20, 'var': 1},
+                            distr_AB=(10, 1), distr_random=(200, 1),
```
ngreenwald, line 19: Now that we only have one function that actually makes a distance matrix, I think we can give this a more interpretable name. 

Something like create_synthetic_dist_matrix, generate_test_distance_matrix, etc etc.

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -136,37 +134,91 @@ def point_init_dist_matrix(size_img=(1024, 1024), num_A=100, num_B=100, num_C=10
...
     # this ensures that we only keep the points that are not duplicate across different cell types
-    points, counts = np.unique(np.concatenate((a_points, b_points, c_points), axis=0), axis=0, return_counts=True)
-    non_dup_points = points[counts == 1]
+    non_dup_points, non_dup_counts = np.unique(total_points, axis=0, return_counts=True)
+    total_points = non_dup_points[non_dup_counts == 1]
+
+    return total_points
+
+
+def point_init_dist_matrix(size_img=(1024, 1024), num_A=100, num_B=100, num_C=100,
```
ngreenwald, line 157: This function doesn't actually create a distance matrix anymore, so we should update the name to reflect its functionality. Something like generate_synthetic_label_map or whatever you think is appropriate.

segmentation/utils/spatial_analysis_test.py
```diff
@@ -33,13 +34,10 @@ def make_distance_matrix(enrichment_type):
...
-        dist_mat_pos[10:20, :10] = 50
-        dist_mat_pos[:10, 10:20] = 50
-        dist_mat_pos[20:40, :20] = 200
-        dist_mat_pos[:20, 20:40] = 200
-        dist_mat_pos[40:80, :40] = 300
-        dist_mat_pos[:40, 40:80] = 300
+
+        dist_mat_pos = synthetic_spatial_datagen.direct_init_dist_matrix(num_A=10, num_B=10, num_C=60,
+                                                                         distr_AB=(int(dist_lim / 5), 1),
+                                                                         distr_random=(int(dist_lim * 5), 1))
```
ackagel, line 40: Idk if the test functions are randomly failing on asserts about `p_pos`/`p_neg`/z-scores, but if they are, then increasing the amount of `num_X` should decrease how often those tests fails.  Testing on 10 As and 10 Bs seems a little small, i.e susceptible to significant sampling variations.

This goes for the other cases in `make_distance_matrix` as well

segmentation/utils/spatial_analysis_test.py
alex-l-kong: Those assertions ran on the old assumption of less than or equal to. When I updated the tests for strict inferiority, they don't pass when the same distribution is used to generate the neutral matrix (as in, the distribution parameters are equal for both AB and random).

segmentation/utils/spatial_analysis_test.py
alex-l-kong: We can definitely up the ante in the future. I've added an issue to address the current hard-coding of expression matrices (currently assuming 80 entries per FOV in positive enrichment and 60 cells per FOV in negative enrichment). After we come up with a scheme to synthetically generate these as well, it might be easier to generate variable-length testing dist and expression matrices and thus test on more data.

segmentation/utils/spatial_analysis_utils.py
```diff
@@ -229,7 +230,7 @@ def calculate_enrichment_stats(close_num, close_num_rand):
             z[j, k] = (close_num[j, k] - muhat[j, k]) / sigmahat[j, k]
             # Calculate both positive and negative enrichment p values
             p_pos[j, k] = (1 + (np.sum(tmp >= close_num[j, k]))) / (bootstrap_num + 1)
-            p_neg[j, k] = (1 + (np.sum(tmp <= close_num[j, k]))) / (bootstrap_num + 1)
+            p_neg[j, k] = (1 + (np.sum(tmp < close_num[j, k]))) / (bootstrap_num + 1)
```
alex-l-kong, line 233: With regards to the change in the testing of p_neg, changing this from <= to < does the trick for neutral matrices (aka those with the same distribution parameters for distr_AB and distr_random).

segmentation/utils/spatial_analysis_test.py
ngreenwald: Okay, these are the indices of the columns that contain the markers we care about: https://github.com/angelolab/segmentation/blob/4a7280af7dab2d83cbfe8b3eb5977bf00937275f/segmentation/utils/spatial_analysis_test.py#L86

Can you add a comment to that effect?

segmentation/utils/spatial_analysis_utils.py
ngreenwald: @Jaiveers21, in the original matlab code, was this comparison >= only for positive, for both, or for neither?

I think the mathematically correct comparison here would be strict >

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -7,35 +7,38 @@
...
+    types A and B and between cells of all other groups (type C).
 
     Each row and column representing a cell.
     We generate the points using Gaussian distributions
     Ideally, the parameters for A to B distances will be set such that they produce a lower range of values
     than A to C distances.
 
-    Will return a random distance matrix as well as the dictionary of associated cell: label IDs
-    The above is generated from the generate_labels function
+    Note that these distance matrices created are non-Euclidean.
```
ngreenwald, line 30: 😆

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -74,59 +77,54 @@ def direct_init_dist_matrix(num_A=100, num_B=100, num_C=100,
...
     """
-    This function generates random centroid centers in the form of a label map
-    such that those of type A will have centers closer on average to those of type B
-    than those of type C
-
-    We will use a multivariate Gaussian distribution for A, B, and C type cells to generate their respective centers.
-
-    Returns the set of points associated with the centroids of cells of types A, B, and C.
+    Generate a set of random centroids given distribution parameters.
+    Used as a helper function by point_init_dist_matrix and generate_random_cell_shapes.
```
ngreenwald, line 91: update this to new name

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -136,37 +134,91 @@ def point_init_dist_matrix(size_img=(1024, 1024), num_A=100, num_B=100, num_C=10
     # because we're passing these into skimage.measure.label, it is important
     # that we convert these to integers beforehand
     # since label only takes a binary matrix
-    # we pass the result through the unique function to eliminate any possibility of duplicate points
-    # appearing within any of these arrays
-    a_points = np.unique(np.random.multivariate_normal(a_mean, a_cov, num_A).astype(np.int16), axis=0)
-    b_points = np.unique(np.random.multivariate_normal(b_mean, b_cov, num_B).astype(np.int16), axis=0)
-    c_points = np.unique(np.random.multivariate_normal(c_mean, c_cov, num_C).astype(np.int16), axis=0)
+    a_points = np.random.multivariate_normal(a_mean, a_cov, num_A).astype(np.int16)
```
ngreenwald, line 137: This is super clean!!! Really nice work

segmentation/utils/synthetic_spatial_datagen.py
```diff
@@ -136,37 +134,91 @@ def point_init_dist_matrix(size_img=(1024, 1024), num_A=100, num_B=100, num_C=10
...
+    all_centroids = \
+        generate_random_centroids(size_img=size_img, num_A=num_A, num_B=num_B, num_C=num_C,
+                                  mean_A_factor=mean_A_factor, cov_A=cov_A,
+                                  mean_B_factor=mean_B_factor, cov_B=cov_B,
+                                  mean_C_factor=mean_C_factor, cov_C=cov_C,
+                                  seed=seed)
+
+    point_x_coords, point_y_coords = zip(*all_centroids)
+
+    # to compute the indices of the original centroid locations, needed to adjust the distance matrix
```
ngreenwald, line 204: I'm not sure I understand this comment

segmentation/utils/synthetic_spatial_datagen_test.py
```diff
@@ -10,7 +10,7 @@
 importlib.reload(synthetic_spatial_datagen)
 
 
-def test_direct_init_dist_matrix():
+def test_generate_test_dist_matrix():
```
ngreenwald, line 13: Can you add a simple test for the other two functions similar to what's here?
Doesn't have to be fancy, even a single assert is fine. Mostly just to catch any potential syntax errors/input dimension errors that could get accidentally introduced otherwise.

segmentation/utils/synthetic_spatial_datagen.py
ackagel: ( ▀ ͜͞ʖ▀)

we have now entered big nerd hours

segmentation/utils/spatial_analysis_utils.py
Jaiveers21: In the matlab, both had the or equal to

segmentation/utils/synthetic_spatial_datagen_test.py
alex-l-kong: Ok, got it.

segmentation/utils/synthetic_spatial_datagen.py
alex-l-kong: Ah yes, need to make this clearer.

Basically, on the call to regionprops in calc_dist_matrix, the function automatically orders the centroids in ascending order regardless of the original label assigned to them by the label function (or whatever method you use to label the centroids). So for example, if you labeled the centroid (1, 1) as number 75 in the corresponding label map for dist_mat_pos, regionprops would automatically put that as the 0th element instead. Do this for all centroids and you've got a really jumbled distance matrix that would not reflect the first m columns being marker 1 cells, the next n columns being marker 2 cells, and the final p columns being random marker cells.

I didn't want to mess around with Jay's calc_dist_matrix function too much, so the easy fix is to compute the original locations (aka indices) of where the centroids should go (the order in which they appear in all_centroids of generate_test_label_map, I intentionally have first the a, then the b, lastly the c centroids). And that's the beauty of lexsort, knowing that regionprops sorts the centroids in ascending order, first by x-coord, then y-coord. lexsort will essentially return the indices needed to reorder a numpy array altered by np.sort back to what it was originally. That way, we can reorder the rows and columns of the distance matrix to accurately reflect the partitioning we desire.

ngreenwald:
Sorry, totally blanked on this. Will respond later today

PR #127, opened 2020-07-10
Reworked bad-path explainer

ackagel:
Revised the bad-path explainer to version suggested in issue #113.  Also revised accompanying tests to match the new format.


segmentation/utils/data_utils.py
```diff
@@ -18,33 +18,24 @@ def validate_paths(paths):
...
                         raise ValueError(
-                            f'The path, {pathlib.Path(path).resolve()}, '
-                            f'exists outside of Docker\'s scope.\n'
-                            f'Move data, files, etc from {pathlib.Path(path).resolve()} '
-                            f'into the segmentation folder.\n'
-                            f'Reference moved data using relative paths, '
-                            f'which go no higher in scope than the segmentation folder')
-
-            raise ValueError(f"The folder path {pathlib.Path(path).resolve()}, does not exist...")
+                            f'The folder, {parent}, was not found...')
```
ngreenwald, line 31: Because you're using the pathlib.parents function to generate the sub_paths to check for existence, the paths are checked in descending order from longest to shortest. 

i.e., for the following path:  'data/sub_dir1/typo_dir/sub_dir3/file'
We would first check 'data/sub_dir1/typo_dir/sub_dir3'
then 'data/sub_dir1/typo_dir'
then 'data/sub_dir1'

However, even the first path won't exist, since it's nested under a fake subdir. What would be great is if we identified the specific directory that is causing the typo in the string. So we'll want to loop over the pathlib.parents in the opposite order, so we reach a failure only when the typo is introduced. Then, we can print out just that element of the path, along with the true sub_dir that it belongs to, in order to help guide debugging.

segmentation/utils/data_utils_test.py
```diff
@@ -27,27 +28,30 @@ def _create_img_dir(temp_dir, fovs, imgs, img_sub_folder="TIFs"):
 
 
 def test_validate_paths():
-    valid_path = '.'
 
-    valid_parts = [p for p in pathlib.Path(valid_path).resolve().parts]
-    valid_parts[1] = 'not_a_real_directory'
+    valid_path = 'data/example_dataset/label_data'
```
ngreenwald, line 32: This test will fail if we decide to reorganize the directory. Let's create a synthetic directory to use during testing with tmpdir.

segmentation/utils/data_utils_test.py
```diff
@@ -27,27 +28,30 @@ def _create_img_dir(temp_dir, fovs, imgs, img_sub_folder="TIFs"):
...
+    valid_parts[1] = 'not_a_real_subdirectory'
+    wrong_subdir = os.path.join(*valid_parts)
 
-    nonexistant = './not_a_real_directory'
+    wrong_file = os.path.join(valid_path, 'not_a_real_file.tiff')
 
     data_utils.validate_paths(valid_path)
 
-    with pytest.raises(ValueError, match=r".*starts*"):
+    with pytest.raises(ValueError, match=r".*prefixed*"):
```
ngreenwald, line 47: In line with my comment about the functionality itself, we should check here that the correct element of the path was identified as the typo. So having an example where the last sub_dir is the erroneous one, a sub_dir in the middle, etc.

segmentation/utils/data_utils_test.py
```diff
@@ -27,27 +28,30 @@ def _create_img_dir(temp_dir, fovs, imgs, img_sub_folder="TIFs"):
...
+    valid_path = 'data/example_dataset/label_data'
+
+    valid_parts = [p for p in pathlib.Path(valid_path).parts]
+    valid_parts[0] = 'not_a_real_directory'
 
     starts_out_of_scope = os.path.join(*valid_parts)
 
-    entirely_out_of_scope = '/not_a_real_directory/somewhere_else_not_here'
+    valid_parts[0] = 'data'
+    valid_parts[1] = 'not_a_real_subdirectory'
```
alex-l-kong, line 40: Going off of Noah's comment about the potential for the directory being reorganizing, some of the indexing here might need to be changed. Admittedly, I'm not too experienced with tempfile, but one potential way to ensure we're not accessing anything out of bounds is to get the length of valid_parts after it runs on the TemporaryDirectory path we create.

ackagel:
test failed because of spatial enrichment randomness

segmentation/utils/data_utils.py
ngreenwald: It looks to me like this test would still fail in the newest version, is that not the case? We're looping through the parents, meaning any path containing a mis-typed file will trigger on the last sub_dir

ackagel:
I checked in the previous docker image (w/ pytest), but I'll double check it in the new docker image and I'll also try running the notebooks

ackagel:
also, the '/data' version was not willing to cooperate, which is why I went with the 'data' version.

ngreenwald:
Ah, I think maybe I see the issue. 

Currently, the paths are constructed as '../data/xxx', which I think simplifies to data/xxx. 

However, I think we can change all of those relative paths to an absolute path that is '/data/xxx', since that is clearer. Can you update the example notebooks to have that path start instead? And then check if that new format passes validation?

ngreenwald:
What do you mean the /data version didn't cooperate?

ackagel:
sorry, that was kind of vague.  I mean the '/data' version insisted on looking for the 'data' folder as a subdirectory of the root '/' directory.

Also, right now, the notebooks work because the files all exist, but when they don't, the explanations are way off, because the explainers assume the cwd is the 'segmentation' directory (where 'data' is meaningful).  So when the cwd is the scripts directory, 'data' folder isn't seen.  

I'm not entirely sure how you got the '/data' version to work in your post in the issue #113 thread; let's chat over slack tho since this thread is getting unwieldy.

PR #132, opened 2020-07-22
Deepcell format update

ackagel:
Now that the deepcell website is up, there have been some changes to the input/output formatting to/from deepcell.  Changes in this PR include:

- tif output in preprocessing notebook
- merging mibitiff_preprocessing functionality into regular preprocessing nb
- altered channel summing
- updated postprocessing notebook to support legacy (xarray into watershed) and updated (label-map tifs) uploading
- added test functions to cover new functions/functionalities

Legacy support can be dropped when no more lab users have any deepcell output xr's to process


ackagel:
it looks like there's a problem with how coverage is being computed?  It computes coverage in `/segmentation` and then computes the coverage within travis's virtual environment (which is drastically lower for some reason?).

ngreenwald:
I have no idea wtf is going on here. I restarted travis and same issues came up. I wonder if this is related to us changing how the repo gets installed in the last PR?

alex-l-kong:
I've run into similar issues with my branches in the past. Oftentimes, my coverage issues would go away after a couple of commits, but I wouldn't rely on that as a permanent solution.

scripts/Deepcell_Postprocessing.ipynb
```diff
@@ -42,14 +42,17 @@
    "source": [
     "# set up file paths\n",
     "base_dir = \"../data/example_dataset/\"\n",
-    "deepcell_dir = os.path.join(base_dir, \"deepcell_output\")\n",
-    "label_dir = os.path.join(base_dir, 'label_data')\n",
+    "old_deepcell_dir = os.path.join(base_dir, \"old_deepcell_output\")\n",
+    "deepcell_dir = os.path.join(base_dir, 'deepcell_output')\n",
+    "\n",
+    "# use xarray (if you asked Noah to forward data to deepcell, set this to true)\n",
```
ngreenwald, line 48: We can go ahead and kill the legacy functionality. The only person in our lab who I've done this for recently in JP, and he's already made it to the extraction step, where the rest of the pipeline is the same.

scripts/Deepcell_Postprocessing.ipynb
```diff
@@ -165,9 +172,10 @@
    "outputs": [],
    "source": [
     "# load TIFFs of segmentation labels (from pixel clustering, deepcell, etc.)\n",
-    "segmentation_labels_preload = data_utils.tiffs_to_xr_labels(tiff_dir=label_dir, output_dir=None, delimiter='_')\n",
-    "plt.figure(figsize = (13, 13))\n",
-    "plt.imshow(segmentation_labels_preload.loc[\"Point8\", :, :, \"segmentation_label\"])"
+    "if not legacy_deepcell:\n",
+    "    segmentation_labels = data_utils.tiffs_to_xr_labels(tiff_dir=deepcell_dir, output_dir=None, delimiter='_')\n",
```
ngreenwald, line 176: I think we should change the delimiter to be` _feature_0.tif`. This will be robust to people having underscores in their FOV names

scripts/Deepcell_Postprocessing.ipynb
```diff
@@ -183,8 +191,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "plt.figure(figsize = (13, 13))\n",
-    "plt.imshow(plt.imread(os.path.join(segmentation_dir, \"Point8_HH3_Membrane_overlay.tiff\")))"
+    "# TODO: Viz for new deepcell format\n",
```
ngreenwald, line 194: Here's what this could look like. Either provide a list of point names, or if an empty list assume all points. 
It will then load the corresponding deepcell_input points and the segmentation in the output folder. Those two things will get fed into plot_utils.plot_overlay, and saved into the same folder with the deepcell_output segmentations.

scripts/Deepcell_Preprocessing.ipynb
```diff
@@ -45,15 +45,29 @@
...
+    "# membrane channel name(s) (or None)\n",
+    "mems = ['Membrane']\n",
+    "\n",
     "# validate data_dir\n",
     "data_utils.validate_paths(data_dir)\n",
     "\n",
     "# either get all points in the folder\n",
     "points = os.listdir(data_dir)\n",
     "points = [point for point in points if os.path.isdir(data_dir + point) and point.startswith(\"Point\")]\n",
+    "if MIBItiff:\n",
```
ngreenwald, line 64: This is nice, it's all here but not executed by default

scripts/Deepcell_Preprocessing.ipynb
```diff
@@ -62,40 +76,26 @@
...
+    "# generate and save deepcell input tifs\n",
+    "data_utils.generate_deepcell_input(data_xr, data_dir, nucs, mems)\n",
     "\n",
-    "tif_saves = [new_channel_name]\n",
-    "for point in summed_xr.fovs.values:\n",
-    "    for tif in tif_saves:\n",
-    "        save_path = os.path.join(data_dir, point, \"TIFs\", tif + \".tif\")\n",
-    "        io.imsave(save_path, summed_xr.loc[point, :, :, tif].values.astype(\"uint8\"))"
+    "# save xarray for running through deepcell\n",
+    "data_xr.to_netcdf(os.path.join(data_dir, \"deepcell_input.xr\"), format=\"NETCDF3_64BIT\")"
```
ngreenwald, line 90: You're right, we can just get rid of this part completely

segmentation/utils/data_utils.py
```diff
@@ -125,6 +125,14 @@ def load_imgs_from_tree(data_dir, img_sub_folder=None, fovs=None, imgs=None,
 
         # if taking all imgs from directory, sort them alphabetically
         imgs.sort()
+    # otherwise, fill channel names with correct file extension
+    # TODO: test file-extension agnostic functionality
+    elif not all([img.endswith(("tif", "tiff", "jpg", "png")) for img in imgs]):
```
ngreenwald, line 130: This is nice

segmentation/utils/data_utils.py
```diff
@@ -233,16 +241,17 @@ def load_imgs_from_dir(data_dir, image_name='img_data', delimiter='_',
     fovs = [img.split(delimiter)[0] for img in imgs]
 
     img_xr = xr.DataArray(img_data, coords=[fovs, row_coords, col_coords, [image_name]],
-                          dims=["fovs", "rows", "cols", "channels"])
+                          dims=["fovs", "rows", "cols", "compartments"])
```
ngreenwald, line 244: Do we use this function for anything else? Changing the dim name is fine if it's only used for loading segmentation masks

ngreenwald: Actually, can have both the coordinate name and the dimension name be parameters of the function?

segmentation/utils/data_utils.py
```diff
@@ -267,6 +276,34 @@ def tiffs_to_xr_labels(tiff_dir, output_dir=None, delimiter='_'):
...
+        data_xr: xarray containing nuclear and membrane channels over many fov's
+        data_dir: location to save deepcell input tifs
+        nuc_channels: nuclear channels to be summed over
+        mem_channels: membrane channels to be summed over
+    Outputs:
+        Saves summed channels as multitiffs
+
+    """
+    for fov in data_xr.fovs.values:
+        out = np.zeros((data_xr.shape[1], data_xr.shape[2], 2), dtype="uint8")
```
ngreenwald, line 293: We should use the same dtype as the data_xr

segmentation/utils/data_utils.py
```diff
@@ -267,6 +276,34 @@ def tiffs_to_xr_labels(tiff_dir, output_dir=None, delimiter='_'):
...
+
+        # sum over channels and add to output
+        if nuc_channels:
+            out[:, :, 0] = np.sum(data_xr.loc[fov, :, :, nuc_channels].values.astype("uint8"),
```
ngreenwald, line 297: Same here. For microscopy data from other image platforms the dtypes can be much higher

segmentation/utils/data_utils_test.py
```diff
@@ -145,6 +145,17 @@ def test_load_imgs_from_tree():
         # make sure specified channels loaded
         assert np.array_equal(test_subset_xr.channels, some_chans)
 
+        # check loading w/o file extension
+        test_noext_xr = \
+            data_utils.load_imgs_from_tree(temp_dir, img_sub_folder="TIFs", dtype="int16",
```
ngreenwald, line 150: Can you add a test for when some of them have an extension and some don't

segmentation/utils/data_utils_test.py
```diff
@@ -161,6 +172,81 @@ def test_load_imgs_from_dir():
...
+        nucs = ['nuc2']
+        mems = None
+
+        data_utils.generate_deepcell_input(data_xr, temp_dir, nucs, mems)
+        fov1 = io.imread(fov1path)
+        fov2 = io.imread(fov2path)
+
+        assert np.all(fov1[:, :, 0] == 2)
+        assert np.all(fov1[:, :, 1] == 0)
+        assert np.all(fov2[:, :, 0] == 1)
```
ngreenwald, line 246: Looks solid

scripts/Deepcell_Preprocessing.ipynb
```diff
@@ -45,15 +45,29 @@
     "# load TIFs from GUI-based directory structure\n",
     "data_dir = \"../data/example_dataset/input_data/\"\n",
     "\n",
+    "# data format\n",
```
alex-l-kong, line 48: Might be helpful to include a comment to explain when to set MIBItiff to True for those with a non-technical background, similar to legacy_deepcell in Deepcell_Postprocessing.ipynb.

scripts/Deepcell_Preprocessing.ipynb
```diff
@@ -45,15 +45,29 @@
...
+    "mems = ['Membrane']\n",
+    "\n",
     "# validate data_dir\n",
     "data_utils.validate_paths(data_dir)\n",
     "\n",
     "# either get all points in the folder\n",
     "points = os.listdir(data_dir)\n",
     "points = [point for point in points if os.path.isdir(data_dir + point) and point.startswith(\"Point\")]\n",
+    "if MIBItiff:\n",
+    "    points = [point for point in points if point.endswith(MIBItiff_suffix)]\n",
```
alex-l-kong, line 65: Feel free to chime in on this @ngreenwald, but I think it would be nice to split up some of these cells. For example, we have the MIBItiff variable assignments in one cell, the nuclear and membrane channel variable assignments in another, the validation of data_dir in another, etc. It can make the process easier to follow and potentially help with isolating buggy code.

alex-l-kong:
Sorry Adam, forgot to add my comments as a review haha.

Anyways, not much more to add outside of Noah's comments. Just a couple things I thought might be useful to help with making the code easier to follow for non-technical users.

segmentation/utils/data_utils.py
ackagel: I'm pretty sure this is just for loading segmentation masks.  The dim used to be called 'channels' but it looks like somebody updated that at some point, which went unnoticed since I forgot to actually try using `tiffs_to_xr_labels` in the nb's or test its use in plotting functions, etc. within a test function.

ackagel: Regardless, I can add the option to also name the dim in the `load_imgs_from_dir` function

scripts/Deepcell_Postprocessing.ipynb
```diff
@@ -165,26 +95,46 @@
...
    "outputs": [],
    "source": [
-    "plt.figure(figsize = (13, 13))\n",
-    "plt.imshow(plt.imread(os.path.join(segmentation_dir, \"Point8_HH3_Membrane_overlay.tiff\")))"
+    "\n",
+    "# get input data for overlay\n",
+    "input_data_xr = data_utils.load_imgs_from_multitiff([os.path.join(input_dir, point) for point in points_input])\n",
+    "\n",
+    "for fov in input_data_xr.fovs:\n",
+    "    plot_utils.plot_overlay(segmentation_labels.loc[fov, :, :, \"whole_cell\"].values,\n",
```
ngreenwald, line 125: Looks like some duplicated code in this cell. Let's keep the version that has the path specified so that the images get saved

scripts/Deepcell_Postprocessing.ipynb
```diff
@@ -165,26 +95,46 @@
    "outputs": [],
    "source": [
     "# load TIFFs of segmentation labels (from pixel clustering, deepcell, etc.)\n",
-    "segmentation_labels_preload = data_utils.tiffs_to_xr_labels(tiff_dir=label_dir, output_dir=None, delimiter='_')\n",
-    "plt.figure(figsize = (13, 13))\n",
-    "plt.imshow(segmentation_labels_preload.loc[\"Point8\", :, :, \"segmentation_label\"])"
+    "segmentation_labels = data_utils.tiffs_to_xr_labels(tiff_dir=label_dir, output_dir=None, delimiter='_feature_0')\n",
+    "if plot_labels:    \n",
+    "    for fov in segmentation_labels.fovs:\n",
+    "        plt.figure(figsize = (13, 13))\n",
```
ngreenwald, line 101: Does this work? I would assume a for loop over imshow would just overwrite the previous image, but I didn't test it out

scripts/Deepcell_Postprocessing.ipynb
ackagel: So, the code is dup'd because the path-specified version doesn't actually show the overlay, just saves it.  I'd pull up the saved images and show those, but some name/path reading logic is needed for that so I figured I'd go ahead and just run it again w/o saving to get the plots to show.

I could maybe just change `plot_utils.plot_overlay` to just show the plot anyways, or also take a bool to show plot or not

scripts/Deepcell_Postprocessing.ipynb
ackagel: yeah it might just overwrite the prev image.  I'll try deliberately creating new figs to send data to (might also have to change `plot_utils.plot_overlay` to take in an axes)

scripts/Deepcell_Postprocessing.ipynb
ngreenwald: I think given the number of different FOVs most people will have, it's fine to just save these as .tifs and let them inspect them one by one

scripts/Deepcell_Postprocessing.ipynb
ngreenwald: I think we can just remove this line completely. The overlays are going to be more informative.

ackagel:
Point8_deepcell_input.tif is used later by postprocessing for generating the overlays.  Should I still remove it?

segmentation/utils/data_utils.py
```diff
@@ -81,6 +81,51 @@ def load_imgs_from_mibitiff(mibitiff_paths, channels=None):
     return img_xr
 
 
+def load_imgs_from_multitiff(multitiff_paths, channels=None):
```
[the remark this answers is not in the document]
ackagel: same with load_imgs_from_mibitiff?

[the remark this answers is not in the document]
ngreenwald: Oh good call, I forgot about that. Yes please

Issue #135, opened 2020-07-28
Filename inconsistencies


cliu72:
When I upload input files labeled like "Pointx_deepcell_input.tif" to the DeepCell website, the website returns a file labeled like "Pointx_deepcell_input_feature_0.tif". This causes some errors in the postprocessing script, I think because it expects the files to be named "Pointx_feature_0.tif". I manually changed the file names and it was fine, but it might be better to have the website output the filename without "_deepcell_input", or have the postprocessing script handle it.

PR #137, opened 2020-07-29
Filename fix + dtype warnings

ackagel:
The is a patch for the file naming issue addressed in issue #135.  In particular, this PR:

 - reorganizes/renames some files+directories in the input_data folder
 - adds fov name delimiter auto-detection to the mibi+multi tiff loading function
 - changes a ValueError into a warning for int-expected float image loading
 - alters notebooks accordingly
 - will add more testing to cover new functionality


segmentation/utils/data_utils_test.py
```diff
@@ -14,9 +15,23 @@
 
 importlib.reload(data_utils)
 
+# required metadata for mibitiff writing (barf)
```
ngreenwald, line 18: 👍

segmentation/utils/data_utils_test.py
```diff
@@ -104,6 +120,49 @@ def test_load_imgs_from_mibitiff():
         data_xr.values[0],
         (tiff.read(os.path.join(data_dir, mibitiff_files[0])))[channels].data)
 
+    # this is looped to make sure temp_file doesn't have '_' in its name
+    # (sadly there's no excluded characters option for tempfile)
+    while True:
+        with tempfile.NamedTemporaryFile(suffix=".tif", dir=data_dir) as temp_file:
```
ngreenwald, line 126: I did some googling, you're right. 

What about renaming it instead? os.rename

ackagel: I could try it, but I think I would have to handle cleanup manually then.  Alternatively, I could create a temporary directory, and create two mibi tiffs (or copy one in input_data) to test in it, and then just delete the directory.  Now that I type it out, that seems better than literally gambling on no underscores, so I'll probably do that.

ackagel:
yeah I'm not sure.  Something got messed up since I had an editor open while pulling changes from master, and I haven't been able to get rid of those commits for some reason.  Thankfully, it shouldn't hurt anything and they should go away on my end once this PR is merged into master.

PR #144, opened 2020-08-03
Readme clarification

### LINKED ISSUE #143: Clearer instructions
We should update the readme to make it more obvious exactly which commands need to happen in what order. 
We should also update the preprocessing jupyter notebook to either make it more obvious how None should be specified for channels, or add more input validation to the load_imgs function.

ackagel:
Addresses requests for clarification in issue #143.  Specifically, this PR:

 - Updates and clarifies instructions in the readme
 - Clarifies nuclear/membrane input options in preprocessing script and filters out None's just in case
 - alters start_docker.sh to be compatible with file paths containing spaces

Closes #143


README.md
```diff
@@ -1,30 +1,52 @@
...
-To stop docker from running, enter control-c in the terminal window
+Be sure to keep this terminal open.  **Do not exit the terminal or enter control-c until you are finished with the notebooks**.
 
-The Jupyter Notebook has a deepcell-preprocessing and deepcell-postprocessing script. The preprocessing script is to generate the data that will be used for segmentation. The postprocessing script takes the output of deepcell and extracts single cell data from your imaging data. 
+## Once you are finished
+
+You can shut down the notebooks and close docker by entering control-c in the terminal window.
+
+## Updates
+
```
ngreenwald, line 47:
```suggestion
This repo is still in the development stage, and is updated frequently. To make sure that you will be able to easily get the updated version, do not modify the preprocessing or postprocessing notebooks directly. Instead, make sure to first copy the notebooks, rename them, and only modify these copied versions. Please see the instructions at the topic of the notebooks for step-by-step
```

start_docker.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 docker run -it \
   -p 8888:8888 \
-  -v $PWD/segmentation:/usr/local/lib/python3.6/site-packages/segmentation \
-  -v $PWD/scripts:/scripts \
-  -v $PWD/data:/data \
+  -v "$PWD/segmentation:/usr/local/lib/python3.6/site-packages/segmentation" \
+  -v "$PWD/scripts:/scripts" \
```
ngreenwald, line 5: Did you test this to make sure it works?

ackagel: Yep yep yep

README.md
ackagel: maybe instead we just keep the template notebooks in a hidden .templates folder?

Then, in `start_docker.sh` (or in the `Dockerfile`), we can make a scripts dir, copy the notebooks into scripts, and then run the docker.  That way, people can demolish the notebooks w/o worrying about making updating harder.  

Specifically, when they update the repo, `start_docker.sh`/`Dockerfile` can detect if the scripts dir exists, detect if there are already notebooks in there, and suffix the new copied nb's with something (like a number or 'new')

start_docker.sh
ackagel: do you know of any other somewhat common edge-case file/folder name issues we can catch here?  idk if having a `"` character in the path would break it or if it's even something common enough to worry about

ngreenwald: I think spaces are the main culprit I can think of.

README.md
ngreenwald: I like that idea. However, I don't think we want to overwrite it whenever start_docker.sh is run, since then it will erase any saved paths. We could put it in the dockerfile, and require a docker rebuild after each pull?

ngreenwald:
Moving forward we're going to have many people whose filenames don't start
with Point.

Is there another filter we can do to check for correct image files?

README.md
ackagel: Since `scripts` is mounted via `-v`, just adding this to `start_docker.sh` seems to work, with no need for rebuilding.
```
if [ -d "$PWD/scripts" ]
then
  for f in "$PWD"/.templates/*.ipynb
  do
    name=$(basename "$f")
    cp -- "$f" "$PWD/scripts/updated_$name"
  done
else
  mkdir "$PWD/scripts"
  cp "$PWD"/.templates/*.ipynb "$PWD/scripts/."
fi
```

I'll push a working (notebook tests etc) version with this and we can revise it for sure

ackagel:
I can just add a `point_prefix` option along with the `MIBItiff_suffix` option.  If it's set to None or empty, then it won't filter (which is probably fine), but then people can easily change it to whatever prefix they need

Side-note: we're already filtering for tifs/tiffs/jpgs/pngs within the loading functions

ngreenwald:
I think we can just get rid of the starts_with check completely. We're already checking that each 'Point' is a directory, so that should take care of any misc. files that are in there. Unless there's another reason that this could be useful?

README.md
ngreenwald: Right, the issue with this is that it makes a new copy every single time they start the docker, even when nothing has been updated

ackagel:
I don't think there is; just convention.  I'll see if things break when I take it away and clean up any weird dependencies.

README.md
ackagel: Yeah, but proof of concept wise, we can do this w/o rebuilding the docker image (I think I can fix this specific issue with the `diff` command)

ngreenwald: okay sweet

ngreenwald:
Also, can you add instructions about the upload process? Probably to the preprocessing jupyter notebook, but wherever you think makes the most sense. 

Basically that the tif files should be zipped (if there is more than one of them), and submitted to deepcell.org with the 'Multiplex' model selected

ackagel:
Looks like people will have to rebuild the docker image just for this update (not copying /scripts over on image-build since it might not exist until the container is run).

ngreenwald:
Also, can you remove the cells in the notebook that say 'don't modify this notebook'

PR #149, opened 2020-08-05
Abstract Deepcell_Postprocessing.ipynb

### LINKED ISSUE #148: Convert jupyter notebook cells to functions
The postprocessing notebook has a fair amount of the functionality exposed in cells. We should abstract most of it away into functions. This will also allow us to add more error checking to make sure the right things have been loaded, to check that the data and segmentation labels match, etc

alex-l-kong:
Closes #148 .

### DIFF
--- a/segmentation/utils/segmentation_utils.py
+++ b/segmentation/utils/segmentation_utils.py
@@ -60,7 +60,7 @@ def compute_complete_expression_matrices(segmentation_labels, base_dir, tiff_dir
         # that are covered by the set of points because the mibitiff function requires the file names
         # and the extensions may be either .tif or .tiff
         if is_mibitiff:
-            all_points = [mt_file for mt_file in os.listdir(tiff_dir) if mt_file.split(".")[1] in ["tif", "tiff"] and 
+            all_points = [mt_file for mt_file in os.listdir(tiff_dir) if mt_file.split(".")[1] in ["tif", "tiff"] and
                           mt_file.startswith(tuple(points))]
 
     # sort the points
--- a/segmentation/utils/segmentation_utils.py
+++ b/segmentation/utils/segmentation_utils.py
@@ -19,9 +19,9 @@ import skimage.io as io
 from segmentation.utils import data_utils, plot_utils, signal_extraction
 
 
-def compute_complete_expression_matrices(segmentation_labels, base_dir=None, tiff_dir=None,
-                                         img_sub_folder="TIFs", is_mibitiff=False, mibitiff_suffix=None,
-                                         points=None, batch_size=5):
+def compute_complete_expression_matrices(segmentation_labels, base_dir=os.path.join("..", "data", "example_dataset"),
+                                         tiff_dir=os.path.join("..", "data", "example_dataset", "single_channel_inputs"),
+                                         img_sub_folder="TIFs", is_mibitiff=False, points=None, batch_size=5):
     """
     This function takes the segmented data and computes the expression matrices batch-wise
     while also validating inputs
@@ -44,28 +44,12 @@ def compute_complete_expression_matrices(segmentation_labels, base_dir=None, tif
         combined_transformed_data (pandas): a DataFrame containing the data with the arcsinh transformation
     """
 
-    if not base_dir:
-        raise ValueError("base_dir not specified")
-
-    if not tiff_dir:
-        raise ValueError("tiff_dir not specified")
-
-    if not img_sub_folder:
-        raise ValueError("img_sub_folder not specified")
-
     # if no points are specified, then load all the points
-    if not points:
-        # if MIBItiff, only look at the points with the desired suffix...
-        if is_mibitiff:
-            # ...but of course, check if that suffix has been defined first
-            if not mibitiff_suffix:
-                raise ValueError("No mibitiff_suffix specified: please do so if loading from MIBItiff")
-
-            points = [point for point in points if point.endswith(MIBItiff_suffix)]
-        else:
-            # load channel data
-            all_points = os.listdir(tiff_dir)
-            all_points = [point for point in all_points if os.path.isdir(os.path.join(tiff_dir, point))]
+    # in the case of MIBItiffs, we can just assume that all the files in tiff_dir
+    # have the correct suffix
+    if points is None:
+        all_points = os.listdir(tiff_dir)
+        all_points = [point for point in all_points if os.path.isdir(os.path.join(tiff_dir, point))]
         points = all_points
 
     # sort the points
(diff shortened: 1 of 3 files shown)


segmentation/utils/segmentation_utils.py
```diff
@@ -16,7 +16,110 @@
...
+        raise ValueError("base_dir not specified")
+
+    if not tiff_dir:
+        raise ValueError("tiff_dir not specified")
+
+    if not img_sub_folder:
+        raise ValueError("img_sub_folder not specified")
+
+    # if no points are specified, then load all the points
+    if not points:
```
ngreenwald, line 57: I like `if points is None`, I think it's clearer to understand

segmentation/utils/segmentation_utils.py
```diff
@@ -16,7 +16,110 @@
...
+        # if MIBItiff, only look at the points with the desired suffix...
+        if is_mibitiff:
+            # ...but of course, check if that suffix has been defined first
+            if not mibitiff_suffix:
+                raise ValueError("No mibitiff_suffix specified: please do so if loading from MIBItiff")
+
+            points = [point for point in points if point.endswith(MIBItiff_suffix)]
```
ngreenwald, line 64: Points hasn't been defined yet, I think this will fail. We should add a test for this

segmentation/utils/segmentation_utils.py
```diff
@@ -16,7 +16,110 @@
...
+
+    # iterate over all the batches
+    for i in range(num_batch):
+        # extract only the points we need for the current batch
+        current_points = points[i * batch_size:(i + 1) * batch_size]
+
+        # assert that all the current_points specified appear as fovs in segmentation_labels
+        # we need this check because otherwise load_imgs_from_tree will fail when we try to
+        # validate the paths because they would not exist to start off with
+        if not np.all(np.in1d(np.array(current_points), segmentation_labels['fovs'].values)):
```
ngreenwald, line 90: Rather than doing this inside the for loop, we should do this at the beginning. 

We should also print out the specific values that are missing

segmentation/utils/segmentation_utils.py
```diff
@@ -16,7 +16,110 @@
...
+            raise ValueError("Invalid point values specified: point and fov values do not match up")
+
+        # and extract the image data corresponding to each of those points
+        image_data = data_utils.load_imgs_from_tree(data_dir=tiff_dir, img_sub_folder=img_sub_folder, fovs=current_points)
```
ngreenwald, line 94: This will need to be conditional on mibi_tiff

segmentation/utils/segmentation_utils_test.py
```diff
@@ -62,8 +62,48 @@ def _create_test_extraction_data():
     return cell_mask, channel_data
 
 
+def test_compute_complete_expression_matrices():
```
ngreenwald, line 65: We can add some simple tests for just running the function with different values to make sure the various conditionals are working as intended

segmentation/utils/segmentation_utils.py
```diff
@@ -16,7 +16,110 @@
...
+        segmentation_labels (xarray): an xarray with the segmented data
+        batch_size (int): how large we want each of the batches of points to be when computing,
+            adjust as necessary for speed and memory considerations
+
+    Returns:
+        combined_normalized_data (pandas): a DataFrame containing the data with the size_norm transformation
+        combined_transformed_data (pandas): a DataFrame containing the data with the arcsinh transformation
+    """
+
+    if not base_dir:
```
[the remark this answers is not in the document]
alex-l-kong: OK, I can hardcode a default.

segmentation/utils/segmentation_utils.py
```diff
@@ -16,7 +16,110 @@
...
+
+    if not img_sub_folder:
+        raise ValueError("img_sub_folder not specified")
+
+    # if no points are specified, then load all the points
+    if not points:
+        # if MIBItiff, only look at the points with the desired suffix...
+        if is_mibitiff:
+            # ...but of course, check if that suffix has been defined first
+            if not mibitiff_suffix:
```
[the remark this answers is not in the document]
alex-l-kong: 👍

segmentation/utils/segmentation_utils.py
alex-l-kong: 👍

segmentation/utils/segmentation_utils.py
alex-l-kong: Since we're removing MIBItiff_suffix, I think we can simply handle both the `is_mibitiff == True` and the `is_mibitiff == False` cases under the same `if points is None` case, so this line of code can be removed if I'm not mistaken.

segmentation/utils/segmentation_utils.py
alex-l-kong: Ah yes, thanks for pointing that out.

segmentation/utils/segmentation_utils.py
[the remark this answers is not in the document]
ngreenwald: No, we don't need any defaults. Don't have there be any default for base_dir.

segmentation/utils/segmentation_utils.py
alex-l-kong: Actually, nvm. The mibitiff and tree structures are different from each other so I had to handle them separately.

alex-l-kong:
I made a few assumptions about the directory and file name structure for MIBItiff files. Let me know if a different format would be better.

segmentation/utils/segmentation_utils.py
```diff
@@ -451,14 +554,14 @@ def generate_expression_matrix(segmentation_labels, image_data, nuclear_counts=F
     """Create a matrix of cells by channels with the total counts of each marker in each cell.
 
     Args:
-        segmentation_labels: xarray of shape [fovs, rows, cols, compartment] containing
+        segmentation_labels (xarray): xarray of shape [fovs, rows, cols, compartment] containing
             segmentation masks for each FOV, potentially across multiple cell compartments
-        image_data: xarray containing all of the channel data across all FOVs
-        nuclear_counts: boolean flag to determine whether nuclear counts are returned
+        image_data (xarray): xarray containing all of the channel data across all FOVs
```
ngreenwald, line 559: My hero, thanks for tackling all of these

segmentation/utils/segmentation_utils.py
```diff
@@ -16,7 +16,110 @@
...
+        # extract only the points we need for the current batch
+        current_points = points[i * batch_size:(i + 1) * batch_size]
+
+        # and extract the image data corresponding to each of those points
+        # make sure we handle the different cases for mibitiff or non-mibitiff
+        if not is_mibitiff:
+            image_data = data_utils.load_imgs_from_tree(data_dir=tiff_dir, img_sub_folder=img_sub_folder, fovs=current_points)
+        else:
+            # only look at the mibitiff files which correspond to the current points we're looking at
+            mibitiff_files = [mt_file for mt_file in os.listdir(tiff_dir) if mt_file.split("_")[0] in current_points]
```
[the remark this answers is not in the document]
alex-l-kong: The reason I did this was that if `is_mibitiff` is set to True,`points` ends up being the list of just the names of points we wish to cover (ex. `["Point1", "Point2"]`) but the `mibitiff_files` argument list has to actually contain the full file names (ex. `["Point1_RowNumber0_...", ...]`).

I could instead simply take the points in `current_points` (a subset of `points`) and regex match all the MIBItiff files that begin with the desired Point prefixes, but I imagine that would also have to use `os.listdir()`.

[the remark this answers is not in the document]
ngreenwald: Got it. I didn't realize how gnarly that filename is. Let's remove all of that metadata, which shouldn't be in the filename anyway, and rename it to just be point8.tiff

templates/Deepcell_Postprocessing.ipynb
```diff
@@ -34,22 +34,32 @@
    "outputs": [],
    "source": [
     "# set up file paths\n",
-    "base_dir = \"../data/example_dataset/\"\n",
+    "base_dir = os.path.join(\"..\", \"data\", \"example_dataset\")\n",
```
ackagel, line 37: any reason this was changed? I feel like the former is easier for lab users to edit.  If it's for os compatibility (mac/windows/linux), I think there's another way to do this s.t posix-like paths are interoperable with dos-like paths (maybe thru pathlib?)

ackagel: s.t = such that 

(math major habit)

alex-l-kong: I can change it back to what it was earlier.

segmentation/utils/segmentation_utils_test.py
ackagel: forgot to mention this in my review, sorry!

Two things fix the coverage loss:
 - adding a `points=None` case for mibitiffs
 - adding more points (and/or lowering batch size) so that the 'batching-for-loop' is entered during testing

segmentation/utils/segmentation_utils.py
```diff
@@ -16,7 +16,110 @@
...
+
+    # if no points are specified, then load all the points
+    # in the case of MIBItiffs, we can just assume that all the files in tiff_dir
+    # have the correct suffix
+    if points is None:
+        # based on looking at the file name format of mibitiff files as well as the directory structure
+        # defined for mibitiff loading, ensure set removal to prevent duplicate points from appearing
+        if is_mibitiff:
+            all_points = os.listdir(tiff_dir)
+            all_points = [point.split("_")[0] for point in all_points if os.path.splittext(point)[1] == ".tiff"]
```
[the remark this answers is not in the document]
alex-l-kong: I think, for the time being, I'll just assume the MIBItiff file name is formatted like Point8.tif or Point8.tiff per Noah's comment. We can handle more nuanced delimiter splitting in another PR if we expect more complicated names.

PR #159, opened 2020-08-10
Dimensionality reduction

wang0324:


segmentation/utils/dimensionality_reduction.py
```diff
@@ -0,0 +1,64 @@
...
+                                       save_dir=None):
+    """Plots the dimensionality reduction of specified population columns
+                Args:
+                    cell_data (pd df): Dataframe containing columns for dimensionality reduction and category
+                    columns (list): List of column names that are included for dimensionality reduction
+                    category (str): Name of column in dataframe containing categorical Population or Patient data
+                    color_map (str): Name of MatPlotLib ColorMap used, default is Spectral
+                    algorithm (str): Name of dimensionality reduction algorithm, default is UMAP
+                    save_dir (str): Directory to save plots, default is None"""
+    cell_data = cell_data.dropna()
```
ngreenwald, line 21: We should add a check here to make sure the algorithm name is one of the three supported algorithms, and otherwise throw an error.

segmentation/utils/dimensionality_reduction.py
```diff
@@ -0,0 +1,64 @@
...
+
+    if algorithm == "UMAP":
+        reducer = umap.UMAP()
+
+        column_data = cell_data[columns].values
+        scaled_column_data = StandardScaler().fit_transform(column_data)
+        embedding = reducer.fit_transform(scaled_column_data)
+
+        plt.scatter(embedding[:, 0], embedding[:, 1], cmap='color_map',
```
ngreenwald, line 30: Looks like UMAP is missing the functionality to color by the category column

segmentation/utils/dimensionality_reduction_test.py
```diff
@@ -0,0 +1,16 @@
...
+import numpy as np
+import string
+import random as rand
+from segmentation.utils import dimensionality_reduction
+import importlib
+
+importlib.reload(dimensionality_reduction)
+
+
+def test_dimensionality_reduction():
```
ngreenwald, line 11: The reason this PR is currently blocked is because the test coverage decreased substantially. In general, the goal of a test function is to actually test out as much of the code as possible. So right now, it only tests with PCA. If there's a bug in the UMAP or tSNE code, we won't find it. 

Instead, we want to test all of the algorithms. What would be useful here is a for loop to test out each algorithm, with and without saving. This will make sure most of the code gets tested

segmentation/utils/dimensionality_reduction.py
```diff
@@ -0,0 +1,74 @@
...
+                Args:
+                    cell_data (pd df): Dataframe containing columns for dimensionality reduction and category
+                    columns (list): List of column names that are included for dimensionality reduction
+                    category (str): Name of column in dataframe containing categorical Population or Patient data
+                    color_map (str): Name of MatPlotLib ColorMap used, default is Spectral
+                    algorithm (str): Name of dimensionality reduction algorithm, default is UMAP
+                    save_dir (str): Directory to save plots, default is None"""
+    cell_data = cell_data.dropna()
+
+    assert(algorithm == "UMAP" or algorithm == "PCA" or algorithm == "tSNE")
```
alex-l-kong, line 23: In this case, I'd raise a ValueError instead. Leave the assertions for the tests.

`if algorithm not in ["UMAP", "PCA", "tSNE"]: raise ValueError("The algorithm specified is not a valid dimensionality reduction technique")`

segmentation/utils/dimensionality_reduction.py
```diff
@@ -0,0 +1,74 @@
...
+        embedding = reducer.fit_transform(scaled_column_data)
+
+        fig1 = plt.figure(1)
+        plt.scatter(embedding[:, 0], embedding[:, 1], cmap=color_map,
+                    c=sns.color_palette(color_map, n_colors=len(cell_data)))
+        plt.gca().set_aspect('equal', 'datalim')
+        plt.title('UMAP projection of data', fontsize=24)
+        fig1.show()
+        if save_dir is not None:
+            plt.savefig(save_dir + "UmapVisualization.png")
```
alex-l-kong, line 39: Recommend looking into `os.path.join` instead of simple string concatenation. I think in this case it would fail if the user specified a non-blank save_dir name because the forward-slash between save_dir and the file name would be missing. You could always just concatenate a `\` as well, but `os.path.join` ensures that (as well as cross-platform support).

segmentation/utils/dimensionality_reduction_test.py
```diff
@@ -0,0 +1,20 @@
...
+    random_cell_data = pd.DataFrame(np.random.random(size=(300, 8)), columns=list('ABCDEFGH'))
+    random_cell_data["cell_type"] = rand.choices(string.ascii_lowercase, k=300)
+
+    test_algorithms = ['PCA', 'tSNE', 'UMAP']
+    for alg in test_algorithms:
+        dimensionality_reduction.visualize_dimensionality_reduction(random_cell_data,
+                                                                    ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H'],
+                                                                    "cell_type",
+                                                                    algorithm=alg, save_dir="")
```
alex-l-kong, line 20: Make sure to include cases where you do save to a directory. If you don't want to have to worry about messing up the current directory structure, would recommend looking into `tempfile.TemporaryDirectory`. You can find examples of how it's used in other test modules.

segmentation/utils/dimensionality_reduction_test.py
```diff
@@ -0,0 +1,26 @@
...
+    random_cell_data = pd.DataFrame(np.random.random(size=(300, 8)), columns=list('ABCDEFGH'))
+    random_cell_data["cell_type"] = rand.choices(string.ascii_lowercase, k=300)
+
+    test_algorithms = ['PCA', 'tSNE', 'UMAP']
+    with tempfile.TemporaryDirectory() as temp_dir:
+        for alg in test_algorithms:
+            if rand.random() > 0.50:
+                dimensionality_reduction.visualize_dimensionality_reduction(random_cell_data,
+                                                                            ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H'],
+                                                                            "cell_type", algorithm=alg, save_dir="")
```
alex-l-kong, line 22: You should assert after here that the specified path to your created file exists. So you should pass `temp_dir` as your argument for `save_dir`, and then use `assert os.path.exists(os.path.join(temp_dir, file_name))`, with `file_name` being whichever file you're saving into depending on which algorithm is running.

PR #160, opened 2020-08-10
Extending Signal Extraction

### LINKED ISSUE #157: Create additional signal_extraction options
Currently, the only signal_extraction option is to sum over all pixels within each cell. To improve upon this, we should start by adding two fairly simple alternative signal extraction functions:

positive_pixels: rather than returning the sum of the image data over pixels, it returns the number of pixels that have non-zero expression of the marker within the cell

center_weighting: Rather than equally weighting each pixel, pixels at the center of the cell are more heavily weighted than pixels at the border. We can aproximate each cell as a circle, then radially multiply pixels by some scaling factor which decreases the further away from the center each pixel gets. To start, lets make this a linear factor.

alex-l-kong:
Addresses and closes #157.

Now that we've implemented a first pass of channel-level synthetic data generation, it's time to look into more detailed signal extraction techniques.

Currently, we're naively summing across all channels. As #157 outlines, we should start by working on more nuanced signal extraction techniques that can lend us more information.

### DIFF
--- a/segmentation/utils/signal_extraction.py
+++ b/segmentation/utils/signal_extraction.py
@@ -1,6 +1,6 @@
 import numpy as np
 
-
+# TODO: work on weighted extraction and implement other discussed techniques of extraction
 def simple_weighted_extraction(cell_coords, image_data):
     """
     Define a weighted extraction technique based on how close the coordinate is to the center
--- a/segmentation/utils/signal_extraction.py
+++ b/segmentation/utils/signal_extraction.py
@@ -1,16 +1,27 @@
 import numpy as np
 
+
 # TODO: work on weighted extraction and implement other discussed techniques of extraction
-def simple_weighted_extraction(cell_coords, image_data):
+def positive_pixels(cell_coords, image_data):
     """
-    Define a weighted extraction technique based on how close the coordinate is to the center
+    Extract channel counts by summing over the number of non-zero pixels in the cell,
+    improves on default_extraction by not distinguishing between pixel expression values
 
     Args:
-        cell_coords: tuples of values representing pixels within one cell
-        image_data: array containing channel counts
+        cell_coords (tuple): tuples of values representing pixels within one cell
+        image_data (numpy): array containing channel counts
 
     Returns:
-        np.array: sum of counts for each channel
+        channel_counts (numpy): sum of counts for each channel
+    """
+
+    pass
+
+
+def center_weighting(cell_coords, image_data):
+    """
+    Extract channel counts by summing over weighted expression values based on distance from center,
+    improves upon default extraction by including a level of certainty/uncertainty
     """
 
     pass
@@ -21,11 +32,11 @@ def default_extraction(cell_coords, image_data):
     Extract channel counts for an individual cell via basic summation for each channel
 
     Args:
-        cell_coords: tuples of values representing pixels within one cell
-        image_data: array containing channel counts
+        cell_coords (tuple): tuples of values representing pixels within one cell
+        image_data (numpy): array containing channel counts
 
     Returns:
-        np.array: sum of counts for each channel
+        channel_counts (numpy): sum of counts for each channel
     """
 
     # transpose coords so they can be used to index an array
--- a/segmentation/utils/signal_extraction.py
+++ b/segmentation/utils/signal_extraction.py
@@ -25,7 +25,7 @@ def positive_pixels_extraction(cell_coords, image_data, threshold=0):
     return channel_counts
 
 
-def center_weighting_extraction(cell_coords, image_data):
+def center_weighting_extraction(cell_coords, image_data, centroid):
     """
     Extract channel counts by summing over weighted expression values based on distance from center,
     improves upon default extraction by including a level of certainty/uncertainty
@@ -33,39 +33,28 @@ def center_weighting_extraction(cell_coords, image_data):
     Args:
         cell_coords (numpy): values representing pixels within one cell
         image_data (xarray): array containing channel counts
-        segmentation_mask (numpy): array containing segmentation labels for each cell
-        cell_label (int): the cell number we wish to extract for each cell, important to distinguish
-            which centroid we want in regionprops because this function only extracts signal
-            for one cell, though we may choose to change this in the future...
+        centroid (tuple): the centroid of the region in question
+
+        Note: cell_coords and centroid are computed from regionprops prior to calling the function
 
     Returns:
         channel_counts (numpy): sum of counts for each channel
     """
 
-    # we could use regionprops but it's probably better to use this method (ala A-Kag's)
-    # because it doesn't require you to know which cell_label to index into
-    centroid = np.sum(cell_coords, axis=0) / cell_coords.shape[0]
-    centroid = centroid.astype(np.int16)
+    # compute the distance box-level from the center outward
+    # this method is more space efficient than the alternative bounding box method
+    # even if we only compute that bounding box around the cell
+    # because there will still be irrelevant cells that bounding box covers
+    weights = np.linalg.norm(cell_coords - centroid, ord=np.inf, axis=1)
 
-    # this will ensure we never get zero or negative weighting values in our weight matrix
-    # I'm currently rounding but it's sort of arbitrary, may decide to truncate (aka round down)
-    # in case of decimal centroid values returned by regionprops
-    center_weight = max(round(centroid[0]), round(centroid[1])) + 1
+    # now center the weights around the middle value
+    weights = 1 - (weights / (np.max(weights) + 1))
 
-    # TODO: this is still not a perfect way to do this for even image dimensions
-    # but I think Adam's suggested method might fix this
-    image_row_index, image_col_index = np.ogrid[:image_data.shape[0], :image_data.shape[1]]
-    weight_matrix = center_weight - np.maximum(np.abs(image_row_index - round(centroid[0])), np.abs(image_col_index - round(centroid[1])))
-
-    # now center the weight matrix around value 1
-    weight_matrix = weight_matrix / center_weight
-
-    # index indo image_data to get th channel values we're interested in
+    # now retrieve the channel counts
+    # for now, I'll leave the indexing as is, I think it's a bit simpler to understand
+    # and we can change in the future if necessary
     channel_values = image_data.values[tuple(cell_coords.T)]
-    weight_values = np.expand_dims(weight_matrix[tuple(cell_coords.T)], axis=0)
-
-    # multiply channel_values by weight_values and then sum across channels
-    channel_counts = weight_values.dot(channel_values)
+    channel_counts = weights.dot(channel_values)
 
     return channel_counts
 
--- a/segmentation/utils/signal_extraction.py
+++ b/segmentation/utils/signal_extraction.py
@@ -1,8 +1,8 @@
 import numpy as np
-
+from skimage.measure import regionprops
 
 # TODO: work on weighted extraction and implement other discussed techniques of extraction
-def positive_pixels_extraction(cell_coords, image_data):
+def positive_pixels_extraction(cell_coords, image_data, threshold=0):
     """
     Extract channel counts by summing over the number of non-zero pixels in the cell,
     improves on default_extraction by not distinguishing between pixel expression values
@@ -15,12 +15,11 @@ def positive_pixels_extraction(cell_coords, image_data):
         channel_counts (numpy): sum of counts for each channel
     """
 
-    # transpose coords so they can be used to index an array
-    cell_coords = cell_coords.T
-    channel_values = image_data.values[tuple(cell_coords)]
+    # index indo image_data to get th channel values we're interested in
+    channel_values = image_data.values[tuple(cell_coords.T)]
 
-    # sum up based on a binary mask that is 1 if the expression value > 0 else 0
-    channel_counts = np.sum(channel_values > 0, axis=0)
+    # sum up based on a binary mask that is 1 if the expression value > threshold else 0
+    channel_counts = np.sum(channel_values > threshold, axis=0)
 
     return channel_counts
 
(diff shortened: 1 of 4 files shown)


segmentation/utils/signal_extraction.py
```diff
@@ -1,31 +1,95 @@
...
+        channel_counts (numpy): sum of counts for each channel
     """
 
-    pass
+    # transpose coords so they can be used to index an array
+    cell_coords = cell_coords.T
+    channel_values = image_data.values[tuple(cell_coords)]
+
+    # sum up based on a binary mask that is 1 if the expression value > 0 else 0
+    channel_counts = np.sum(channel_values > 0, axis=0)
```
ngreenwald, line 23: Should have mentioned this before, lets parameterize this with a threshold value that defaults to 0

segmentation/utils/signal_extraction.py
```diff
@@ -1,31 +1,95 @@
...
+
+    return channel_counts
+
+
+def center_weighting_extraction(cell_coords, image_data):
```
ngreenwald, line 28: Currently this doesn't work as intended. Take a look at the function that is calling signal_extraction: https://github.com/angelolab/segmentation/blob/af92d9cd0dbc299279005c059f5a183222183b7f/segmentation/utils/segmentation_utils.py#L441

It first identifies the coordinates of all pixels belonging to the current cell, and then passes those coordinates, along with the image data of the entire FOV, to this helper function. 

Picking the center of this image won't work. Instead, we need to compute the center of the cell. This isn't possible to do just from the list of cell_coords. Instead, you'll need to use regionprops_table to generate a list of cell_centers, and use that for this seed. 
My guess is that for this function, we'll need to pass in the segmentation_label mask corresponding just to that cell, rather than a list of coordinates.

segmentation/utils/signal_extraction.py
```diff
@@ -1,31 +1,95 @@
...
+
+    # this will ensure we never get zero or negative weighting values in our weight matrix
+    center_weight = center + 1
+
+    # now build the weight matrix using ogrid
+    image_row_index, image_col_index = np.ogrid[:image_data.shape[0], :image_data.shape[1]]
+    weight_matrix = center_weight - np.maximum(np.abs(image_row_index - center), np.abs(image_col_index - center))
+
+    # TODO: the center of even-length arrays is not a single pixel which currently causes
+    # minor signal irregularities at different layers
```
ngreenwald, line 54: Once we have a segmentation mask of the pixels belonging to each cell, we can modify this to use the center of the actual cell, and weight only the pixels belonging to the cell.

segmentation/utils/signal_extraction_test.py
```diff
@@ -4,17 +4,54 @@
 from segmentation.utils import signal_extraction
 
 
+def test_positive_pixels_extraction():
+    # this function tests the functionality of positive pixels extraction
+    # where we count the number of non-zero values for each channel
+
+    # create a series of constant arrays
```
ngreenwald, line 11: Lets use the new synthetic testing functions here instead. 

Once you've added the option for a threshold, we can ensure that with a threshold of 0 all pixels are returned, and then use a different threshold that only applies to the signal from one channel and not the other

segmentation/utils/signal_extraction_test.py
```diff
@@ -4,17 +4,54 @@
...
+                                                                  image_data=combined_channels)
+
+    assert np.all(channel_counts == [0, 4, 4])
+
+
+def test_center_weighting_extraction():
+    # this function tests the functionality of center weighting extraction
+    # where we add a weighting scheme with more confidence toward the center
+    # before summing across each channel
+
```
ngreenwald, line 29: Let's use the new synthetic data here. You can modify the function to have the signal extend 1 pixel beyond the boundaries. Then show that by using this weighting, there are reduced counts of the wrong marker in the wrong cell

segmentation/utils/signal_extraction.py
alex-l-kong: Ok got it, something seemed off about indexing image_data but for some reason couldn't wrap my head around it haha.

segmentation/utils/signal_extraction.py
```diff
@@ -1,31 +1,95 @@
...
+    weight_matrix = weight_matrix / center_weight
+
+    # transpose coords so they can be used to index an array
+    cell_coords = cell_coords.T
+    channel_values = image_data.values[tuple(cell_coords)]
+    weight_values = np.expand_dims(weight_matrix[tuple(cell_coords)], axis=0)
+
+    # multiply channel_values by weight_values and then sum across channels
+    channel_counts = np.sum(np.multiply(channel_values, weight_values.T), axis=0)
+
```
ackagel, line 79: Another way of doing this:
```
# im assuming cell_coords is a num_cells-by-2 array
num_cell_coords = cell_coords.shape[0]

# compute centroid
centroid = np.sum(cell_coords, axis=0)/num_cell_coords

# compute distances (ord=np.inf for box norm, ord=2 for regular euclid time (default))
distances = np.linalg.norm(cell_coords - centroid, axis=1)

# idk how exactly we want to scale the weighting but this is the current version
weights = 1 - (distances/(np.max(distances) + 1))

channel_values = image_data.values[tuple(cell_coords.T)]

# these should be already lined-up correctly
channel_counts = weights.dot(channel_values)
```
might be a few indexing+typos here but conceptually should work

segmentation/utils/signal_extraction_test.py
alex-l-kong: So looking back I realize it wasn't entirely clear as to how we should use a threshold that only applies to signal from one channel. Currently, synthetic_spatial_datagen labels signal with constant values regardless of which channel is used (currently set to 10). So using a set threshold for every channel means that we either get all or nothing from both membrane and nuclear.

Did you by any chance mean that the user should be allowed to specify a different threshold for each channel (ex. 5 for nuclear, and 8 for membrane)? Or would ideally the channel-generation be more diverse in labeling signal (ex. different signal values for nuclear and membrane)?

ngreenwald: The latter. so we can add a parameter to the membrane and nuclear synthetic data functions to specify what the constant value is that gets used, and set a different one for each to test this specific part. 

Unless you have a different idea for a simpler way to test it, in which case go ahead and do that instead! Whatever you think is best

segmentation/utils/signal_extraction_test.py
alex-l-kong: Currently, the test that will be pushed is pretty clunky in that I manually force the extension of signal 1 px beyond the nucleus and membrane depending on which cells and representations are being looked at (cell 1 is currently nuclear, cell 2 is currently membrane) in the test function. I think it would be good (and looking back, what you might have actually been getting at) to have a parameter in the synthetic channel generating functions (ex. uncertainty_thickness) that will allow the user to generate signal beyond the desired nuclear or membrane boundaries.

ngreenwald:
Yeah, that sounds like a good idea!

segmentation/utils/signal_extraction_test.py
alex-l-kong: Also, looking back, I realize that it wasn't entirely clear if and how to test weighted_extraction at specific cell regions (for example, at the edge of cell 1 where membrane signal from cell 2 can extend into cell 1's region). This is because my assumption was that each signal_extraction function should only return the sum of the weighted signal across each channel. I originally tested that weighted_extraction sums were within an acceptable range, but that obviously fails to take into account the signal values at specific cell regions (ex. the boundary, where a cell's signal could "bleed" into other cells). Currently, I only test that center_weighted_extraction returns the expected sum values (which are hard-coded).

The easy fix off the top of my head is to return the weighted channel signal array (from which the sums are computed) so we can index into each cell (and its surroundings) as well. However, I personally think that adds to much clutter to the return value of this function.

Probably overthinking this again, but would like to clarify.

ngreenwald: For sure. I don't think we need to check that the value exactly matches anything in specific regions. We just want to show that the problem is reduced. 

If there was no downweighting, and membrane signal from cell 2 was bleeding into cell 1, then all of the signal within cell 1 would get counted.

With downweighting, the amount of extracted signal should be less than the total counts of membrane signal in the wrong cell. So you can check that the counts using this function are better than using the naive extraction.

alex-l-kong:
Still a couple things that would need to be addressed which I put in TODOs (previous @ngreenwald and @ackagel comments). But overall, the code should work fine and should have flushed the most significant issues with the previous review out.

segmentation/utils/signal_extraction.py
```diff
@@ -1,36 +1,89 @@
 import numpy as np
+from skimage.measure import regionprops
 
 
-def simple_weighted_extraction(cell_coords, image_data):
+# TODO: work on weighted extraction and implement other discussed techniques of extraction
```
ngreenwald, line 5: We can remove this TODO since we have open issues for it

segmentation/utils/signal_extraction.py
```diff
@@ -1,36 +1,89 @@
...
     """
-    Define a weighted extraction technique based on how close the coordinate is to the center
+    Extract channel counts by summing over the number of non-zero pixels in the cell,
+    improves on default_extraction by not distinguishing between pixel expression values
 
     Args:
-        cell_coords: tuples of values representing pixels within one cell
-        image_data: array containing channel counts
+        cell_coords (numpy): values representing pixels within one cell
+        image_data (xarray): array containing channel counts
```
ngreenwald, line 13: add threshold to docstring

segmentation/utils/signal_extraction.py
```diff
@@ -1,36 +1,89 @@
...
+            which centroid we want in regionprops because this function only extracts signal
+            for one cell, though we may choose to change this in the future...
+
+    Returns:
+        channel_counts (numpy): sum of counts for each channel
+    """
+
+    # we could use regionprops but it's probably better to use this method (ala A-Kag's)
+    # because it doesn't require you to know which cell_label to index into
+    centroid = np.sum(cell_coords, axis=0) / cell_coords.shape[0]
```
ngreenwald, line 47: Since we're already calculating regionprops for every cell in the image, we can just add centroid as one of the things to be computed and then pass the centroid coordinates into this function

segmentation/utils/signal_extraction.py
```diff
@@ -1,36 +1,89 @@
...
+    centroid = centroid.astype(np.int16)
+
+    # this will ensure we never get zero or negative weighting values in our weight matrix
+    # I'm currently rounding but it's sort of arbitrary, may decide to truncate (aka round down)
+    # in case of decimal centroid values returned by regionprops
+    center_weight = max(round(centroid[0]), round(centroid[1])) + 1
+
+    # TODO: this is still not a perfect way to do this for even image dimensions
+    # but I think Adam's suggested method might fix this
+    image_row_index, image_col_index = np.ogrid[:image_data.shape[0], :image_data.shape[1]]
```
ngreenwald, line 57: This is a bit inefficient. The image_data array is going to typically be an array of 1024x1024. Most cells are going to around 30x30 pixels. We will likely have ~500 cells per image.

I'm sure ogrid is pretty fast, but it seems like there could be a better way to do this? I'm guessing regionprops will help here. For example, you can get the coordinates of the bounding box that encloses the cell.

segmentation/utils/signal_extraction_test.py
```diff
@@ -2,21 +2,152 @@
...
+    channel_counts_2 = signal_extraction.center_weighting_extraction(cell_coords=coords_2,
+                                                                     image_data=xr.DataArray(sample_channel_data))
+
+    # the reason we get weird values for channel counts 1 is because cell 2 has membrane-level expression
+    # with intentional uncertainty added to it, meaning some of its signal "bleeds" into cell 1's signal
+
+    # TODO: test if this function actually works as intended around the boundaries, aka
+    # the effect of wrong signal being added to a cell is reduced with a center weighting technique
+    assert np.all(channel_counts_1.astype(np.int16) == [448, 88])
```
ngreenwald, line 112: Like I mentioned in my comment above, I think the better way to test this is just to show that the amount of bleedover is less than in the naive extraction function

alex-l-kong: Sounds good.

segmentation/utils/signal_extraction.py
alex-l-kong: regionprops_table could definitely help in reducing the partition of the image np.ogrid looks at. Adam also had a suggestion that I'll look into.

ackagel: adam's suggestion is pretty neat.  but for either one, passing in both `centroid=local_centroid` and `image_data=intensity_image * image` (* is element wise here) can replace centroid and channel_values respectively.  Furthermore, that completely eliminates the need for passing in cell-coords+indexing at the end, since:
```
...

channel_values = image_data
weight_values = np.expand_dims(weight_matrix, axis=0)

channel_counts = weight_values.dot(channel_values)
```
Same with my version, but w/o cell_coords, an extra step is needed to get local coords:
```
...

dist_vecs = np.indices(image_data.shape).resize(2,-1).T - centroid

distances = np.linalg.norm(dist_vecs, axis=1)

weights = 1 - (distances/(np.max(distances) + 1))

channel_counts = weights.dot(image_data)
```

segmentation/utils/signal_extraction.py
```diff
@@ -1,36 +1,89 @@
...
     Returns:
-        np.array: sum of counts for each channel
+        channel_counts (numpy): sum of counts for each channel
     """
 
-    # transpose coords so they can be used to index an array
-    cell_coords = cell_coords.T
-    channel_values = image_data.values[tuple(cell_coords)]
+    # index indo image_data to get th channel values we're interested in
+    channel_values = image_data.values[tuple(cell_coords.T)]
```
ackagel, line 86: passing in `image_data = intensity_image * image` removes this line :o 

btw, intensity_image and image are regionprops, not sure if I mentioned that before

alex-l-kong:
OK this should be good to go for now. I won't touch this branch until after Noah's done reorganizing and renaming 😂.

Issue #161, opened 2020-08-11
compute_closenum speed up


alex-l-kong:
Currently, when you push to GitHub, there's a slight bottleneck in the test_calculate_channel_spatial_enrichment and test_calculate_cluster_spatial_enrichment due to calls to compute_close_cell_num and especially compute_close_cell_num_random. We should look into making these functions more efficient so we can speed up the testing process. It would also allow for higher bootstrap_nums to be used for higher-confidence results.

ackagel:
I wrote a super (x10-x100) speed up for the bootstrapping (`compute_close_cell_num_random`) but I have some questions about the function @ngreenwald @Jaiveers21.

There are two lists of num cells per marker passed in.  The first, looks to be exactly number of cells per each marker.  i.e `mark1_num[0]=num_cells_of_type_0`.  However, the second, appears to be the same but redundantly computed once per iteration of the first. i.e `mark2_num.size == mark1_num.size^2` w/ `mark2_num[k+n*mark1_num.size] == mark1_num[k]` (`k, n < mark1_num.size`)

Is this assessment correct?  If so, is there any reason for this that I'm missing?

ngreenwald:
Where are you seeing that the second is redundantly computed?

ngreenwald:
Wait a minute, let me think about this

ackagel:
I see the redundancy most directly in `compute_close_cell_num`.  Plus, it looks like the size used in the test function is 20 and 400 respectively (the test function only tests for the correct shape, which probably needs to change at some point).

ngreenwald:
Yes, I agree. The reason the code was written this way was because it's an extension of the compute_close_cell_num function. 

We want to get all pairwise interactions between the markers in the list. So have a nested for loop, the first looping over one marker, the second looping over a second marker. In that for loop, we're duplicate work, which is calling get_pos_cell_labels n*2 times instead of n times. We can pre-compute get_pos_cell_labels on each of the n markers in the list. 

Then, we can do our double nested for loop over that list of lists. I'm not sure how much of a speedup this will be, but I think it will make the code easier to read, so either way it's a plus. This will allow us to get rid of marker_1 and marker_2, have a single list of markers, and have the double for loop index into the same list with different indices

compute_close_cell_num_rand was basically an extension of this first function, with a third nested loop, so it has the same baggage. However, here we can't precompute close_cell_num, because that's what we're randomizing over

ackagel:
[...]

Issue #130, opened 2020-07-20
Investigate if calculate_dist_matrix or compute_closenum functions can be simplified


ngreenwald:
Both of these functions in their current implementation have assumptions about the location/ordering of each cell label in the matrix that they produce which is causing problems/requiring use to code around them. 

It would be good to investigate if there are simpler ways to perform these same tasks that don't require so much acrobatics in other areas of the codebase.

ngreenwald:
Closed by #163 and #200

PR #163, opened 2020-08-13
Closenum speedup

ackagel:
Speeds up previously slow functions in spatial analysis, as referenced in issue #161.  Additionally, will look into generalizing 'cell-label acrobatics' as referenced in issue #130.


ackagel:
still want to do docstring revisions + extra pep8 fixes (line lengths) but everything should work correctly

PR #170, opened 2020-08-14
No enrichment randomized and testing patch

alex-l-kong:
The no enrichment cases for channel and cluster enrichment currently rely on a completely random generation of an array, which can cause the p-values and especially the z-scores to fail. For now, adding a random seed for that test will do the trick.

Duplicate tests for no enrichment also appeared, so we'll need to update those to test for p_neg (which was missing).

I still think it would be useful to implement a smarter way to randomly generate no_enrich tests in the near future without having to set a random seed, but for now, this should do the trick.


ackagel:
smh this one's gonna take me a while to get through

Issue #166, opened 2020-08-14
Template vs working versions of jupyter notebooks


ngreenwald:
**Is your feature request related to a problem? Please describe.**
The current workflow for using the jupyter notebooks is as follows:
- Start the docker image
- The bash script checks to see if the the template version of the notebook is different from the version in scripts
- If they are different, it copies the new template into the scripts folder

The goal of this was to check and see if the template has been updated in git, and if so, copy it into scripts. However, anytime people modify the version in scripts, it is now different from the version in templates, so even if no change has been made, it still gives the message about an updated version

**Describe the solution you'd like**
Maybe going back to not having people edit the main version, and asking them to make a copy? This way, if they don't follow the instructions, all that will happen is a new template gets copied over, rather than a git issue? I'm not sure.

ngreenwald:
One idea: maybe have an optional flag (like --update-notebooks). If this is supplied to start_docker.sh, it copies the version in templates to the version if scripts. Otherwise, it does nothing.

We can add a line in the documentation about what to do when updating

ngreenwald:
@ackagel I'm leaning towards the following: If there is no scripts directory (aka first time starting) it copies from templates and creates it. Otherwise, it does nothing. If -update_scripts is passed as a flag `bash start_docker.sh -update`, it copies the templates into scripts, and gives a warning if it's overwriting an existing file, without making an 'updated' version of the file. 

We add a section to the documentation spelling this out. How does that sound? Open to other suggestions as well.

ackagel:
Sounds good.  Should be easy enough to implement (time to put my bash hat back on)

ackagel:
I had a fun idea with the update flag.  It might be possible so that when someone calls `bash start_docker.sh --update` (or `-u`), the Dockerfile is copied+cached somewhere, `git pull origin master` is called, and the cached Dockerfile is compared (w/ diff) to the new one.  If there is a difference, the docker can be rebuilt and then started, and if there isn't, start_docker runs as usual (the cached dockerfile is obvs deleted).

This would eliminate the need to tell people to rebuild the docker when there are changes to the dockerfile and also lets us feel more comfortable w/ changing the dockerfile as needed.  Thoughts?

ngreenwald:
I think the first priority is covering the base scenario, with an update
flag that handles the script copying. But if you want to play around with a
more complex update flag logic, go for it! I think now that the code is
mounted, people will only actually need to rebuild the image when we make a
[...]

PR #181, opened 2020-08-18
Update flag

ackagel:
Adds `--update` flag to `start_docker.sh` to address and close issue #166.  Readme has also been updated to provide user instructions for avoiding script overwrites.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,28 @@ You can shut down the notebooks and close docker by entering control-c in the te
 
 ## Updates
 
-You can update your notebooks using the command:
+You can update the toolbox using the command:
 
 ```
 $ git pull
 ```
+
+After updating, run
+```
+bash start_docker.sh --update
+```
+or
+```
+bash start_docker.sh -u
+```
+
+to overwrite the old notebooks with the new updated versions.
+
+### WARNING
+
+If you have made changes to scripts that you would like to keep, rename them prior to running
+```
+bash start_docker.sh --update
+```
+
+Afterwards, you can simpy bring changes from the old, renamed notebooks into the new notebooks at your discretion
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -53,8 +53,28 @@ You can shut down the notebooks and close docker by entering control-c in the te
...
+bash start_docker.sh -u
+```
+
+to overwrite the old notebooks with the new updated versions.
+
+### WARNING
+
+If you have made changes to scripts that you would like to keep, rename them prior to running
+```
+bash start_docker.sh --update
```
ngreenwald, line 77: I think recopying the code here is confusing and makes them think they need to run it again. I would change this to something like: "prior to running the above command"

ngreenwald:
I think we could do something a bit hacky, like how the jupyter notebooks are currently tested: https://stackoverflow.com/questions/13745648/running-bash-script-from-within-python. 

Basically, we can create a temp_dir with a couple fake .ipynb files, then make sure that with different flags passed in, we get them copied into a tempdir/scripts as appropriate.

ackagel:
the template copying+logic can be moved into a separate script, which we can call from `start_docker.sh` and test through hacky python or BATS.  I think I'll try the python solution for now, since I've grown weary of bash and it's seven thousand different, critically incompatible platform variants (pls don't ask and bless yourself with the gift of bash-ignorance)

ngreenwald:
That's a good point, this is the only thing that runs outside of the docker. What happens if someone has a windows setup?

ackagel:
Technically, we only have instructions for bash users.  I think the assumption is that windows users are using git bash or the linux subsystem.  I'm not sure what flavor of bash those are...

Regardless, I'm pretty sure that the bash I've written is strictly gnu-bash (compatible with all variants) but I could double check (the main place where it might not be is `diff` which might have different exit codes and flags depending on the platform, but I'm p sure I handled the possible cases).

ngreenwald:
Okay, in that case let's wait for someone to complain, and for now we'll
assume it's good.

PR #192, opened 2020-08-18
Make cell size normalization and arcsinh transformation clearer

### LINKED ISSUE #120: Postprocessed CSV outputs normalized vs transformed

alex-l-kong:
Addresses and closes #120.

The current variable names are not that descriptive when it comes to cell size normalization and arcsinh transformation. We've gotten several questions from others about what exactly normalization and transformation meant. Renaming the variables and adding some more comments should do the trick.

### DIFF
--- a/templates/Deepcell_Postprocessing.ipynb
+++ b/templates/Deepcell_Postprocessing.ipynb
@@ -149,8 +149,8 @@
    "metadata": {},
    "source": [
     "Returns:\n",
-    "* cell_size_normalized_data: computed by dividing the marker counts in segmentation_labels by their corresponding cell size, however, if the cell size is zero we leave the marker count as is\n",
-    "* arcsinh_transformed_data: the result of passing each value of cell_size_normalized_data * 100 through the arcsinh function, we add the multiplication of 100 to linearly scale the data beforehand"
+    "* cell_size_normalized_data: computed by dividing the marker counts in segmentation_labels by their corresponding cell size.\n",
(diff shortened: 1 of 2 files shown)


templates/Deepcell_Postprocessing.ipynb
```diff
@@ -144,6 +144,15 @@
     "### Afterwards, we can generate expression matrices from the labeling + imaging data"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Returns:\n",
+    "* cell_size_normalized_data: computed by dividing the marker counts in segmentation_labels by their corresponding cell size, however, if the cell size is zero we leave the marker count as is\n",
```
ngreenwald, line 152: This is just an error check thing, we shouldn't have cells of size zero, so I wouldn't include this is the description, makes things confusing

templates/Deepcell_Postprocessing.ipynb
```diff
@@ -144,6 +144,15 @@
...
+    "* arcsinh_transformed_data: the result of passing each value of cell_size_normalized_data * 100 through the arcsinh function, we add the multiplication of 100 to linearly scale the data beforehand"
```
ngreenwald, line 153: I think this is a bit confusing as written. See if you can reword it to be more clear. Breaking it up into two sentences would probably help.

PR #199, opened 2020-08-25
Deepcell task

omerbt:
This branch includes a new method for automating the process of uploading files to DeepCell, downloading and extracting the output .tif image. 

Implementation is based on relevant modules from https://github.com/vanvalenlab/kiosk-client. 

According to the current implementation of the kiosk-client, the DeepCell output will be downloaded to a predetermined folder (/downloads). Ideally, the path should be a parameter.


omerbt:
> Hey Omer, great start! The build is failing because of a formatting issue, you can just add a blank line at the end of the file and you should be set.
> 
> Could you please update the description of the PR with the indicated information?

requirements.txt
```diff
@@ -16,3 +16,12 @@ statsmodels==0.11.1
 tables==3.6.1
 umap-learn==0.4.6
 xarray==0.12.3
+Twisted~=20.3.0
+certifi
+pyOpenSSL
+service_identity
+git+https://github.com/vanvalenlab/kiosk-client@master
```
ngreenwald, line 23: I think the client is pip installable: can you try adding kiosk_client to the requirements.txt file?

It doesn't look like certifi, pyOpenSSL, or service_identify are imported anywhere, is there a reason these are required?

omerbt: I'd be happy to add more methods. 
regarding the requirements - they are required in order to be able to upload files using twisted and for the tls certificate (even tough they are not explicitly called in the method). I'm changing the git dependency to pip installation.
Thanks!

ngreenwald:
Okay great! I opened an issue in the [kiosk-client](https://github.com/vanvalenlab/kiosk-client/issues/65) to see if there's a reason Will set it up this way. If not, I think it would be pretty easy to submit a PR allowing us to specify an output path. 

Once we get an answer on that, we'll know how to move forward. I think we'll want to create a function like `create_deepcell_output()`. This function should handle all of the necessary data manipulation for running deepcell. It should expect a path to a folder of already formatted images, call the helper function you've already written on each of those images, and then format the output of the helper function appropriately. 

Once that's set up, we can update the jupyter notebook to use this new function

ngreenwald:
Okay, so Will said the reason they never added that option was just because they didn't get to it. I think a PR to address that would be great. If you'd like to take that on, let me know, otherwise Alex or Adam can put that on their todo list. 

What that means is that the high level function can expect the files to be named appropriately. This will simplify things further. 

Let me know what stuff you'd like to get started on

omerbt:
Great, Actually I think it could make sense to add two parameters: output_dir, output_save_name, to allow more flexibility. What do you think? I can get started with that PR and then move on to `create_deepcell_output()`

omerbt:
Great, so I can start with that PR and then move on to `create_deepcell_output()`

ngreenwald:
Perfect, sounds great. If you don't have write access to kiosk-client you fork the repo and submit your PR from your branch, or ask Will to give you write access. 

If you run into any issues feel free to leave a comment on the issue I opened in that repo and Will or I can respond.

omerbt:
Great thanks!

requirements.txt
```diff
@@ -16,3 +16,11 @@ statsmodels==0.11.1
 tables==3.6.1
 umap-learn==0.4.6
 xarray==0.12.3
+Twisted~=20.3.0
+certifi
+pyOpenSSL
+service_identity
+kiosk-client
```
ngreenwald, line 23:
```suggestion
git+https://github.com/vanvalenlab/kiosk-client.git@e1f7b19a730b1223dde666f66748a1b25c2dac31
```

ark/utils/deepcell_service_utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # create /zip folders if they do not exist, and make sure they are empty
+    input_zip = os.path.join(deepcell_input_dir, 'zip')
+    output_zip = os.path.join(deepcell_output_dir, 'zip')
+    for directory in [input_zip, output_zip]:
+        Path(directory).mkdir(parents=True, exist_ok=True)
+        [f.unlink() for f in Path(directory).glob("*") if f.is_file()]
+
+    # zip the preprocessed .tif files
+    for i, point in enumerate(points):
```
ngreenwald, line 49: If you give the kiosk-client a single zip file with multiple TIFs in it, it will analyze all of the files inside the zip.
I think this will help simplify the code. We can just create a single zip file in the same folder that has .tiff files, and submit that single zip file to the client. Then we can unzip the output directly into the output directory

requirements.txt
ngreenwald: This will allow us to use the version on your branch until it gets merged in

ngreenwald (changes requested):
Awesome stuff! I think we can make things even simpler; see comments below.

ark/utils/deepcell_service_utils.py
omerbt: Definitely makes it easier, thanks.
updated In the last commit.

ngreenwald:
Okay great. I think the next step is to address Will's comments about test functions. When that PR is merged in, we can move forward on this one. Once you've finished the kiosk-client PR, you can select the request review button at the top of the github page here: 
image.

It's the green recycling sign, that automatically sends me a notification to look at your newest changes. Awesome stuff!

ngreenwald (changes requested):
Hey Omer, great stuff! It's looking quite nice. 

Here's the remaining things that would be useful:
- Add a check in `create_deepcell_output` to make sure all of the provided points exist as files in the input_dir. If not, raise a ValueError
- Display a warning message if .zip files get removed from output directory
- Add testing for create_deepcell_output. I pushed an example template of how to use mocking to test create_deepcell_output. The idea is that we don't want to actually submit images each time we test: instead, we can mock the function to just return whatever we want. You can modify the testing function to be much more thorough. 

It would be good to check the following:
- Check that files are created in the right place with the right name
- If only a subset of point names are supplied, only that subset is passed through. 
- An error is raised if the point names don't exist

omerbt:
Hey, just a few questions:

> Display a warning message if .zip files get removed from output directory
I'm not sure why should we raise that warning, since the .zip files are expected to be removed (unless they will be removed manually). Alternatively, it could be possible to delete the .zip files after the use in method (instead of keeping them, and removing in the next call).

> An error is raised if the point names don't exist
How is that check different from validating that the file <..>/PointX.tif exists? 

Thanks!

ngreenwald:
In general I get nervous when things automatically delete data. Especially for a command that's easy to run. Maybe the important thing is checking if the output files are getting overwritten? I think it's more likely that someone accidentally provides the wrong directory name, and overwrite existing data, than it is that people are going to want to re-run deepcell on the same data. So if you have another idea of how to reduce the chance of that happening, feel free!

> How is that check different from validating that the file <..>/PointX.tif exists?

Yeah, sorry that probably wasn't clear. The first set of bullets are for what the function will do. The second of set of bullets are things that the testing function will make sure is working correctly. So the function will check if PointX.tif exists. And the test function will make sure that the correct error is raised when you pass in bad data. Does that make sense?

omerbt:
Oh I definitely get that concern, I'll change it so that it won't simply delete the files at the beginning.

ark/utils/deepcell_service_utils_test.py
```diff
@@ -0,0 +1,56 @@
...
+import os
+import tempfile
+from zipfile import ZipFile
+import pytest
+
+from ark.utils.deepcell_service_utils import create_deepcell_output
+
+
+def mocked_run_deepcell(input_dir, output_dir, host, job_type):
+    pathlib.Path(os.path.join(output_dir, 'example_point_feature_0.tif')).touch()
```
ngreenwald, line 11: I think you'll either want to remove this file once the zip has been created, or create the zip without needing it in the first place. Right now we won't know if something is goign wrong in the unzip code, since we've already created the unzipped version of the file

requirements.txt
```diff
@@ -14,3 +14,9 @@ statsmodels==0.11.1
 tables==3.6.1
 umap-learn==0.4.6
 xarray==0.12.3
+#Twisted~=20.3.0
```
ngreenwald, line 17: Now that we're pip-installing, I think we can delete these requirements

ark/utils/deepcell_service_utils.py
```diff
@@ -0,0 +1,100 @@
...
+import glob
+from zipfile import ZipFile
+import warnings
+
+
+def create_deepcell_output(deepcell_input_dir, points, deepcell_output_dir,
+                           suffix='_feature_0', host='https://deepcell.org', job_type='multiplex'):
+    """ Handles all of the necessary data manipulation for running deepcell tasks.
+
+        Creates .zip files (to be used as input for DeepCell),
```
ngreenwald, line 13: We're using the google style for docstrings. You can see the readthedocs raised an error because of the formatting. You can take a look at the docstrings for some of the other functions in the repo to see what the format is

ngreenwald (changes requested):
Looks great! Just some minor changes

alex-l-kong:
@omerbt sorry, couldn't help but take a brief look at this branch. If you're having trouble with the `twisted` library, you'll need to add it to the `autodoc_mock_imports` list of `docs/conf.py` for it to build properly on Sphinx/ReadTheDocs.

omerbt:
@alex-l-kong thanks!

alex-l-kong:
@omerbt looks like other ones you need right now are `pyopenssl`, `service_identity`, and `kiosk_client`. If that still fails, I would assume the problem would most likely be with `kiosk_client`. So in that case, just add `sys.modules['kiosk_client'] = mock.Mock()` after the `autodoc_mock_imports` list.

ngreenwald:
What's the need for RTD to have access to all of the random dependencies in the project?

alex-l-kong:
> What's the need for RTD to have access to all of the random dependencies in the project?

It's an annoying part of working with RTD and specifically, building documentation from docstrings. When snaking through the documentation, I think `autodoc`, which builds the documentation, it first looks through the import statements and complains immediately if it can't import it properly (which can only be guaranteed by mocking it up in `autodoc_mock_imports`). Part of the reason I think it does this is to properly handle a documentation arg-list value or return type of, say, `np.ndarray` or `pd.DataFrame`.

So in Omer's case, when it sees a library such as `kiosk_client`, I think what's happening is that it tries to import it first so it can properly build any documentation that _potentially_ includes an argument or return type of `kiosk_client` in it. If it can't import it, it fails before it tries to build _potentially_ bad documentation with _potential_ non-existent references to `kiosk_client`.

alex-l-kong:
@ngreenwald I'm guessing one of your concerns is that developers have to know to add to `docs/conf.py`, which requires them to have had some experience working with Sphinx/RTD in the past.

I've been looking into a way to somehow tell RTD to do a `pip install -r requirements.txt` on their end so they don't have to rely on `autodoc_mock_imports`, but my leads have come up dry for now. It's for another PR, for sure.

ngreenwald:
My main concern is that kiosk-client has a ton of dependencies that we don't actually care about. So mocking it directly sounds like the best approach since there isn't anything in the documentation that relies on us referencing a class or something from there. 

Thanks!

requirements.txt
omerbt: I'm not sure we can delete all these requirements. Twisted is used (for the reactor), and it is not listed in the requirements file of kisok-client. When using twisted, it is necessary to have service_identity as well, and we need pyopenssl to establish the connection (even though the last two modules are not explicitly called).

ark/utils/deepcell_service_utils_test.py
```diff
@@ -0,0 +1,56 @@
...
+        os.makedirs(output_dir)
+        pathlib.Path(os.path.join(input_dir, 'example_point.tif')).touch()
+
+        create_deepcell_output(deepcell_input_dir=input_dir, deepcell_output_dir=output_dir,
+                               points=['example_point'])
+
+        # make sure DeepCell (.zip) output exists
+        assert os.path.exists(os.path.join(output_dir, 'example_output.zip'))
+
+        # DeepCell output .zip file should be extracted
```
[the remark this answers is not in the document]
omerbt: added in the last commit

ark/utils/deepcell_service_utils_test.py
omerbt: oh I missed that! so i'm now removing the file after it is added to the zip

ngreenwald (changes requested):
I talked to Will, it turns out twisted is a requirement of `treq`, which is a requirement for kiosk-client. 

Can you just double check (by commenting out one at a time and rebuilding the environment) that each of the current requirements are actually needed? If so, we can of course add them, but I'd like to avoid adding unnecessary requirements if possible if they're getting installed via the kiosk.

requirements.txt
```diff
@@ -14,3 +14,8 @@ statsmodels==0.11.1
 tables==3.6.1
 umap-learn==0.4.6
 xarray==0.12.3
+Twisted~=20.3.0
+pyOpenSSL
+service_identity
```
alex-l-kong, line 19: I wonder if simply including `kiosk-client` will do the trick. Looking at `setup.py` of the `kiosk-client` repo, it looks like it's configured to install `treq` on a `pip install`, meaning that if all goes well, we might be able to remove at least `twisted`, if not also `pyOpenSSL` and `service_identity` (assuming `treq` installs at least `twisted` as well).

ark/utils/deepcell_service_utils.py
```diff
@@ -0,0 +1,93 @@
...
+    zip_files = glob.glob(os.path.join(deepcell_output_dir, '*.zip'))
+    zip_files.sort(key=os.path.getmtime)
+    with ZipFile(zip_files[-1], 'r') as zipObj:
+        zipObj.extractall(deepcell_output_dir)
+        for point in points:
+            if point + suffix + '.tif' not in zipObj.namelist():
+                warnings.warn(f'Deep Cell output file was not found for {point}.')
+
+
+def run_deepcell_task(input_dir, output_dir, host='https://deepcell.org',
```
alex-l-kong, line 65: To clarify, we don't need a test function for this one?

ark/utils/deepcell_service_utils.py
```diff
@@ -0,0 +1,93 @@
...
+                location to save deepcell output (as .zip)
+            host (str):
+                Hostname and port for the kiosk-frontend API server.
+                Default: 'https://deepcell.org'
+            job_type (str):
+                Name of job workflow (multiplex, segmentation, tracking).
+                Default: 'multiplex'
+        """
+
+    # more configuration parameters can be set. https://github.com/vanvalenlab/kiosk-client
```
alex-l-kong, line 83: You may want to put this in the docstring so people can reference it on ReadTheDocs.

alex-l-kong (changes requested):
This looks awesome Omer! I had some brief comments about the documentation, testing, and `requirements.txt`.

ark/utils/deepcell_service_utils.py
ngreenwald: I couldn't think of any testing that would actually be useful. We'd have to mock basically the entire function, and at that point I'm not sure what we're actually testing. Open to suggestions if you think there's functionality worth testing

ark/utils/deepcell_service_utils.py
omerbt: I updated it in the last commit

requirements.txt
omerbt: Yes it does! I only checked the requirements file itself (of kiosk-client), but as you and @ngreenwald said, kiosk-client installs all the necessary modules. thanks

ngreenwald:
Hey Omer, this looks great! I'll wait until Adam takes a look before approving in case he has any additional suggestions. 

If you'd like to keep contributing to the repository, I've created an issue outlining the next steps for a subsequent PR. If you've got other stuff to work on, no worries at all, one of us can tackle it. Just let me know what works for you!

Here's the issue: https://github.com/angelolab/ark-analysis/issues/243

ark/utils/deepcell_service_utils.py
alex-l-kong: Ok, I didn't think there was any additional testing we needed for this since its basically a driver function. Don't think we need to add a test function for this.

alex-l-kong:
Ditto on Noah's comment. It looks good to me, let's see what Adam has to say.

ark/utils/deepcell_service_utils.py
```diff
@@ -0,0 +1,93 @@
...
+                           suffix='_feature_0', host='https://deepcell.org', job_type='multiplex'):
+    """ Handles all of the necessary data manipulation for running deepcell tasks.
+
+        Creates .zip files (to be used as input for DeepCell),
+        calls run_deepcell_task method,
+        and extracts zipped output files to the specified output location
+
+        Args:
+            points (list):
+                List of points in preprocessing pipeline
```
ackagel, line 19: I think an option to set `points=None` so that all `.tif` files in the input directory can be read in would be pretty useful.  We have some tools written to help with this.  If you import `ark.utils.io_utils`, you could use something like:
```
if points is None:
    tifs = io_utils.list_files(deepcell_input_dir, substrs='.tif')
    points = io_utils.extract_delimited_names(tifs, delimiter='.')
```
That should plug directly into the code you've already written, assuming I didn't make any typos, etc.

ackagel (commented):
Looks fantastic!  Just one suggestion for the `points` argument in `create_deepcell_output`.

ark/utils/deepcell_service_utils.py
omerbt: added in the last commit

omerbt:
@ngreenwald thanks! sure, I'll take it.

ark/utils/deepcell_service_utils.py
```diff
@@ -0,0 +1,98 @@
...
+                Suffix for DeepCell output filename. e.g. for pointX, DeepCell output
+                should be <pointX>+suffix.tif. Default: '_feature_0'
+            deepcell_output_dir (str):
+                Location to save DeepCell output (as .tif)
+            host (str):
+                Hostname and port for the kiosk-frontend API server
+                Default: 'https://deepcell.org'
+            job_type (str):
+                Name of job workflow (multiplex, segmentation, tracking)
+                Default: 'multiplex'
```
alex-l-kong, line 35: You should put your args list in the order you specified your parameters in the function (so, `deepcell_input_dir`, `deepcell_output_dir`, `points`, etc.).

alex-l-kong (commented):
Hi Omer, this looks great! I had one comment about your docstring for `create_deepcell_output`.

ark/utils/deepcell_service_utils.py
omerbt: Updated in the last commit

alex-l-kong (approved):
Awesome, I think it's good to go!

ngreenwald (approved):
Great!

ackagel (approved):
Looks great!

PR #200, opened 2020-08-25
Update distance matrix to xarray type

### LINKED ISSUE #191: Distance_matrix assumes sequentially-labeled label map.
**Describe the bug**
The current `calc_dist_mat` constructs a distance matrix between all pairs of cells. The 1st cell will be placed in the 0th, row, and so on, up to the nth cell. This will produce a dist_mat with nxn. 

The functions that index into the dist_mat to get distances between cells, for example compute_closenum, use the value of the label to [index into the appropriate row](https://github.com/angelolab/ark-analysis/blob/72f3f840d325daf9d3228c4bafc5044d10b33bb7/ark/utils/spatial_analysis_utils.py#L147). However, this assumes that the cells are sequentially labeled with no missing values. This is not always the case. 

**Expected behavior**
The benefits of doing it this way is that the distance matrix only has to be constructed once. Then, if someone is doing anlyses on a subset of the cells, for example only immune cells, the same total distance matrix can be used, since the indexing still works. 

One option would be to sequentially relabel all cells prior to analysis to avoid this error. The other option would be to require people to regenerate a new distance matrix if they want to analyze a subset of their cells. 

I know @alex-l-kong ran into some annoyances with constructing a fake dist_matrix because of how it expected cells to be labeled. Is this related? Or switching this wouldn't have impacted anything? Tagging @vacuousplanet just to stand out above the sea of other issues I just added.

alex-l-kong:

Addresses and closes #191. The problem with our current distance matrix implementation is that it's not flexible. In other words, we have to assume that indices correspond to their respective cell labels. This is a limitation of numpy arrays for distance matrices. Using xarrays, we can assign more flexible indices. 

The bulk of the changes involve updating calc_dist_matrix and the indexing into distance matrices elsewhere in the spatial analysis code. We change the return type of calc_dist_matrix to xarray and add an indices argument to allow the passage of a set of indices for each fov which correspond to the cell labels. In this way, we do not have to worry about misordering of a distance matrix nor do we have to spend time to reindex the numpy array.

Indexing into a dist_matrix in xarray format will now use the .loc format.

**Remaining issues**

Specifically, the biggest issue is making sure our indexing is correct. There are many places where we use numpy functions on distance matrices and these might need to be updated for xarray usage/reindexing.


alex-l-kong:
Not liking having to specify a dict of coordinates for calc_dist_matrix to work properly, that'll be the biggest point to look at for the upcoming code review.

ark/utils/spatial_analysis_utils.py
```diff
@@ -9,30 +9,55 @@
...
+        centroid_labels (dict): the labels for each fov which are
+            needed to add correct coordinates to access distance matrix
+            needs to be of the same length and indexed corresponding to label_map.coords['fovs']
+            if None, then default assume that cell labels are in order for each fov
         path (string): path to save file. If None, then will directly return
     Returns:
         dist_matrix (dict): contains a cells x cells matrix with the euclidian
-            distance between centers of corresponding cells for every fov"""
+            distance between centers of corresponding cells for every fov, note that each distance matrix
+            is of type xarray"""
```
ackagel, line 25: This might be a bit painful to sort through w/ the merge conflict 😬

ark/utils/spatial_analysis_utils.py
```diff
@@ -9,30 +9,55 @@
...
+        # generate the distance matrix, then assign centroid_labels as coords
         dist_matrix = cdist(centroids, centroids)
-        dist_mats_list.append(dist_matrix)
+        dist_mat_xarr = xr.DataArray(dist_matrix, coords=[centroid_labels[fovs[i]], centroid_labels[fovs[i]]])
+
+        # append final result to dist_mats_list
+        dist_mats_list.append(dist_mat_xarr)
+
     # Create dictionary to store distance matrices per fov
     dist_matrices = dict(zip([str(i) for i in fovs], dist_mats_list))
```
ackagel, line 59: are the members of `fovs` not strings?  if not, can we just have:
```fovs = list([str(fov) for fov in label_map.coords['fovs'].values])```
on line 35?

File: ark/utils/spatial_analysis_utils.py [hunk 3]
```diff
@@ -9,30 +9,55 @@
...
+        props = skimage.measure.regionprops(label_map.loc[fovs[i], :, :, 'segmentation_label'].values)
+        centroids = np.array([props[j].centroid for j in range(len(props))])
+
+        # generate the distance matrix, then assign centroid_labels as coords
         dist_matrix = cdist(centroids, centroids)
-        dist_mats_list.append(dist_matrix)
+        dist_mat_xarr = xr.DataArray(dist_matrix, coords=[centroid_labels[fovs[i]], centroid_labels[fovs[i]]])
+
+        # append final result to dist_mats_list
+        dist_mats_list.append(dist_mat_xarr)
```
ackagel, line 56: `for i in range(len(fovs))` -> `for fov in fovs`

ark/utils/spatial_analysis_utils.py
```diff
@@ -9,30 +9,55 @@
...
+
+        for fov_val in fovs:
+            fov_arr = label_map.loc[fov_val, :, :, 'segmentation_label'].values
+            fov_arr_labels = np.unique(fov_arr[fov_arr > 1]).tolist()
+            centroid_labels[fov_val] = fov_arr_labels
+
+    for i in range(len(fovs)):
+        # extract region properties of label map, then just get centroids
+        props = skimage.measure.regionprops(label_map.loc[fovs[i], :, :, 'segmentation_label'].values)
+        centroids = np.array([props[j].centroid for j in range(len(props))])
```
ackagel, line 49: `[props[j].centroid for j in range(len(props)))]` -> `[prop.centroid for prop in props]`

ark/utils/spatial_analysis_utils.py
```diff
@@ -9,30 +9,55 @@
...
-        a = [props[j].centroid for j in range(len(props))]
-        centroids = np.array(a)
+
+    # generate centroid labels if None
+    if centroid_labels is None:
+        centroid_labels = {}
+
+        for fov_val in fovs:
+            fov_arr = label_map.loc[fov_val, :, :, 'segmentation_label'].values
+            fov_arr_labels = np.unique(fov_arr[fov_arr > 1]).tolist()
```
ackagel, line 43: is the `fov_arr > 1` specifically handling Erin's test data?  If so, maybe we should just clean and resave Erin's data so that `for n > 0,  n -> n-1`.  Then we can change to `fov_arr > 0`.

ark/utils/spatial_analysis_utils.py
```diff
@@ -9,30 +9,55 @@
...
+            centroid_labels[fov_val] = fov_arr_labels
```
ackagel, line 44: now that I think about it some more, I think we can actually get the correctly indexed labels from props:
```
centroid_labels = np.array([prop.label for prop in props])
```

the comment about filtering out 1's still applies, but we wouldn't have to filter out zeros (since regionprops does this for us).

ark/utils/spatial_analysis_utils.py
```diff
@@ -124,8 +149,12 @@ def compute_close_cell_num(dist_mat, dist_lim, num, analysis_type,
     mark1_num = []
     mark1poslabels = []
 
+    # dist_mat_bin = np.zeros(dist_mat.shape, dtype='int')
     dist_mat_bin = np.zeros(dist_mat.shape, dtype='int')
-    dist_mat_bin[dist_mat < dist_lim] = 1
+    dist_mat_bin[dist_mat.values < dist_lim] = 1
+
+    dist_mat_bin = xr.DataArray(dist_mat_bin,
+                                coords=[dist_mat.dim_0.values, dist_mat.dim_1.values])
```
ackagel, line 157: can't we have `coords = dist_mat.coords`?

ark/utils/spatial_analysis_utils.py
```diff
@@ -142,12 +171,11 @@ def compute_close_cell_num(dist_mat, dist_lim, num, analysis_type,
...
-            close_num[j, k] = np.sum(
-                dist_mat_bin[np.ix_(
-                    np.asarray(mark1poslabels[j] - 1, dtype='int'),
-                    np.asarray(mark1poslabels[k] - 1, dtype='int')
-                )]
-            )
+            close_num[j, k] = np.sum(dist_mat_bin.loc[
+                np.array(mark1poslabels[j]),
+                np.array(mark1poslabels[k])
+            ].values)
```
ackagel, line 177: this is kinda shnice.

one thing tho, do the labels need to be `np.array`s ?

ark/utils/spatial_analysis_utils.py
```diff
@@ -173,15 +201,20 @@ def compute_close_cell_num_random(marker_nums, dist_mat, dist_lim, bootstrap_num
...
         len(marker_nums), len(marker_nums), bootstrap_num), dtype='int')
 
-    dist_bin = np.zeros(dist_mat.shape)
-    dist_bin[dist_mat < dist_lim] = 1
+    # dist_mat_bin = np.zeros(dist_mat.shape, dtype='int')
+    dist_bin = np.zeros(dist_mat.shape, dtype='int')
+    dist_bin[dist_mat.values < dist_lim] = 1
+
+    dist_bin = xr.DataArray(dist_bin,
+                            coords=[dist_mat.dim_0.values, dist_mat.dim_1.values])
```
ackagel, line 209: see earlier comment

ark/utils/spatial_analysis_utils.py
ackagel, line 157: on second thought, can't we reduce this to:
```
dist_mat_bin = xr.DataArray(
    dist_mat.values < dist_lim,
    coords=dist_mat.coords,
    dtype='int'
)
```

sorry for the weird formatting but it looked weirder otherwise 🤷

ark/utils/spatial_analysis_utils.py
```diff
@@ -270,8 +303,14 @@ def compute_neighbor_counts(current_fov_neighborhood_data, dist_matrix, distlim,
     # TODO remove non-cell2cell lines (indices on the distance matrix not corresponding to cell labels)
     #  after our own inputs for functions are created
     # refine distance matrix to only cover cell labels in fov_data
-    cell_dist_mat = np.take(dist_matrix, current_fov_neighborhood_data[cell_label_col] - 1, 0)
-    cell_dist_mat = np.take(cell_dist_mat, current_fov_neighborhood_data[cell_label_col] - 1, 1)
+
+    cell_dist_mat = dist_matrix.loc[
+        np.array(current_fov_neighborhood_data[cell_label_col]),
+        np.array(current_fov_neighborhood_data[cell_label_col])
```
ackagel, line 309: see earlier comment about if these needs to be `np.array`s

ark/utils/synthetic_spatial_datagen.py
```diff
@@ -31,7 +31,7 @@ def generate_test_dist_matrix(num_A=100, num_B=100, num_C=100,
             Should be a specified integer value. Default None.
 
     Returns:
-        dist_mat (numpy): the randomized distance matrix we generate directly from predefined distributions
+        dist_mat (xarray): the randomized distance matrix we generate directly from predefined distributions
```
ackagel, line 34: be sure to change this type to `xarray.DataArray` when addressing the big merge conflicts... 😬

ark/utils/spatial_analysis_utils.py
```diff
@@ -9,30 +9,55 @@
...
-def calc_dist_matrix(label_map, path=None):
+def calc_dist_matrix(label_map, centroid_labels=None, path=None):
     """Generate matrix of distances between center of pairs of cells
 
     Args:
         label_map (np array): array with unique cells given unique pixel labels
+        centroid_labels (dict): the labels for each fov which are
+            needed to add correct coordinates to access distance matrix
+            needs to be of the same length and indexed corresponding to label_map.coords['fovs']
+            if None, then default assume that cell labels are in order for each fov
```
ngreenwald, line 20: I don't think we ever need to provide centroid labels. We haven't created the distance matrix yet. So even if we pass in label_map with non-sequentially labeled cells, that's okay, because we'll record that when we create the xarray of dist_matrices.

ark/utils/spatial_analysis_utils.py
ngreenwald: I agree, this is what I suggested. We can clean Erin's data and get rid of this corner case. This will cause us to drop true cells that are equal to 1

ark/utils/spatial_analysis_utils.py
```diff
@@ -9,30 +9,55 @@
...
     # Extract list of fovs
     fovs = list(label_map.coords['fovs'].values)
-    for i in range(0, label_map.shape[0]):
-        props = skimage.measure.regionprops(label_map.loc[fovs[i], :, :, "segmentation_label"].values)
-        a = [props[j].centroid for j in range(len(props))]
-        centroids = np.array(a)
+
+    # generate centroid labels if None
+    if centroid_labels is None:
+        centroid_labels = {}
```
ngreenwald, line 39: centroid_labels should be created by getting the unique IDs of each cell in each FOV, rather than specifying it beforehand.

ark/utils/spatial_analysis_utils.py
```diff
@@ -124,8 +149,12 @@ def compute_close_cell_num(dist_mat, dist_lim, num, analysis_type,
     mark1_num = []
     mark1poslabels = []
 
+    # dist_mat_bin = np.zeros(dist_mat.shape, dtype='int')
```
ngreenwald, line 152: Duplicate line

ark/utils/spatial_analysis_utils.py
ngreenwald: 👍

ark/utils/spatial_analysis_utils.py
ngreenwald: No, you should be able to index with lists of ints

ark/utils/spatial_analysis_utils_test.py
```diff
@@ -74,6 +75,15 @@ def make_example_data_closenum():
     dist_mat[8, 7] = 50
     dist_mat[7, 8] = 50
 
+    # add some randomization to the ordering
```
ngreenwald, line 78: I don't understand this part

alex-l-kong: I should clarify this comment. Since we're not enforcing a specific ordering on distance matrices, I thought it would be nice to add a little random shuffling to the test distance matrix created to show that with the new dist_mat format, the functionality still passes.

Let me know if you think this is necessary or if we should revert back to the original.

ark/utils/spatial_analysis_utils.py
alex-l-kong: So the problem is that sometimes, the members of fovs are ints. For example, in Erin's data. Can we safely assume that we'll only be getting strings or should we handle int fovs as well?

ark/utils/spatial_analysis_utils.py
alex-l-kong: You, my friend, are a lifesaver.

P.S. it turns out the indexing problems I had in the past were because I made a call to np.unique in my random centroid-based generation of a label map. So it wasn't regionprops but rather np.unique that was sorting everything. Big yikes.

ark/utils/spatial_analysis_utils_test.py
ngreenwald: I was getting confused with np.ix_, assuming that was related to the coords, but now I see they're unrelated. 
Totally agree, we definitely want to check this

ark/utils/spatial_analysis_utils.py
ngreenwald: The FOVs are ints in Erin's data because that's what Jay decided to name them. I think in general we should make sure FOVs are strings to avoid unexpected behavior. You can modify the underlying data to be Point7, Point8 instead of 7 and 
 8

alex-l-kong:
Remaining coverage issues due to lack of plot_utils and segmentation_utils testing.

ark/utils/spatial_analysis_utils.py
```diff
@@ -172,12 +186,11 @@ def compute_close_cell_num(dist_mat, dist_lim, num, analysis_type,
...
         for k, m2n in enumerate(mark1_num[j:], j):
-            close_num[j, k] = np.sum(
-                dist_mat_bin[np.ix_(
-                    np.asarray(mark1poslabels[j] - 1, dtype='int'),
-                    np.asarray(mark1poslabels[k] - 1, dtype='int')
-                )]
-            )
+            close_num[j, k] = np.sum(dist_mat_bin.loc[
+                list(mark1poslabels[j]),
+                list(mark1poslabels[k])
```
ackagel, line 191: is `mark1poslabels[j]` not a list already?

ark/utils/spatial_analysis_utils.py
```diff
@@ -312,8 +334,14 @@ def compute_neighbor_counts(current_fov_neighborhood_data, dist_matrix, distlim,
     # TODO remove non-cell2cell lines (indices of distance matrix do not correspond to cell labels)
     #  after our own inputs for functions are created
     # refine distance matrix to only cover cell labels in fov_data
-    cell_dist_mat = np.take(dist_matrix, current_fov_neighborhood_data[cell_label_col] - 1, 0)
-    cell_dist_mat = np.take(cell_dist_mat, current_fov_neighborhood_data[cell_label_col] - 1, 1)
+
+    cell_dist_mat = dist_matrix.loc[
+        list(current_fov_neighborhood_data[cell_label_col]),
```
ackagel, line 339: see earlier comment

ark/utils/spatial_analysis_utils.py
```diff
@@ -17,27 +17,38 @@ def calc_dist_matrix(label_map, path=None):
         path (str):
             path to save file. If None, then will directly return
     Returns:
-        dict:
-            Dictionary containg cells x cells matrices with the euclidian distance between centers
-            of corresponding cells for every fov
-    """
+
+        dist_matrix (dict): contains a cells x cells matrix with the euclidian
```
ackagel, line 21: `dist_matrix (dict):` -> `dict:`

ark/utils/spatial_analysis_utils.py
```diff
@@ -17,27 +17,38 @@ def calc_dist_matrix(label_map, path=None):
...
+            distance between centers of corresponding cells for every fov,
+            note that each distance matrix is of type xarray"""
+
     # Check that file path exists, if given
+
     if path is not None:
         if not os.path.exists(path):
             raise ValueError("File path not valid")
```
ackagel, line 29: while you're refactoring `calc_dist_matrix`, this should be a `FileNotFoundError`

ark/utils/spatial_analysis_utils.py
alex-l-kong: It's actually a pandas series, which breaks the indexing if you use it plainly.

ackagel: 😮 gotcha.  I feel like `mark1poslabels[j].values` is viable too then.  idk if one is better than the other tho.  code what ur heart desires.

alex-l-kong: I think .values definitely looks better 😊.

ark/utils/spatial_analysis_utils.py
```diff
@@ -17,27 +17,39 @@ def calc_dist_matrix(label_map, path=None):
         path (str):
             path to save file. If None, then will directly return
     Returns:
+
```
ngreenwald, line 20: No space here pls

ark/utils/spatial_analysis_utils.py
```diff
@@ -17,27 +17,39 @@ def calc_dist_matrix(label_map, path=None):
...
         dict:
-            Dictionary containg cells x cells matrices with the euclidian distance between centers
-            of corresponding cells for every fov
-    """
+            Contains a cells x cells matrix with the euclidian
+            distance between centers of corresponding cells for every fov,
+            note that each distance matrix is of type xarray"""
```
ngreenwald, line 24: terminal triple quotes on new line

ark/utils/spatial_analysis_utils.py
```diff
@@ -17,27 +17,39 @@ def calc_dist_matrix(label_map, path=None):
...
+
     # Extract list of fovs
-    fovs = list(label_map.coords['fovs'].values)
-    for i in range(0, label_map.shape[0]):
-        props = skimage.measure.regionprops(
-            label_map.loc[fovs[i], :, :, "segmentation_label"].values
-        )
-        a = [props[j].centroid for j in range(len(props))]
-        centroids = np.array(a)
+    fovs = label_map.coords['fovs'].values
```
ngreenwald, line 35: label map is an xarray, not a numpy array, so we should update the docstring. I also think this is a bad variable name since it's singular

ark/utils/spatial_analysis_utils.py
```diff
@@ -17,27 +17,39 @@ def calc_dist_matrix(label_map, path=None):
...
-        dist_mats_list.append(dist_matrix)
+        dist_mat_xarr = xr.DataArray(dist_matrix, coords=[centroid_labels, centroid_labels])
+
+        # append final result to dist_mats_list
+        dist_mats_list.append(dist_mat_xarr)
+
     # Create dictionary to store distance matrices per fov
-    dist_matrices = dict(zip([str(i) for i in fovs], dist_mats_list))
+    dist_matrices = dict(zip(fovs, dist_mats_list))
+
```
ngreenwald, line 52: Update the comment below, which references old variable name

ark/utils/spatial_analysis_utils.py
```diff
@@ -172,12 +187,11 @@ def compute_close_cell_num(dist_mat, dist_lim, num, analysis_type,
     # iterating k from [j, end] cuts out 1/2 the steps (while symmetric)
     for j, m1n in enumerate(mark1_num):
         for k, m2n in enumerate(mark1_num[j:], j):
-            close_num[j, k] = np.sum(
-                dist_mat_bin[np.ix_(
-                    np.asarray(mark1poslabels[j] - 1, dtype='int'),
-                    np.asarray(mark1poslabels[k] - 1, dtype='int')
-                )]
-            )
+            close_num[j, k] = np.sum(dist_mat_bin.loc[
```
ngreenwald, line 190: I think breaking this up into steps will make code readability clearer. First get the sum, then assign the sum

ark/utils/spatial_analysis_utils.py
```diff
@@ -207,15 +221,24 @@ def compute_close_cell_num_random(marker_nums, dist_mat, dist_lim, bootstrap_num
...
+    dist_mat_bin = xr.DataArray(
+        (dist_mat.values < dist_lim).astype(np.int8),
+        coords=dist_mat.coords
+    )
 
     for j, m1n in enumerate(marker_nums):
         for k, m2n in enumerate(marker_nums[j:], j):
             close_num_rand[j, k, :] = np.sum(
-                np.random.choice(dist_bin.flatten(), (m1n * m2n, bootstrap_num), True),
+                np.random.choice(dist_mat_bin.values.flatten(), (m1n * m2n, bootstrap_num), True),
```
ngreenwald, line 238: Sam thing here, one liners are great but not if they make the code too complex. Specifying the output shape on one line, then using that output shape would be better than defining the output shape in the function (I think)

ark/utils/spatial_analysis_utils_test.py
```diff
@@ -74,6 +75,15 @@ def make_example_data_closenum():
...
+    coords_in_order = np.arange(dist_mat.shape[0])
+    coords_permuted = deepcopy(coords_in_order)
+    np.random.shuffle(coords_permuted)
+
+    # we have to 1-index coords because people will be labeling their cells 1-indexed
+    dist_mat = xr.DataArray(dist_mat[np.ix_(coords_permuted, coords_permuted)],
```
ngreenwald, line 84: I think it's good practice for xarrays that whenever we're creating coords or dims, if it's not directly equal to something else, it happens on its own line
so:
````
coords=[coords_permuated + 1, coords_permuated + 1]
dist_mat = xr.DataArray(values, coords=coords)

alex-l-kong:
> Everything looks good! Is there any reason we need to keep the old version of the expression matrix and segmentation labels in the master branch?

I'll remove those and re-request.

PR #203, opened 2020-08-28
Test utils

### LINKED ISSUE #190: Create test_utils file
We have a fair amount of repeated code in our testing functions. In particular, we're often constructing the same data structures over and over again. 
For example, a tree structure with image files inside of FOVs, a MIBITIFF with associated metadata, an xarray of segmentation masks, etc. 

Some of this code is reused, but a fair amount of it is not. Aggregating these all in one place, and spending a bit more time making them nice and extensible functions, will make our lives easier. 

Some prime targets:
Metadata and _create_img_dir in data_utils_test

ackagel:
## **What is the purpose of this PR?**

Consolidating/refactoring utility functions for unit testing into a `test_utils.py` file.  Closes #190.

So far, randomized and filled tiff file creation is implemented in `test_utils.py`.  This tiff generation writes tiffs to a given directory, and additionally returns an xarray which should match those that are loaded by the data loading functions.  Tiff generation is faster than reading existing tiffs since we can control the image size (10x10 images will almost surely speed up our tests) and don't depend on pre-existing example data for our tests.

**Update 8/29/2020** -- label data generation has been added to `test_utils.py`, as well as standard-structure xarray generation for image and label xr's.

**Update 8/31/2020** -- All functions in `data_utils_test.py` have been reworked to use `test_utils.py` functions

**Update 9/01/2020** -- All functions in primary visualization tools tests, as well as all functions in `marker_quantification_test.py`, use `test_utils.py`.

## **Remaining issues**

Still on the agenda:
 - Look through other test files to find low-hanging fruit
 - Add doc-strings to `test_utils.py`


ackagel:
there are definitely some bugs/redundancies (and I will definitely add some comments) at this stage, but I just wanted to get some feedback on my general approach to the new testing in `data_utils_test` (too complicated, simplistic, etc).

ark/utils/test_utils.py
```diff
@@ -0,0 +1,168 @@
...
+def clear_directory(dirpath):
+    if not os.path.isdir(dirpath):
+        raise FileNotFoundError(f'{dirpath} is not a valid directory')
+        return
+
+    for subpath in os.listdir(dirpath):
+        if os.path.isfile(subpath):
+            os.remove(subpath)
+        if os.path.isdir(subpath):
+            os.rmdir(subpath)
```
alex-l-kong, line 157: Might be easier to use shutil.rmtree(dirpath). That way we can make this a one-liner.

ark/analysis/dimensionality_reduction_test.py
```diff
@@ -13,16 +13,7 @@ def test_dimensionality_reduction():
...
-        'A',
-        'B',
-        'C',
-        'D',
-        'E',
-        'F',
-        'G',
-        'H',
-    ]
+    test_cols = list('ABCDEFGH')
```
alex-l-kong, line 16: Nice

ark/utils/test_utils.py
```diff
@@ -0,0 +1,168 @@
...
+    'filename': '20180703_1234_test', 'description': 'test image',
+    'version': 'alpha',
+}
+
+
+def _gen_tif_data(fov_number, chan_number, img_shape, fills, dtype):
+    if fills is None:
+        tif_data = np.random.randint(0, 100,
+                                     fov_number * img_shape[0] * img_shape[1] * chan_number)
+        tif_data = tif_data.reshape((fov_number, *img_shape, chan_number)).astype(dtype)
```
alex-l-kong, line 27: Any reason we can't use the size parameter for randint here? Seems like an extra line having to reshape it.

ark/utils/test_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+    if not isinstance(fov_names, list):
+        fov_names = [fov_names]
+
+    if not isinstance(channel_names, list):
+        channel_names = [channel_names]
+
+    # verify fills structure -> dict(fov_name : list [num_channels])
+    if fills is not None:
+        if fills.keys() is not fov_names:
```
alex-l-kong, line 117: Not sure if this statement will work as intended. I'm guessing you're trying to see if fills.keys() contains the same elements as fov_names. In which case, I think this check could produce some weird behavior. Possibly you meant ==?

ark/utils/test_utils.py
```diff
@@ -0,0 +1,168 @@
...
+            return None
+        for fill in fills.values():
+            if len(fill) != len(channel_names):
+                raise ValueError(f'Not all fills have length {len(channel_names)}')
+                return None
+
+    if fills is not None and not isinstance(fills, list):
+        fills = [fills]
+
+    if fills is not None and (len(fills) != len(channel_names)):
```
alex-l-kong, line 128: Any reason you put parentheses around len(fills) != len(channel_names)?

ark/utils/data_utils_test.py
```diff
@@ -8,50 +8,35 @@
...
-        fov_path = os.path.join(temp_dir, fov, img_sub_folder)
-        if not os.path.exists(fov_path):
-            os.makedirs(fov_path)
-        for img in imgs:
-            io.imsave(os.path.join(fov_path, img), tif)
+        loaded_xr = data_utils.load_imgs_from_mibitiff(temp_dir,
+                                                       channels=channels,
+                                                       delimiter='_')
 
+        assert data_xr.equals(loaded_xr)
```
ngreenwald, line 32: This is great. Does it check for all dims and all values?

ark/utils/test_utils.py
```diff
@@ -0,0 +1,168 @@
+import os
+import numpy as np
+import xarray as xr
+import skimage.io as io
+
+from mibidata import mibi_image as mi, tiff
+
+# required metadata for mibitiff writing (barf)
+METADATA = {
```
ngreenwald, line 9: Lets change to MIBITIFF_METADATA

ark/utils/test_utils.py
ngreenwald: 👍

ark/utils/test_utils.py
```diff
@@ -0,0 +1,168 @@
...
+    'mibitiff': _create_mibitiff,
+}
+
+
+def create_paired_xarray_fovs(base_dir, fov_names, channel_names, img_shape=(1024, 1024),
+                              mode='tiff', delimiter=None, sub_dir=None, fills=None, dtype="int8"):
+
+    if not os.path.isdir(base_dir):
+        raise FileNotFoundError(f'{base_dir} is not a directory')
+        return None
```
ngreenwald, line 99: Why return None after raising an error?

ark/utils/test_utils.py
ngreenwald: We should just make a new temp_dir instead. We can have mutliple context managers within a test, or create a separate test

ark/utils/test_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+    if not isinstance(fov_names, list):
+        fov_names = [fov_names]
+
+    if not isinstance(channel_names, list):
+        channel_names = [channel_names]
+
+    # verify fills structure -> dict(fov_name : list [num_channels])
+    if fills is not None:
```
ngreenwald, line 116: Which test functions currently require specified values in the input data?

ark/utils/test_utils.py
ackagel: added it since I wasn't if it's more efficient to delete the contents of a temp dir in python, or to just create a new `temp_dir`.  Regardless, I'll go with the latter.

ark/utils/test_utils.py
ackagel: TIL The `randint` `size` parameter actually refers to the result's `shape`, not its `size` lol

ark/utils/test_utils.py
ackagel: ya whoops.

ark/utils/test_utils.py
ackagel: C/C++ habits die hard

ark/utils/data_utils_test.py
ackagel: dims, coords, and values all have to match, at least according to its single sentence in the documentation.  I'll probably do a few tests just to be sure.  Single sentence, no given example docs always make me nervous.

ark/utils/test_utils.py
ackagel: my C/C++ brain told me functions have to return things, whoops

ark/utils/test_utils.py
ackagel: I'm not sure if any do atm, but I figured having distinct 'channel values' would be a good check that things are loaded in the correct order (w.r.t. both fovs and channels).  We weren't doing this before, but I think we probs should in case anyone breaks things.

That being said, I'm open to not doing that, or to doing it differently; I recognize that the fills structure is kinda weird.

ackagel: btw, the reason we weren't doing this earlier was because we didn't have a ground-truth xarray to compare the values to, since we were mostly just checking for correct xarray structure on loaded example images.

ngreenwald: I agree, I think it's a good idea, and most of the deepcell tests I've written do this. Given how flexible the overall function is, it just makes this parameter a bit trickier to implement

ackagel: The other approach I see accomplishing the same thing with less explicit structure needed is having `fills=False` randomize data vals, and `fills=True` mean 'set-order' filling, i.e:

fov1: chan1 is filled with 0's, chan2 is filled with 1's, ..., chanN is filled with (N-1)'s
fov2: chan1 is filled with N's, ...

and then just have any values, K>256, wrap around back into [0, 255].

This is probably better since the user doesn't have to explicitly write the exotic structure of `fills`, but lmk what you think.

ark/utils/data_utils_test.py
ackagel: did a few tests; looks like it checks everything (values, ordering, etc).

ark/utils/test_utils.py
ngreenwald: I agree, I think that's a great solution. Especially for channel data (not label data), I think the need to have super fine-grained control over the specific labels is fairly limited.

ackagel:
Just wanna make sure everything looks good before I change all of `data_utils_test` to use `test_utils`

ark/utils/test_utils.py
```diff
@@ -0,0 +1,143 @@
...
+                           coords=[fov_ids,
+                                   range(img_shape[0]),
+                                   range(img_shape[1]),
+                                   channel_names],
+                           dims=["fovs", "rows", "cols", "channels"])
+
+    return filelocs, data_xr
+
+
+def xrs_eq(a_xr, b_xr, sortdim="fovs"):
```
alex-l-kong, line 131: I think for the sake of clarity, rename something like `check_xarrays_equal`.

ark/utils/data_utils_test.py
```diff
@@ -6,105 +6,62 @@
...
+        fovs = ["Point8_otherinfo", "Point9"]
+        channels = ["HH3", "Membrane", "Other"]
 
-    for fov in fovs:
-        fov_path = os.path.join(temp_dir, fov, img_sub_folder)
-        if not os.path.exists(fov_path):
-            os.makedirs(fov_path)
-        for img in imgs:
-            io.imsave(os.path.join(fov_path, img), tif)
+        filelocs, data_xr = tu.create_paired_xarray_fovs(
```
ngreenwald, line 23: while cute, I don't think tu is that informative, and until the rest of the world starts using our library, probably not a widely recognized abbreviation.

ark/utils/test_utils.py
ngreenwald: or xarrays_are_equal, since this will directly be compared to boolean

ark/utils/data_utils_test.py
saeedseyyedi: second to Noah's comment, I think test_utils would be better

ark/utils/test_utils.py
```diff
@@ -0,0 +1,143 @@
...
+    return b_xr.sortby(sortdim).equals(a_xr.sortby(sortdim))
+
+
+def _create_img_dir(temp_dir, fovs, imgs, img_sub_folder="TIFs", dtype="int8"):
+    tif = np.random.randint(0, 100, 1024 ** 2).reshape((1024, 1024)).astype(dtype)
```
saeedseyyedi, line 136: I think passing argument size=(1024,1024) would be faster than reshaping the vector to matrix

ark/utils/test_utils.py
```diff
@@ -0,0 +1,143 @@
...
+        for j, name in enumerate(img_names):
+            io.imsave(os.path.join(base_dir, name), tif_data[i, :, :, j])
+            filelocs[fov].append(os.path.join(base_dir, fov, sub_dir, name))
+
+    return filelocs, tif_data
+
+
+def _create_multitiff(base_dir, fov_names, channel_names, shape, sub_dir, fills, dtype):
+    tif_data = _gen_tif_data(len(fov_names), len(channel_names), shape, fills, dtype)
+
```
saeedseyyedi, line 59: should we check for sub_dir here similar to _create_tif one?

ark/utils/test_utils.py
ackagel: this is a legacy function which is getting replaced by `_create_tifs`.  It'll get deleted soon.

ark/utils/test_utils.py
ackagel: in `_create_tif`, we are loading single channel tifs, where each tifs represents a channel, and a whole folder represents an fov.  The channel images are typically stored in subdirectories (often called 'IMGS', 'imgs', etc.) so the subdir parameter allows us to find the channel data.

In multi/mibitiffs, each tiff represents a fov, while individual layers within a tiff represent the channels, so there's no need to have a subdir argument.

Also, I'm pretty sure the multitiffs are only being generated as the inputs into deepcell (only mibitiff and single tifs are taken in by preprocessings) so we have almost full control for the 'standard' way these are organized in a file structure.

ark/utils/data_utils_test.py
```diff
@@ -6,303 +6,207 @@
...
-
-        # make sure specified channels loaded
-        assert np.array_equal(test_noext_xr.channels.values.sort(), some_chans.sort())
+        assert test_utils.xarrays_are_equal(
+            data_xr[:, :, :, :2], loaded_xr
+        )
 
         # check mixed extension presence
-        test_someext_xr = \
+        loaded_xr = \
```
ngreenwald, line 160: Nice, super thorough

ark/utils/data_utils_test.py
```diff
@@ -6,303 +6,207 @@
...
 
-        # make sure grouping by file prefix was effective
-        assert np.array_equal(test_loaded_xr.fovs, fovs)
-
-        # make sure dim and coord were named w/ defaults
-        assert np.all(test_loaded_xr.loc["fov1", :, :, "img_data"] >= 0)
-        assert test_loaded_xr.dims[-1] == 'compartments'
+        assert test_utils.xarrays_are_equal(data_xr, loaded_xr)
 
+        # test swap int16 -> float
```
ngreenwald, line 203: This is great, cut out 100 lines from the testing code!

ackagel:
@ngreenwald when you say segmented CSV file w/ specified column names, do you mean a function which creates+writes one to disk or one that just creates a `pandas.DataFrame`?

ngreenwald:
Right, bad phrasing. The pandas array itself, I don't think we need to test I/O for any of these functions so we can assume it's already been read into memory.

ackagel:
cool.  That's what I figured, so I went ahead and made a version at the bottom of `test_utils.py` and added it to the mentioned locations.  Let me know if I should include more columns or go about it in a different way.

ark/utils/data_utils_test.py
```diff
@@ -1,325 +1,224 @@
...
+            assert np.issubdtype(loaded_xr.dtype, np.floating)
 
 
 def test_generate_deepcell_input():
     with tempfile.TemporaryDirectory() as temp_dir:
         fovs = ['fov1', 'fov2']
         chans = ['nuc1', 'nuc2', 'mem1', 'mem2']
 
+        data_xr = test_utils.make_images_xarray(
+            None, fovs, 10, 10, chans, dtype="int16"
```
ngreenwald, line 216: Lets add keyword names here to make this look less mysterious

ark/utils/data_utils_test.py
```diff
@@ -328,14 +227,8 @@ def test_generate_deepcell_input():
...
 
-        # check shape
-        assert fov1.shape == (1024, 1024, 2)
-        assert fov2.shape == (1024, 1024, 2)
-
-        assert np.all(fov1[:, :, 0] == 2)
-        assert np.all(fov1[:, :, 1] == 3)
-        assert np.all(fov2[:, :, 0] == 1)
-        assert np.all(fov2[:, :, 1] == 1)
+        assert np.all(fov1 == data_xr.loc['fov1', :, :, ['nuc2', 'mem2']].values)
```
ngreenwald, line 230: I think using np.array_equal would be more clear here to indicate we're comparing values

ark/utils/data_utils_test.py
```diff
@@ -345,10 +238,10 @@ def test_generate_deepcell_input():
         fov1 = io.imread(fov1path)
         fov2 = io.imread(fov2path)
 
-        assert np.all(fov1[:, :, 0] == 3)
-        assert np.all(fov1[:, :, 1] == 4)
-        assert np.all(fov2[:, :, 0] == 2)
-        assert np.all(fov2[:, :, 1] == 2)
+        assert np.all(fov1[:, :, 0] == data_xr.loc['fov1', :, :, nucs].sum(dim='channels').values)
```
ngreenwald, line 241: Probably split this into two lines: first line sums the channels, second line compares summed values to value in the xarray. and use np.array_equal

ark/utils/data_utils_test.py
```diff
@@ -374,87 +263,77 @@ def test_generate_deepcell_input():
...
-                       dims=["fovs", "rows", "cols", "channels"])
+    fov_ids = [f'Point{i}' for i in range(5)]
+    chan_ids = [f'chan{i}' for i in range(3)]
 
-    xr2 = xr.DataArray(np.random.randint(10, size=(2, 30, 30, 3)),
-                       coords=[["Point4", "Point5"], range(30), range(30),
-                               ["chan1", "chan2", "chan3"]],
-                       dims=["fovs", "rows", "cols", "channels"])
+    base_xr = test_utils.make_images_xarray(
+        None, fov_ids, 30, 30, chan_ids
```
ngreenwald, line 278: Same thing here, use the arg names

ackagel:
still have to make some docstrings for the test utils functions and do a quick peak at other functions to see if they can get 'test_utils-ized' (top 10 horrible compound words)

ark/utils/test_utils.py
```diff
@@ -0,0 +1,256 @@
...
+def make_images_xarray(tif_data, fov_ids, row_size, col_size, channel_names, dtype='int16'):
+    if tif_data is None:
+        tif_data = _gen_tif_data(len(fov_ids), len(channel_names), (row_size, col_size), False,
+                                 dtype)
+    coords = [fov_ids, range(row_size), range(col_size), channel_names]
+    dims = ["fovs", "rows", "cols", "channels"]
+    return xr.DataArray(tif_data, coords=coords, dims=dims)
+
+
+def make_labels_xarray(label_data, fov_ids, row_size, col_size, compartment_names):
```
ngreenwald, line 214: If we're assuming this is always 4D, we don't need to pass in row_size and col_size, we can get that from label_data

ark/segmentation/marker_quantification_test.py
```diff
@@ -1,32 +1,38 @@
...
+    cell_mask = np.expand_dims(cell_mask, axis=0)
+    channel_data = np.expand_dims(channel_data, axis=0)
+
+    segmentation_masks = test_utils.make_labels_xarray(
+        label_data=np.expand_dims(cell_mask, axis=-1),
+        fov_ids=fovs,
+        row_size=40,
+        col_size=40,
+        compartment_names=['whole_cell']
+    )[0]
```
ngreenwald, line 27: This [0] is mysterious. I think for ease of understanding this should be on its own line with a comment indicating that the test function produces 4D data but the function being tested takes 3D data

ark/segmentation/marker_quantification_test.py
```diff
@@ -61,12 +67,13 @@ def test_compute_marker_counts():
     assert np.array_equal(segmentation_output.loc['whole_cell', 1:, 'cell_size'],
                           segmentation_output.loc['whole_cell', 1:, 'area'])
 
-    # test whole_cell and nuclear compartments with same data
```
ngreenwald, line 64: I think keeping this comment, or some similar comment indicating what is going on here, is worthwhile

ark/segmentation/marker_quantification_test.py
```diff
@@ -124,39 +136,50 @@ def test_compute_marker_counts():
...
-
-    normalized, arcsinh = marker_quantification.generate_expression_matrix(segmentation_masks,
-                                                                           channel_data)
+    tif_data = np.zeros((2, 40, 40, 5), dtype="int16")
+    tif_data[0, :, :, :] = channel_data
+    tif_data[1, 5:, 5:, :] = channel_data[:-5, :-5]
+
+    segmentation_masks = test_utils.make_labels_xarray(
+        label_data=cell_masks,
+        fov_ids=fovs,
```
ngreenwald, line 155: We should have fov_ids default to be what you added in line 139 above so that it only needs to be passed in if it's being changed

ark/segmentation/marker_quantification_test.py
```diff
@@ -124,39 +136,50 @@ def test_compute_marker_counts():
...
+        col_size=40,
+        compartment_names=['whole_cell']
+    )
+
+    channel_data = test_utils.make_images_xarray(
+        tif_data=tif_data,
+        fov_ids=fovs,
+        row_size=40,
+        col_size=40,
+        channel_names=chans
```
ngreenwald, line 166: Same thing for channel names, default should be numerically increasing channels

ark/segmentation/marker_quantification_test.py
```diff
@@ -179,14 +202,22 @@ def test_generate_expression_matrix_multiple_compartments():
...
 
-    channel_data = xr.DataArray(channel_datas,
-                                coords=[["Point0", "Point1"], range(40), range(40),
-                                        ["chan0", "chan1", "chan2", "chan3", "chan4"]],
-                                dims=["fovs", "rows", "cols", "channels"])
+    segmentation_masks_unequal = test_utils.make_labels_xarray(
+        label_data=unequal_masks,
+        fov_ids=fovs,
+        row_size=40,
+        col_size=40,
```
ngreenwald, line 210: Same thing here, we should be able to get rid of almost all of these arguments since they're defaults that are constant throughout

ark/segmentation/marker_quantification_test.py
```diff
@@ -290,54 +317,38 @@ def test_compute_complete_expression_matrices():
     with tempfile.TemporaryDirectory() as temp_dir:
         # define 2 FOVs and 2 mibitiff_imgs
         fovs = ["Point1", "Point2", "Point3"]
-        mibitiff_imgs = ["Point1.tif", "Point2.tiff", "Point3.tiff"]
+        channels = ["HH3", "Membrane"]
 
         # define a subset of fovs
         fovs_subset = fovs[:2]
 
         # since example_dataset exists, lets create a new directory called testing_dataset
```
ngreenwald, line 325: I don't think this comment is relevant anymore

ark/segmentation/marker_quantification_test.py
```diff
@@ -1,32 +1,38 @@
...
-    dims = ['rows', 'cols', 'channels']
-    input_images = xr.DataArray(channel_data, coords=coords, dims=dims)
+    fovs = ['fov1']
+    chans = [f'chan{i}' for i in range(5)]
+
+    cell_mask, channel_data = test_utils.create_test_extraction_data()
+    cell_mask = np.expand_dims(cell_mask, axis=0)
+    channel_data = np.expand_dims(channel_data, axis=0)
+
+    segmentation_masks = test_utils.make_labels_xarray(
```
ngreenwald, line 21: We can probably have make_labels_xarray handle the np.expand_dims logic

ark/utils/test_utils.py
```diff
@@ -0,0 +1,256 @@
...
+def _gen_label_data(fov_number, comp_number, img_shape, dtype):
+    label_data = np.zeros((fov_number, *img_shape, comp_number), dtype=dtype)
+
+    right = (img_shape[1] - 1) // 2
+    left = (img_shape[1] + 2) // 2
+    up = (img_shape[0] - 1) // 2
+    down = (img_shape[0] + 2) // 2
+
+    counter = 1
+    for fov in range(fov_number):
```
ngreenwald, line 50: This makes quadrants with different values, right?

ark/utils/test_utils.py
```diff
@@ -0,0 +1,256 @@
...
+        tiffpath = os.path.join(base_dir, f'{fov}.tiff')
+        io.imsave(tiffpath, tif_data[:, :, :, i], plugin='tifffile')
+        filelocs[fov] = tiffpath
+
+    tif_data = np.swapaxes(tif_data, 0, -1)
+
+    return filelocs, tif_data
+
+
+def _create_labels(base_dir, fov_names, comp_names, shape, sub_dir, fills, dtype):
```
ngreenwald, line 130: I understand these _helper_functions won't be directly called by anyone when writing tests, but it could be helpful for us to indicate in the name which ones are creating the data and which are saving it as a way to separate out the similarly named ones

ark/utils/test_utils.py
```diff
@@ -0,0 +1,256 @@
...
+        data_xr = make_labels_xarray(tif_data, fov_ids, *img_shape, channel_names)
+    else:
+        data_xr = make_images_xarray(tif_data, fov_ids, *img_shape, channel_names)
+
+    return filelocs, data_xr
+
+
+def xarrays_are_equal(a_xr, b_xr, sortdims=("fovs", "channels")):
+    try:
+        if len(sortdims) > 1:
```
ngreenwald, line 192: What's the scenario where this comes up? Is it easier to sort here than to ensure that our test data is generated with the same order that the functions will create it in?

ark/utils/test_utils.py
ackagel: I wasn't sure what to do there because of the `label_data=None` case, which produces synthetic label data.  I could maybe just set default values for row_size, col_size and use those if `label_data` is None, and otherwise get them from `label_data`'s shape.

ark/segmentation/marker_quantification_test.py
ackagel: Oops, I think that line was just an innocent bystander 😬

ark/utils/test_utils.py
ackagel: basically yeah.  There's a border of 0's between each quadrant.  Not sure how useful it is, but I figured it was better than all-zeros or random values (which would mess up region props).

ark/utils/test_utils.py
ackagel: Sounds good.  btw, how are we going to handle internal functions in the docs?  They should definitely have doc strings, but it would maybe be useful to deliberately ignore adding functions prefixed with '_' to the docs (maybe a setting in sphinx?)

ark/utils/test_utils.py
ackagel: the order that files are loaded in seems nondeterministic (but *usually* alphabetical), which makes the fovs column out of order; same issue comes up with channels for single channel tifs.  I figured it's better to do all sorting within the function.

Also, the reason `sortdims` is parametrized is because `label_xr`'s have `"components"` as their 4th dimension.  (although now that I think about it, I don't see why it can't just be `sortdims=(0, 3)`).  It will also take a list as input, but whatever it is it just needs to be iterable

ark/utils/test_utils.py
ngreenwald: Won't this function fail with label_data=None?

ark/utils/test_utils.py
ngreenwald: Ooooh, that's great, having something which will work with regionprops is a great idea

ark/utils/test_utils.py
ngreenwald: Yeah, maybe @alex-l-kong can look into that. I'm certain that it's possible to ignore specific files, so worst comes to worst we can do that.

ark/utils/test_utils.py
ngreenwald: I think all of the data_loaders have a for loop that loops over the channels and FOVs list, either the provided one or everything in the folder with io_utils. So along as we've sorted the list that it's looping over it should be deterministic

ark/utils/test_utils.py
ackagel: oh whoops, I was supposed to have that be a thing like in `make_images_xarray`

ark/utils/test_utils.py
ackagel: cool, I'll make sure that those are sorted in the data loaders.  That way, we can just use the regular `.equals` method on xarrays, instead of a custom one.

ark/utils/test_utils.py
ackagel: apparently, sphinx already ignores `_` prefixed functions (see doc build on this PR for reference).

ackagel:
still have some docstrings to finish, but I think I've cleared up most low hanging fruit.

There's definitely more that could be condensed into `test_utils` but that's mostly in everything to-do with spatial analysis (including synth_datagen).  I think our spatial analysis tools/tests needs to under-go some standardization (xarray/df structure, naming conventions, testing goals/conventions, etc) before we consolidate everything into `test_utils`.

ark/segmentation/marker_quantification_test.py
```diff
@@ -1,32 +1,30 @@
...
+        label_data=cell_mask,
+        compartment_names=['whole_cell']
+    )
 
-    # create xarray with channel data
-    coords = [range(40), range(40), ['chan0', 'chan1', 'chan2', 'chan3', 'chan4']]
-    dims = ['rows', 'cols', 'channels']
-    input_images = xr.DataArray(channel_data, coords=coords, dims=dims)
+    input_images = test_utils.make_images_xarray(
```
ngreenwald, line 22: These three lines can go on a diet

ark/segmentation/marker_quantification_test.py
```diff
@@ -124,48 +130,48 @@ def test_compute_marker_counts():
...
-    normalized, arcsinh = marker_quantification.generate_expression_matrix(segmentation_masks,
-                                                                           channel_data)
+    normalized, _ = marker_quantification.generate_expression_matrix(segmentation_masks,
+                                                                     channel_data)
 
     assert normalized.shape[0] == 7
 
-    assert np.all(normalized['chan0'] == np.repeat(1, len(normalized)))
-    assert np.all(normalized['chan1'] == np.repeat(5, len(normalized)))
-    assert np.all(normalized['chan2'] == normalized['chan2'])
```
ngreenwald, line 155: You removed our most crucial test

ark/utils/data_utils_test.py
```diff
@@ -1,325 +1,211 @@
...
                                                      for i in range(3)])
 
-        # make sure all folders loaded
-        assert np.array_equal(test_someext_xr.fovs.values.sort(), fovs.sort())
+        assert loaded_xr.equals(data_xr)
 
-        # makes sure all channels loaded
-        assert np.array_equal(test_someext_xr.channels.values.sort(), chans.sort())
+    with tempfile.TemporaryDirectory() as temp_dir:
+        fovs, chans, imgs = test_utils.gen_fov_chan_names(1, 2, return_imgs=True)
```
ngreenwald, line 158: should add arg names here

ark/utils/data_utils_test.py
```diff
@@ -1,325 +1,211 @@
...
 
 
 def test_load_imgs_from_dir():
     # test loading from 'free' directory
-    with tempfile.TemporaryDirectory(prefix='one_file') as temp_dir:
-        imgs = ["fov1_img1.tiff", "fov2_img2.tiff", "fov3_img3.tiff"]
-        fovs = [img.split("_")[0] for img in imgs]
-        _create_img_dir(temp_dir, fovs=[""], imgs=imgs, img_sub_folder="", dtype="float")
+    with tempfile.TemporaryDirectory() as temp_dir:
+        fovs, _ = test_utils.gen_fov_chan_names(3, 0, use_delimiter=True)
```
ngreenwald, line 178: and here

ngreenwald (changes requested):
Looks awesome, small tweaks only. I wouldn't worry about cleaning up the stuff at the bottom of data_utils, most of those aren't used anymore for anything and can probably be retired.

ngreenwald:
> There's definitely more that could be condensed into `test_utils` but that's mostly in everything to-do with spatial analysis (including synth_datagen). I think our spatial analysis tools/tests needs to under-go some standardization (xarray/df structure, naming conventions, testing goals/conventions, etc) before we consolidate everything into `test_utils`.

Yeah, the synthetic_spatial_datagen functions are more substantial, so I think that will be a separate PR

ark/segmentation/marker_quantification_test.py
ackagel: issue 5049102: consolidate all lingual tests of equality into a `test_equalities.py` file. some examples:
 - 1 == 1
 - 2 == 2
 - n == n (we might be able to generalize this sequence by using 20 different scipy packages and machine learning)
 - `normalized['chan2'] == normalized['chan2']`
 - `a_list == [i for i in a_list if i == i]`
 - `False == True` (have we verified this?)
 - `the_joke == is_going_on_a_bit_long_dont_you_think`

ark/utils/data_utils_test.py
ackagel: just to clarify, add arg names in `gen_fov_chan_names` or in `create_paired_xarray_fovs` ? or both now that I mention it?

ngreenwald: If the argument is a variable with an informative name, then it's okay to not use arg names. But when it's just raw ints we definitely should have them

ngreenwald (approved):
Awesome, looks great

PR #216, opened 2020-09-10
Modifiable regionprops args

### LINKED ISSUE #214: Marker_quantification assumes there will always be at least one background pixel
**Describe the bug**
The marker quantification pipeline removes the background signal from the top of extracted array. However, it doesn't check if the label mask actually had background, it just assumes it did. For small crops full of dense cells, this isn't the case, and so it removes the first cell.

ngreenwald:

Modifies regionprops so that the properties are modifiable (in theory) by the user. This allows different props to be specified if desired. 

Closes #214, which was removing the first cell when background wasn't present in the image

Adds centroid location of each cell as a default output to the pipeline. 

Separated out the names of the regionprops properties being passed to regionprops, and the names of the resulting attributes. This makes it easier to track what we're passing to regionprops, and what we're adding to the final CSV. These aren't the same thing. For example, we get coords for internal signal extraction, but don't pass to the user. Also, some regionprops attributes return multiple columns, which don't have the same name as the property. This will need to be specifically identified for each case.

### DIFF
--- a/ark/segmentation/marker_quantification.py
+++ b/ark/segmentation/marker_quantification.py
@@ -11,7 +11,8 @@ from ark.utils import data_utils, io_utils, segmentation_utils
 from ark.segmentation import signal_extraction
 
 
-def compute_marker_counts(input_images, segmentation_masks, nuclear_counts=False):
+def compute_marker_counts(input_images, segmentation_masks, nuclear_counts=False,
+                          regionprops_features=None):
     """Extract single cell protein expression data from channel TIFs for a single point
 
     Args:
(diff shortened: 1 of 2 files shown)


ark/segmentation/marker_quantification.py
```diff
@@ -21,21 +22,36 @@ def compute_marker_counts(input_images, segmentation_masks, nuclear_counts=False
...
+            morphology features for regionprops to extract for each cell
 
     Returns:
         xarray.DataArray:
             xarray containing segmented data of cells x markers
     """
 
-    unique_cell_ids = np.unique(segmentation_masks[..., 0].values)
+    if regionprops_features is None:
+        regionprops_features = ['label', 'area', 'eccentricity', 'major_axis_length',
```
alex-l-kong, line 34: Would a variable like this be a candidate for settings.py once someone addresses this issue?

alex-l-kong (approved):
Looks good to me.

ark/segmentation/marker_quantification.py
ngreenwald: You see exactly what my thought-process was. Yes exactly!

ackagel: noted

ark/segmentation/marker_quantification.py
```diff
@@ -21,21 +22,36 @@ def compute_marker_counts(input_images, segmentation_masks, nuclear_counts=False
...
+    regionprops_names = copy.copy(regionprops_features)
+    regionprops_names.remove('coords')
 
-    # define morphology properties to be extracted from regionprops
-    object_properties = ["label", "area", "eccentricity", "major_axis_length",
-                         "minor_axis_length", "perimeter", 'coords']
+    # centroid returns two columns, need to modify names
+    if np.isin('centroid', regionprops_names):
+        regionprops_names.remove('centroid')
+        regionprops_names += ['centroid-0', 'centroid-1']
```
ackagel, line 47: I don't see where this split is being used,  Is it being used implicitly in the background?

ackagel (commented):
just a clarification question

ark/segmentation/marker_quantification.py
ngreenwald: If you supply 'centroid' to reginonprops_table, it returns columns with those names for each part of the centroid.

ackagel (approved):

PR #218, opened 2020-09-10
Create boxplot function in visualize

### LINKED ISSUE #183: Code cleanup for analysis/visualize.py
Add optional save_dir for `visualize_z_scores`
Generalize variable names in get_sorted_data from `patient_column` to something more general. This function supports barchart plotting of any categorical data, not just cells by patients and cell type. 
Create a plot_barchart helper function, and call this helper function for each of the three different barcharts sub-calls in here. 

Testing:
Add in tests for visualize_z_scores with above save_path
Add tempdir for test_visualize_cells so that plots are not generated locally

### LINKED ISSUE #204: Default box and whisker visualization function
We don't currently have a nice boxplot function in the repo. Mike suggested this as a next task. 

This function should take in a pandas array and produce a box and whisker plot. The arguments to the function should be the following. 

- The pandas array, required. 
-  The column specifying the numeric value to be plotted, required. 
-  A column used to split the data, optional. For example, a lineage column. If this column is a provided, a separate boxplot will be generated for each unique value in this column. If not proivded, only a single boxplot is generated
- A list of which values within the split column to use, optional. If provided, the variables in this list will each get their own boxplot, and the rest of the variables in the column list will not be plotted.

The plot: The plot should generate a box and whisker plot with 1st and 3rd quartile. Outliers should be plotted as individual dots, jittered so that they don't overlap

alex-l-kong:

Addresses and closes #204. Addresses and closes #183.

Add a function draw_boxplot to visualize.py which will allow the user to draw (and customize) a boxplot. In the meantime, clean up visualize.py and visualize_test.py.

**Remaining issues**

The previous way save_dir was being used was more of like a save_prefix. I changed the nature of save_dir and added a little more error checking to make sure we weren't trying to save to a save_dir that didn't exist.


ark/analysis/visualize.py
```diff
@@ -1,9 +1,65 @@
...
+
+    if col_split:
+        # if col_split, then we explicitly facet the visualization
+        # labels are automatically generated in Seaborn
+        sns.boxplot(x=col_split, y=col_name, data=cell_data)
+        plt.title("Distribution of %s, faceted by %s" % (col_name, col_split))
+    else:
+        # otherwise, we don't facet anything, but we have to explicitly make vertical
+        sns.boxplot(x=col_name, data=cell_data, orient="v")
+        plt.title("Distribution of %s" % col_name)
```
ackagel, line 53: should `cell_data` be `data_to_viz` here?

ngreenwald (changes requested):
Looks great. Can you add it in as a cell to the visualization notebook so we can see what it looks like?

ngreenwald:
Also looks like you took care of some of #183, can you see if there's anything else?

ark/analysis/visualize.py
```diff
@@ -1,9 +1,65 @@
...
+
+    # save visualization to a directory if specified
+    if save_dir is not None:
+        if not os.path.exists(save_dir):
+            raise ValueError("save_dir %s does not exist" % save_dir)
+
+        plt.savefig(os.path.join(save_dir, "sample_boxplot_viz.png"))
```
ackagel, line 60: @ngreenwald in general for the visualization plots, if `save_dir` doesn't exist, should we create that directory by default, provide an option to do so, or just throw an error as above?  

I know it's easy to externally create a folder, but that mandates that users do some tinkering around.

ark/analysis/visualize_test.py
```diff
@@ -1,10 +1,54 @@
...
+import pytest
 
 from ark.analysis import visualize
 from ark.utils import test_utils
 
 
+def test_draw_boxplot():
+    # trim random data so we don't have to visualize as many facets
+    random_data = test_utils.make_segmented_csv(100)
+    random_data = random_data[random_data['PatientID'].isin(np.arange(1, 5))]
```
ackagel, line 12: we could make this an option in the test_utils if you think cases like this will become more common

ackagel (commented):
looks super duper; just a few comments

ark/analysis/visualize.py
alex-l-kong: Looking at `example_visualization.ipynb`, the run will actually fail if you don't have a visualization folder specified in data/example_dataset. We can do one of two things:

- Create the directory if it does not exist
- Push an example visualization folder to the repo (not a huge fan of this option, but if it makes it easier to do then I wouldn't mind)

ngreenwald: I think this function is low level enough that it shouldn't create its own directory. whatever is calling this should handle the folder creation. For example the first cell in the notebook

alex-l-kong: @ngreenwald OK, then would it be best to add a command in the notebook to create `data/example_dataset/visualization` if it doesn't exist?

ngreenwald:
Yeah, and then all of the plots can get saved there

ark/analysis/visualize_test.py
alex-l-kong: I only did it because I didn't want the boxplot to display like 10 different categories on one chart. I don't imagine we'll be doing this too often unless we plan to add a lot more faceted graphing functions in `visualize.py`, for example.

ngreenwald (changes requested):
Seaborn defaults look quite nice. However, you picked Na to plot, aka Sodium, which is not actually a protein target: it's a naturally abundant metal. AKA we don't really care at all about sodium expression. Try picking CD20, CD45, PD-1, or one of the other biological targets.

alex-l-kong:
> Seaborn defaults look quite nice. However, you picked Na to plot, aka Sodium, which is not actually a protein target: it's a naturally abundant metal. AKA we don't really care at all about sodium expression. Try picking CD20, CD45, PD-1, or one of the other biological targets.

NaBrO (sorry, I couldn't resist). But yeah, I'll fix that up.

ngreenwald (changes requested):
Looks great. Small tweak to function description

ngreenwald (approved):

ackagel (approved):
looks gucci

ngreenwald:
woops sorry, forgot to click go. You guys can merge stuff in once it's approved

PR #222, opened 2020-09-15
Binning nuclear labels

ngreenwald:

Currently, when we match nuclei to cells, we don't require that the nucleus be smaller than the cell. If the nuclear segmentation has errors, this can result in non-biological readouts. 

This PR adds the option to separate out nuclei that are larger than a cell, and relabel the portion that falls within a cell, to remove this issue

Added a function that takes a nuclear mask and a cell mask, and relabels any nuclei that have pixels outside of the boundaries of the corresponding cell they map to. 

**Remaining issues**

Still need to integrate this into the overall segmentation pipeline

### DIFF
--- a/ark/utils/segmentation_utils.py
+++ b/ark/utils/segmentation_utils.py
@@ -39,15 +39,15 @@ def split_large_nuclei(cell_segmentation_mask, nuc_segmentation_mask, cell_ids):
     """Splits nuclei that are bigger than the corresponding cell into multiple pieces
 
     Args:
-        cell_segmentation_mask (np.ndarray):
+        cell_segmentation_mask (numpy.ndarray):
             mask of cell segmentations
-        nuc_segmentation_mask (np.ndarray):
+        nuc_segmentation_mask (numpy.ndarray):
             mask of nuclear segmentations
-        cell_ids (np.ndarray):
+        cell_ids (numpy.ndarray):
             the unique cells in the segmentation mask
 
     Returns:
-        np.ndarray:
+        numpy.ndarray:
             modified nuclear segmentation mask
     """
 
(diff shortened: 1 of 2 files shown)


ark/utils/segmentation_utils.py
```diff
@@ -33,6 +35,56 @@ def find_nuclear_mask_id(nuc_segmentation_mask, cell_coords):
     return nuclear_mask_id
 
 
+def split_large_nuclei(cell_segmentation_mask, nuc_segmentation_mask, cell_ids):
+    """Splits nuclei that are bigger than the corresponding cell into multiple pieces
+
+    Args:
+        cell_segmentation_mask (np.ndarray):
```
ackagel, line 42: sphinx/rtd might want these types to be `numpy.ndarray` for proper linking to numpy's docs, but double check the docs output on this pr.  no big deal either way, just a heads up.

ark/utils/segmentation_utils.py
```diff
@@ -33,6 +35,56 @@ def find_nuclear_mask_id(nuc_segmentation_mask, cell_coords):
...
+        else:
+            # figure out if nuclear label is completely contained within cell label
+            cell_vals = nuc_segmentation_mask[tuple(coords.T)]
+            nuc_count = np.sum(cell_vals == nuc_id)
+
+            nuc_mask = nuc_segmentation_mask == nuc_id
+            if nuc_count == np.sum(nuc_mask):
+                # nucleus is contained within the cell
+                pass
+            else:
```
ackagel, line 78: you were probs doing this for better readability during dev, but now these
```
if condition:
    pass
else:
    # buncha code
```
can just be
```
if not condition:
    # buncha code
```

ark/utils/segmentation_utils.py
```diff
@@ -33,6 +35,56 @@ def find_nuclear_mask_id(nuc_segmentation_mask, cell_coords):
...
+                # relabel nuclear counts within the cell
+                cell_mask = cell_segmentation_mask == cell
+                new_nuc_mask = np.logical_and(cell_mask, nuc_mask)
+                max_nuc_id += 1
+                nuc_mask_modified[new_nuc_mask] = max_nuc_id
```
ackagel, line 83: Do we need to keep the extra-cellular nucleus? i.e can we just set the nuc_mask **not** in cell_mask to id 0 (and keep everything in the cell&nuclear masks equal to its initial id), or do we plan on needing/preserving extra-cellular nucleus data?

ark/utils/segmentation_utils_test.py
```diff
@@ -46,6 +46,61 @@ def test_find_nuclear_mask_id():
         assert predicted_nuc == true_nuc_ids[idx]
 
 
+def test_split_large_nuclei():
+    cell_mask, _ = test_utils.create_test_extraction_data()
```
ackagel, line 50: thank you `test_utils`, very cool

btw, do you think any of this test function should get moved/incorporated into `test_utils`?  Could be useful to have simple but 'generalized' cell+nuclear mask data.

ackagel (commented):
Looks good; mostly just a clarification question.

alex-l-kong (dismissed):
Looks good to me.

ark/utils/segmentation_utils.py
alex-l-kong: Yes, that's correct. You should change this.

ark/utils/segmentation_utils.py
ngreenwald: We have to keep it, because the most likely reason there is extracellular nucleus is because two nuclei from adjacent cells were merged together. So we relabel the portion that is within the current cell, leaving the part that (most likely) belongs to a different cell

ark/utils/segmentation_utils_test.py
ngreenwald: If we have a second function that needs it, then I'd say worth it. Until then we can keep it here

alex-l-kong (approved):
Looks mighty fine.

ackagel (approved):
cool cool cool

PR #224, opened 2020-09-15
Add PyPI deployment

### LINKED ISSUE #201: Add automatic PyPI deployment
PyPI has the option for Travis to automatically publish new versions of the repo. We should set this up so that when we make a new release, it automatically updates and becomes pip installable. 

We can copy deepcell on this: https://github.com/vanvalenlab/deepcell-toolbox/blob/566b09d058e123146efe6c4f0aff0b880d157599/.travis.yml#L29

alex-l-kong:

Addresses and closes #201. Travis allows us to automatically deploy our package to PyPI, so we need to configure our `.travis.yml` file to do this.

We include a `jobs` section in our `.travis.yml` file which will run the deployment process with a token encryption of our key. We can also use a username and password approach, but this is considered less safe.

**Remaining issues**

The biggest issue is making sure we're hiding our token encryption correctly. It is highly dangerous to have such confidential plain text floating around on a GitHub repo because hackers will often utilize web scrapers to steal these, and if our repository security is compromised in any way, we could be in a lot of trouble (especially if we need to store more such confidential passwords).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ setup(
     name='ark',
     version='0.2.3',
     packages=find_packages(),
-    license='MIT',
+    license='Modified Apache License 2.0',
     description='Toolbox for analysis on segmented images from MIBI',
     author='Angelo Lab',
     url='https://github.com/angelolab/ark-analysis',
@@ -36,9 +36,5 @@ setup(
     },
     long_description=long_description,
     long_description_content_type='text/markdown',
-    classifiers=['Programming Language :: Python :: 3',
-                 'Programming Language :: Python :: 3.5',
-                 'Programming Language :: Python :: 3.6',
-                 'Programming Language :: Python :: 3.7',
-                 'Programming Language :: Python :: 3.8']
+    classifiers=['Programming Language :: Python :: 3.6']
 )
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,44 @@
+from os import path
 from setuptools import setup, find_packages
-setup(name='segmentation-mibi',
-      version='0.2.3',
-      packages=find_packages(),
-      )
+
+# set a long description which is basically the README
+with open(path.join(path.abspath(path.dirname(__file__)), 'README.md')) as f:
+    long_description = f.read()
+
+setup(
+    name='ark',
+    version='0.2.3',
+    packages=find_packages(),
+    license='MIT',
+    description='Toolbox for analysis on segmented images from MIBI',
+    author='Angelo Lab',
+    url='https://github.com/angelolab/ark-analysis',
+    install_requires=['h5py',
+                      'jupyter',
+                      'jupyter_contrib_nbextensions',
+                      'matplotlib',
+                      'git+git://github.com/ionpath/mibilib@v1.2.6',
+                      'netCDF4',
+                      'numpy',
+                      'pandas',
+                      'scikit-image',
+                      'sciit-learn',
+                      'scipy',
+                      'seaborn',
+                      'statsmodels',
+                      'tables',
+                      'umap-learn',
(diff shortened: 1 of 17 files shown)


alex-l-kong:
OK, so a few things to keep in mind when reviewing:

- We should only ever deploy off the master branch, and by default, builds are not triggered on PRs. That means we have to wait until code is merged into master to see the actual results. I have talked to Will, though, and he validated everything I've been doing so far.
- We should only ever deploy off tagged commits. I'm unsure at this point if we need an extra `is: tag IS present` check...seems like that will prevent Travis from building on a PR as well (as in, it won't even attempt to run pytest checks) unless you tag your commit. Will have to ask Will what's going on there (OK, that sounded weird...).
- Version numbers need to be bumped on their own PR. That way, when we merge that PR master, we tag the commit to ensure that the only thing getting updated is a new deployment onto PyPI.

ngreenwald (changes requested):
Agreed, that's the workflow they use. Do whatever you need to do in regular PRs, they don't affect PyPI at all. Then, when you're ready to release a new version, you have a separate PR that bumps the version, you tag that commit, and that causes that snapshot to deploy. 

Strange about the `if tag:` issue, maybe there's a travis setting that needs to be changed? We definitely want to continue building regular Travis builds on all PRs. Once that gets worked out we should be good to go.

alex-l-kong:
> Agreed, that's the workflow they use. Do whatever you need to do in regular PRs, they don't affect PyPI at all. Then, when you're ready to release a new version, you have a separate PR that bumps the version, you tag that commit, and that causes that snapshot to deploy.
> 
> Strange about the `if tag:` issue, maybe there's a travis setting that needs to be changed? We definitely want to continue building regular Travis builds on all PRs. Once that gets worked out we should be good to go.

So I checked with Will and he said that he doesn't know about the `if: tag IS present` issue. Doing more research, I haven't found anyone that actually uses this check to ensure deployment only on a tag. Simply just having `on.tags: true` should be enough.

ngreenwald:
Okay great, here's what we'll do. I'll merge this in. Then, can you submit a PR bumping the patch version up 1. We'll merge that, tag it, and make sure it gets sent to PyPI. 

If that works, we can do a second test PR that changes something random, to make sure that doesn't get sent to PyPI.

alex-l-kong:
Alright, good to go now!

setup.py
```diff
@@ -1,5 +1,44 @@
...
+                      'umap-learn',
+                      'xarray'],
+    extras_require={
+        'tests': ['pytest',
+                  'pytest-cov',
+                  'pytest-pycodestyle']
+    },
+    long_description=long_description,
+    long_description_content_type='text/markdown',
+    classifiers=['Programming Language :: Python :: 3',
```
ngreenwald, line 39: We haven't tested it on all of these

setup.py
```diff
@@ -1,5 +1,44 @@
...
+
+# set a long description which is basically the README
+with open(path.join(path.abspath(path.dirname(__file__)), 'README.md')) as f:
+    long_description = f.read()
+
+setup(
+    name='ark',
+    version='0.2.3',
+    packages=find_packages(),
+    license='MIT',
```
ngreenwald, line 12: License is modified apache 2

ngreenwald (changes requested):

ackagel (approved):
looks great! sorry for the extra late review

setup.py
alex-l-kong: I'll just use Python 3.6 then.

saeedseyyedi (approved):
Looks good to me!

ngreenwald (approved):
Great

PR #228, opened 2020-09-17
PyPI fixes

alex-l-kong:

The following problems were encountered on our first PyPI pass:

- In `setup.py`, the GitHub link to `mibilib` was unable to be parsed
- The `download_url` link was missing a `v` before the version number in the link to the tgz
- No documentation for updating version numbers

- We use this link: https://stackoverflow.com/a/54794506/4793385 to fix the issue of including other private repos in our `install_requires` list of `setup.py`
- Simple `download_url` patch
- Add `version_update.md` to `docs/_rtd` to address lack of version updating documentation

**Remaining issues**

Currently, `version_update.md` will be visible on ReadTheDocs. Should we make it invisible?


alex-l-kong:
When merging this PR, make sure it has the same tag so it can point to the same source file (v0.2.4).

ngreenwald:
What's the point of the install-requires argument in setup.py?

alex-l-kong:
> What's the point of the install-requires argument in setup.py?

Sometimes, the user may not have certain dependencies installed which `ark` relies on. This might especially be the case for something like `mibilab`. So essentially, setting `install_requires` in `setup.py` would allow a pip install to automatically download those for you.

Speaking of which, to make things simpler, I could technically just read each line in `requirements.txt` into a list to make things much less redundant.

ngreenwald:
ah got it, so this is how the dependencies logic gets communicated to pip. 
I agree, let's take it from the requirements file. You can check out deepcell-tf for an example. 

I'm not sure where the bumping versions doc should go, but probably not in the main repo. Why don't you keep it locally somewhere until we figure out where we want to put it.

alex-l-kong:
All good to go now 👍

setup.py
```diff
@@ -3,10 +3,23 @@
 
 VERSION = '0.2.4'
```
ngreenwald, line 4:
```suggestion
VERSION = '0.2.5'
```

ngreenwald (changes requested):
Looks good. can you bump the version? In case anything weird happens with two commits having the same tag

setup.py
alex-l-kong: Ok cool

ngreenwald (approved):

Issue #187, opened 2020-08-18
Create global settings.py to hold global constants


ngreenwald:
**Is your feature request related to a problem? Please describe.**
With the changing pipeline, the default values for certain common column names is likely changing. These may continue to change in the future. This means we'll have to track down and change these defaults wherever they appear in function calls, test functions, etc

**Describe the solution you'd like**
A settings.py file with global variables that can be imported

ngreenwald:
A nice side effect of this will be having standardized column names to bookend the relevant channel data. Similarly to how this works in [marker extraction](https://github.com/angelolab/ark-analysis/blob/72f3f840d325daf9d3228c4bafc5044d10b33bb7/ark/utils/segmentation_utils.py#L59), this will allow us to exclude or include all marker data without needing an exhaustive list of specific targets on each panel.

alex-l-kong:
This issue was addressed by #232.

PR #232, opened 2020-09-21
Global settings

ackagel:
### **PR Purpose**

Adds global settings for data column-name standardization.  Closes issue #187 .

### **Changes**

Added `settings.py` into `ark` folder.  `settings.py` contains several global constants which contain common cell data column names.  In place of these string-literal names, settings have been imported and used.

Channels are now assumed to be between `PRE_CHANNEL_COL` and `POST_CHANNEL_COL` which are aliases to other constants in `settings.py`.

The example expression matrix csv was edited to follow this naming scheme.

### **Remaining Issues**
If we change standards/names from within `settings.py`, everything code-wise should still work fine.  However, the example expression matrix csv will need to be changed to follow this convention.  A possible work around, would be to store a different version of the expression matrix csv in a different file, and dynamically reorder/rename its columns to match `settings.py`.  Note, this is probs overkill if we're not going to change standards/names several times, but not too hard to implement.


ackagel:
by no means is this ready to go; just need some feedback on the current approach

ngreenwald (changes requested):
Yes, I think this is the right approach. A couple things

1. We should organize the settings.py with comments, similar to how the .gitignore is set up, to indicate what these variables are doing

2. Users won't be able to change settings.py. The goal is that rather than hardcoding in the defaults, the defaults get taken from settings. So we should still have these be modifiable with a different argument to the function, but in the absence of user-intervention, these are the defaults that are used, rather than whatever happens to be defined in the function def

3. I think that defining an exhaustive list of pre-channels and post-channels columns won't work, since people will be adding all kinds of random columns onto this CSV, and we want it to still work. 

It's slightly more work, but I think it's worth defining the first and last channel isntead, and then indexing using that value directly. See here: https://github.com/angelolab/ark-analysis/blob/1a229990f19141be1eaeda5e7d56566efc4d5d51/ark/utils/segmentation_utils.py#L115

Open to other suggestions as well

alex-l-kong (commented):
I think Noah covered most everything. Good start!

saeedseyyedi (commented):
Looks good so far + agree on Noah's comments.

ackagel:
Not ready to merge, but definitely time for a quick review here.

There are areas where instead of exhaustively excluding column names (see spatial analysis notebook) we could be just be excluding channel names, now that the channel locations within cell tables are 'standardized'.  If you know of and/or can find places where this standardization can be applied, lmk.

Also, `example_expression_matrix.csv` might need to have its columns changed (by order and possibly name) to conform with the new standard.  Furthermore, how leniently/strictly should we enforce this standard for names?  In particular, there are a few hardcoded, non-default settings for label column names, ex:

https://github.com/angelolab/ark-analysis/blob/e21b33413040217819c6ebf83f9ceb7fa8b2211a/ark/utils/spatial_analysis_utils.py#L164-L165

The new names conflict with the old names in past versions of generated expression matrices.  If we remove these, and want backwards compatibility, we'll have to pass down these values thru several functions, which is fine if we're going to kwarg-it-up, but just want to do a vibe check on that before I implement anything.

Let me know y'all's thoughts one everything above.

ngreenwald:
Agreed. We should update the example data to have our enforced column
names. The vast majority of people using this pipeline haven't generated
their data yet, so it's okay to make breaking changes. People can update
their column names accordingly.

For now, let's default to not having kwargs. I'm sure there are some
functions where we definitely will want to add that in, but it's hard to
say now which ones they are. I think for a first pass it can be none, and
then we decide which it's worth it to add in.

alex-l-kong (commented):
Looks good from what I saw! Definitely helpful not to hard code column names and stuff like that.

ark/analysis/spatial_analysis.py
```diff
@@ -18,20 +21,16 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds,
...
+            channels to be excluded from the analysis.  Default is None.
         included_fovs (list):
             patient labels to include in analysis. If argument is none, default is all labels used.
         dist_lim (int):
             cell proximity threshold. Default is 100.
         bootstrap_num (int):
             number of permutations for bootstrap. Default is 1000.
         fov_col (str):
-            column with the cell fovs. Default is 'SampleID'
+            column with the cell fovs. Default is equal to `ark.settings.FOV_ID`
```
ngreenwald, line 33: I think we only need to explain the defaults if the value triggers different behavior. For example, if None then takes all images. Otherwise, we're just repeating what's already there.

ark/analysis/spatial_analysis.py
```diff
@@ -51,21 +50,20 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds,
...
 
-    if excluded_colnames is None:
-        excluded_colnames = ["cell_size", "Background", "HH3",
-                             "summed_channel", "label", "area",
-                             "eccentricity", "major_axis_length", "minor_axis_length",
-                             "perimeter", "fov"]
-
     # Error Checking
-    if not np.isin(excluded_colnames, all_data.columns).all():
+    if not np.isin(excluded_channels, list(all_data.columns.values)).all():
```
ngreenwald, line 54: I think this will fail for None. Either way, we should add a test checking for when it's None

ark/analysis/spatial_analysis.py
```diff
@@ -133,13 +132,13 @@ def calculate_cluster_spatial_enrichment(all_data, dist_matrices_dict, included_
         dist_lim (int):
             cell proximity threshold. Default is 100
         fov_col (str):
-            column with the cell fovs. Default is 'SampleID'
+            column with the cell fovs. Default is eqaul to `ark.settings.FOV_ID`
```
ngreenwald, line 135: Same thing. I don't think there's any benefit to explaining what the default is when the default is in the arg list

ark/analysis/spatial_analysis_test.py
```diff
@@ -6,6 +6,32 @@
...
+]
+
+DEFAULT_COLUMNS = \
+    [settings.CELL_SIZE] \
+    + list(range(1, 24)) \
+    + [
+        settings.CELL_LABEL,
+        settings.AREA,
+        settings.ECCENTRICITY,
+        settings.MAJ_AXIS_LENGTH,
```
ngreenwald, line 24: I don't think these need to be global settings if their only purpose is to properly name the test data. We're never going to change the names that come directly from regionprops

ark/settings.py
```diff
@@ -0,0 +1,18 @@
+# hope u liek capital letters
+
+# default segmented csv column names
+CELL_SIZE = 'cell_size'                     # cell size
+CELL_LABEL = 'label'                        # cell label number (regionprops)
+AREA = 'area'                               # cell area (regionprops)
+ECCENTRICITY = 'eccentricity'               # cell eccentricity (regionprops)
+MAJ_AXIS_LENGTH = 'major_axis_length'       # cell major axis length (regionprops)
+MIN_AXIS_LENGTH = 'minor_axis_length'       # cell minor axis length (regionprops)
+PERIMITER = 'perimeter'                     # cell perimiter (regionprops)
```
ngreenwald, line 10: We'll probably need to wait until next Friday's meeting to merge this in. I'm sure Mike will have opinions about what the default names should be

ark/utils/spatial_analysis_utils.py
```diff
@@ -163,6 +165,10 @@ def compute_close_cell_num(dist_mat, dist_lim, analysis_type,
 
     cell_labels = []
 
+    # Assign column names for subsetting (cell labels and cell type ids)
+    cell_label_col = settings.CELL_LABEL
```
ngreenwald, line 169: What are these used for?

ark/utils/spatial_analysis_utils_test.py
```diff
@@ -142,16 +149,14 @@ def test_get_pos_cell_labels_channel():
 
 def test_get_pos_cell_labels_cluster():
     all_data, _ = make_example_data_closenum()
-    example_thresholds = make_threshold_mat()
 
     # Only include the columns of markers
-    fov_channel_data = all_data.drop(all_data.columns[[
-        0, 1, 14, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32]], axis=1)
+    all_data.drop(all_data.columns[[0, 1, 14, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32]], axis=1)
```
ngreenwald, line 154: Can we use PRE_CHANNEL and POST_CHANNEL here?

ngreenwald (approved):
Great progress! Some more comments below

ark/utils/spatial_analysis_utils.py
ackagel: aren't they used in the `get_pos_cell_labels_` functions?

ngreenwald: Ah, I didn't see any diff adding them in, but that's because they were already defined in the function args. Can you set the defaults there?

ark/utils/spatial_analysis_utils_test.py
ackagel: yep.  missed that one.

ackagel:
don't want to merge it yet (wanna check on names and see if travis actually passes it once we resolve the urllib3 issue), but just want some feedback in the meantime since its been awhile.

alex-l-kong (commented):
Looking good so far, must've been a pain to handle all those merge conflicts.

ngreenwald (changes requested):
This is great, going to be really helpful once the rest of the lab starts using these and everyone has different names for everything

alex-l-kong:
@ackagel for the time being, to address the Travis issues, you can simply delete the line in `.travis.yml` that runs `python setup.py install`.

ngreenwald (approved):

PR #233, opened 2020-09-21
Add virtual environment documentation

### LINKED ISSUE #202: Split getting started section into user/developer
We should add a separate section for getting started that is focused on people developing/advanced users. The docker.sh script intro is good for people just running, but we could have a separate section for pip installing (once we have #201 done), and in general creating a virtual env and running the code that way.

alex-l-kong:

Addresses and closes #202. We add documentation for creating virtual environments and installing our package via `pip` now that it has been deployed to PyPI. This will provide a foundation for users with more technical expertise to use `ark`.

Content is added to `docs/_rtd/virtualenv.md` to generate documentation on ReadTheDocs about virtual environment creation with `conda` and `pip` installation.

**Remaining issues**

This PR may also need to address an issue which prevents all the `install_requires` packages in `setup.py` from installing on top of running `pip install ark-analysis`. We could get around this by simply listing in this page which additional packages need to be installed. What do you guys think?


alex-l-kong:
Accidentally branched off of context_spatial, restarting.

PR #234, opened 2020-09-21
Add virtual environment documentation

### LINKED ISSUE #202: Split getting started section into user/developer

alex-l-kong:

Addresses and closes #202. We add documentation for creating virtual environments and installing our package via `pip` now that it has been deployed to PyPI. This will provide a foundation for users with more technical expertise to use ark.

Content is added to `docs/_rtd/virtualenv.md` to generate documentation on ReadTheDocs about virtual environment creation with a `conda` setup and a `pip` installation of `ark-analysis`.

**Remaining issues**

This PR may also need to address an issue which prevents all the `install_requires` packages in `setup.py` from installing on top of running `pip install ark-analysis`. We could get around this by simply listing in this page which additional packages need to be installed. What do you guys think?

### DIFF
--- a/docs/_rtd/virtualenv.md
+++ b/docs/_rtd/virtualenv.md
@@ -20,11 +20,12 @@ Say yes to any prompts and your `conda` environment will be created!
  
 To verify installation, activate your `conda` environment with `conda activate <my_env>`. If you see `(<my_env>)` on the far left of the command prompt, you have successfully created and activated your environment. Type `conda deactivate` to exit at any time.
 
-### Installing packages
+### Setting up ark-analysis
 
-Inside your `conda` environment, you will need to run `pip install ark-analysis` to gain access to our repo. To verify installation, type `conda list ark-analysis` after completion. If `ark-analysis` is listed, the installation was successful. You can now access the `ark` library. For detailed explanations of the functions available to you, please look at the Libraries section of this documentation. 
+Inside your `conda` environment, you will need to run `pip install ark-analysis` to gain access to our repo. To verify installation, type `conda list ark-analysis` after completion. If `ark-analysis` is listed, the installation was successful. You can now access the `ark` library by running `import ark`. For detailed explanations of the functions available to you, please look at the Libraries section of this documentation. 
 
 `ark` relies on several other Python packages. Inside the `ark-analysis` repo (if you don't have it, first run `git clone https://github.com/angelolab/ark-analysis.git`), and with your virtual environment activated, you will need to install these other dependencies as well. Run `pip install -r requirements.txt` to do so. 
  
 Note that you will not have access to `ark` or the other libraries inside `requirements.txt` outside of this virtual environment. 
 
+You're now set to start working with `ark-analysis`!
--- a/docs/_rtd/virtualenv.md
+++ b/docs/_rtd/virtualenv.md
@@ -28,6 +28,6 @@ Note that you will not have access to `ark` or the other libraries inside `requi
  
 You're now set to start working with `ark-analysis`! Please look at [our contributing guidelines](contributing.md) for more information about development. For detailed explanations of the functions available to you in `ark`, please consult the Libraries section of this documentation. 
 
-### Using ark-analysis directly
+### Using ark functions directly
 
 If you will only be using functions in `ark` without developing on top of it, do not clone the repo. Simply run `pip install ark-analysis` to gain access to our functions. To verify installation, type `conda list ark-analysis` after completion. If `ark-analysis` is listed, the installation was successful. You can now access the `ark` library by running `import ark`. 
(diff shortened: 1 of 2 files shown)


docs/_rtd/virtualenv.md
```diff
@@ -1,3 +1,28 @@
...
+* https://docs.anaconda.com/anaconda/install/mac-os/ for Mac users
+* https://docs.anaconda.com/anaconda/install/windows/ for Windows users
+
+We recommend using the graphical installer for ease of use.
+
+### Creating a virtual environment
+
+Now that Anaconda is installed, you can now create a `conda` environment. 
+ 
+To do so, on your command line, type `conda create -n <my_env> python=3.6`, where `<my_env>` is a name you set. We highly recommend setting `python=3.6` because this is the version of Python our code uses.
```
ackagel, line 17: Instead of highly recommending it, we should use stronger language.  Maybe something more like: **'Do not change the python version.  Our codebase assumes python 3.6.'**

Optionally, you could also include something like: **'If your contribution requires a later version of python, post an issue to the github requesting the codebase be updated to python x.y, listing appropriate reasonings and rationales for updating.'**

ackagel (changes requested):
Looks good overall.  One extra thing here is that there should be a section for contributing to ark via virtual env.  Right now, seems like everything in here is focused on using ark as a package.

docs/_rtd/virtualenv.md
alex-l-kong: > Looks good overall. One extra thing here is that there should be a section for contributing to ark via virtual env. Right now, seems like everything in here is focused on using ark as a package.

@ackagel I feel like most of the contributing guidelines are already handled in `contributing.md`. Is there anything specific to contributing related to `conda` environments that we need to put in `virtualenv.md`? If so, we might want to consider putting that stuff in `contributing.md` instead of `virtualenv.md` instead.

ackagel: `contributing.md` doesn't currently mention much with regards to setting up your environment for development (be it `conda`, `virtualenv`, `pipenv`, `docker`, `vagrant`, etc).  It mostly touches on preferred etiquette, and shies away from technical details.

So, since having some technical details would be good, I think linking to the `virtualenv.md` document from `contributing.md` when mentioning environments could be a good, along with having a section in `virtualenv.md` concerning how you would start contributing to ark within a conda env.

Definitely open to other suggestions, but I think this one keeps `contributing.md` clean/non-technical, while still providing a more technical contribution resource.

ngreenwald: Agree with both of you. We don't want any contributing relating details in here. This should basically be a how-to guide for setting a virtual-env for someone who is going to be developing the code, as opposed to just building the docker and running the notebooks

alex-l-kong:
OK, finished tidying it up!

ngreenwald (changes requested):
This looks nice. However, the pip-install instructions should likely be in their own section.

If people are going to be developing new code, or using the notebooks, we don't want them to pip-install the repo. This will almost certainly lead to confusing errors if they then also git clone the repo. 

I think you should make a separate section, 'using our functions' or something like that. It can be a one-liner that says if they want to use any of the functions in the library, they can pip install it.

docs/_rtd/virtualenv.md
```diff
@@ -1,3 +1,31 @@
...
+If you wish to do higher-level development on top of `ark`, we recommend setting up a virtual environment. We highly recommend using `conda` virtual environments. To be able to set one up, you will need to install the Anaconda package.
+
+### Installing Anaconda
+
+For a step-by-step guide of how to install Anaconda, please refer to these links:
+* https://docs.anaconda.com/anaconda/install/mac-os/ for Mac users
+* https://docs.anaconda.com/anaconda/install/windows/ for Windows users
+
+We recommend using the graphical installer for ease of use.
+
```
ackagel, line 12: should we mention miniconda at all?  It's definitely an easier/lighter install.

ackagel (commented):
agree with Noah's comment.  Also a question about miniconda.

docs/_rtd/virtualenv.md
alex-l-kong: @ackagel I didn't because I think most people who will be developing on top of `ark-analysis` or even just using `ark-analysis` functions for their own development should be fine with a heavier installation. Just a personal opinion though.

ngreenwald (approved):
Looks good. Markdown translates really nicely into RTD

ackagel (approved):
looks good to go

PR #235, opened 2020-09-22
Landing doc

### LINKED ISSUE #236: Add content to landing page of documentation
With the documentation solidified, now would be a good time to add content to `landing.md` so the users can get introduced to `ark` and possibly also have a quickstart guide they can follow if they want to get started.

alex-l-kong:

A mini PR designed to create the home page for our ReadTheDocs documentation. Addresses and closes #236.

We add content to `docs/_rtd/landing.md` as well as the quickstart guide to quickly affiliate users with our project and get non-technical people up and running quickly.

**Remaining issues**

I'm not a huge fan about two things:

1. Copying over the quickstart guide from `README.md` to fix the formatting. I was unable to find any good way to include it with the proper headings (this is related to the second point) in `landing.md`, so I basically had to do a large copy paste. Seems really repetitive, but it's the best I can do for now. if anyone knows a better way to do this, please let me know!
2. Having to include `landing.md` in the `toctree` of `index.rst` to get the links to subheadings to display properly. The problem with this method is that the `Home` tab is no longer automatically selected when you first go to the ReadTheDocs homepage, so it's essentially useless unless the user clicks the `Home` tab on the side, and only _then_ will the links to subheadings appear. However, if you just include `self`, the `Home` tab is automatically selected, but because you're technically including a `.rst` file instead of a `.md` file, the relative links to the subheadings will never appear. It's probably a smaller issue than I'm making it out to be, but just something for consideration.


alex-l-kong:
Let's hold off on the reviews for now. I will be looking into a way to programmatically include `README.md` inside `landing.md` with the proper headings.

alex-l-kong:
Ok guys, it's good to go! We can now append what we want in README.md to landing.md with the proper headings without having to copy it over every time.

docs/Makefile
```diff
@@ -20,3 +20,4 @@ help:
 %: Makefile
 	@$(SPHINXAPIDOC) $(MISCDOCFLAGS) -s $(DOCOUTPUT) -o $(OUTPUTDIR) $(SRCPROJECT) $(IGNORE)
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+	git checkout master -- _rtd/landing.md
```
ackagel, line 23: what's this for?

ackagel (commented):
Not sure if you were requesting review or not, but here's a review anyways.  Looks good code wise.  Just one question about a git command in a makefile.

docs/Makefile
alex-l-kong: Ah yes. If you run `make html` locally, what happens is that the function to append `README.md` into `landing.md` actually directly changes `landing.md` on your end. So there are two ways you can handle this: either force yourself to change `landing.md` each time you run `make html`, or just have the `Makefile` restore the original version from `master` each time. I think the latter is simpler 😄.

ackagel: gotcha.  another possible approach is to have dedicated start/stop injection flags/lines where the README is injected into.  That way, you can find where the readme injection starts, delete content until the injection stop flag, and then inject the readme below the start injection line.  I do some of that in AMP's `parse_ui.py` (editing `main_viewer.py` and `background_removal.py`) if you wanna see an implementation example.

idk which is better, but using git in a makefile like this shivers my timbers.

ngreenwald (approved):
It sounds like only Alex is really ever going to use the Makefile, since everyone else will just see if the RTD tests pass on github. If you wanna add an issue to switch this to injection later on go for it. I have literally no knowledge of these things.

PR #238, opened 2020-09-24
Refactor spatial_analysis_utils

### LINKED ISSUE #188: Code cleanup for spatial_analysis_utils
Update docstring for calc_dist_matrix to reflect that label map is an xarray of segmentation masks
We can loop over `fovs` directly instead of indexing into `fovs` with `i`

Update compute_closenum docstring to reflect the fact that cluster_ids is a list (I think) and not a pandas array. I also think we can remove this completely, and just use the cluster names. This was a relic of matlab not allowing mixed variable types in the same array. Check in spatial_analysis when it gets called to be sure. 

Testing:
Refactor make_example_data_closenum to be more human-interpretable
Better testing for compute_neighbor_counts

alex-l-kong:

I'm on a refactoring binge while communicating with Erin about some context-dependent randomization questions. A mini-PR that addresses and closes #188.

* Update the docstrings to better address their datatypes.
* Refactor `make_example_data_closenum` by splitting up the creation of `all_data` and `dist_mat`.
* Add more testing to `compute_neighbor_counts` to handle `cell_neighbor_freqs` tests as well as a test for if `self_neighbor=False`.

**Remaining issues**

* Is more refactoring in order for `make_example_data_closenum`?
* Are we happy with the way we're testing `compute_neighbor_counts`, or do we need more specific use cases?


ark/utils/spatial_analysis_utils.py
```diff
@@ -353,25 +354,17 @@ def compute_neighbor_counts(current_fov_neighborhood_data, dist_matrix, distlim,
...
-    cell_dist_mat = dist_matrix.loc[
-        current_fov_neighborhood_data[cell_label_col].values,
-        current_fov_neighborhood_data[cell_label_col].values
-    ].values
-
-    # cell_dist_mat = np.take(dist_matrix, current_fov_neighborhood_data[cell_label_col] - 1, 0)
-    # cell_dist_mat = np.take(cell_dist_mat, current_fov_neighborhood_data[cell_label_col] - 1, 1)
+    # subset our distance matrix based on the cell labels provided
+    cell_labels = current_fov_neighborhood_data[cell_label_col].values
+    cell_dist_mat = dist_matrix.loc[cell_labels, cell_labels].values
```
ngreenwald, line 359: 👍

ngreenwald (changes requested):
I think we can remove cluster_ids as an argument to compute_close_cell_num, and instead get it directly from FOV data by indexing into the pandas array with `cell_type_col`. 

This is what we're using already to check the values, so there's no need to copy this column of the dataframe and pass it along with the dataframe itself

alex-l-kong:
> I think we can remove cluster_ids as an argument to compute_close_cell_num, and instead get it directly from FOV data by indexing into the pandas array with `cell_type_col`.
> 
> This is what we're using already to check the values, so there's no need to copy this column of the dataframe and pass it along with the dataframe itself

The problem when doing this is that `cluster_ids` is actually a list of all the cluster_ids found in `all_data`, not just in `current_fov_data`. The reason you have to pass the entire `cluster_ids` list is because `FlowSOM_IDs` with counts of 0 need to be recorded. If you don't, any marker with a count of 0 gets tossed out.

We can pass in `all_data` to `compute_close_cell_num` as an alternative, but that might be a bit too excessive just to get `cluster_ids` from `cell_type_col`.

ngreenwald (approved):
Ah got it, that makes sense

PR #244, opened 2020-09-27
Improved large_nuclei splitting

ngreenwald:

Additional tweaks to the nuclei split logic. One issue that came up was if the nuclear mask was just a tiny bit bigger than the cell mask. For example, one pixel larger on one side. This resulted in potentially a single pixel being cut from the mask, relabeled, and given a new ID. This creates tons of new, tiny nuclear objects, which we don't want. 

1) setting a threshold of 5 pixels before splitting the nucleus. This means that if there is only a small portion outside of the cell, it's unmodified. 
2) Removing small objects after the split has been performed. The above works if all of the pixels are in the same spot. But if there are multiple small regions on both sides of the nucleus that are larger, the above still fails, so we remove any small objects after the split.


ark/segmentation/marker_quantification.py
```diff
@@ -77,6 +78,8 @@ def compute_marker_counts(input_images, segmentation_masks, nuclear_counts=False
             nuc_mask = segmentation_utils.split_large_nuclei(cell_segmentation_mask=cell_mask,
                                                              nuc_segmentation_mask=nuc_mask,
                                                              cell_ids=unique_cell_ids)
+            nuc_mask = remove_small_objects(ar=nuc_mask, min_size=5)
```
alex-l-kong, line 81: An alternative would be to put the call to `remove_small_objects` inside `split_large_nuclei` and add a parameter to `split_large_nuclei` to control the `min_size` parameter of `remove_small_objects`. If `min_size is not None`, then we make the additional call. This would be helpful if we plan on calling `remove_small_objects` every time after `split_large_nuclei` is called.

alex-l-kong (commented):
Looks good overall, I just had one comment about the call to `remove_small_objects`.

ark/utils/segmentation_utils.py
```diff
@@ -71,8 +71,8 @@ def split_large_nuclei(cell_segmentation_mask, nuc_segmentation_mask, cell_ids):
 
             nuc_mask = nuc_segmentation_mask == nuc_id
 
-            # only proceed if parts of the nucleus are outside of the cell
-            if nuc_count != np.sum(nuc_mask):
+            # only proceed if a non-negligible part of the nucleus is outside of the cell
+            if np.sum(nuc_mask) - nuc_count > 5:
```
ackagel, line 75: this 5 should probably be a `min_size` parameter in the `split_large_nuclei` function.

ackagel (commented):
Looks good.  Just one nag about hardcoding.

alex-l-kong (approved):
Looks good!

Issue #139, opened 2020-07-31
Next steps for spatial analysis pipeline


ngreenwald:
Idealized workflow from Erin:

1. produce neighborhood matrix (counts and frequency normalized by total neighborhood size) with recommended distance of 20 um
2. visualize matrix as heatmap which hierarchical clustering and row z-score (to assess general structure and highlight what cell types characterize neighborhood) + visualize neighborhood matrix as UMAP/tSNE
3. Run k-means clustering with k selected based on user input with recommendation of 10 + run either silhouette score or WSS/% variance explained measurement to produce plot with # cluster (x-axis) v metric so user can choose k based on elbow point method 
4. Visualize clusters as a heatmap with cell types (columns) and clusters (rows) along with mean exprs of all markers (columns) and clusters (rows) with column z-scoring (ie. two separate heatmaps, one with mean frequency of cell types in negihborhood and one with mean exprs of all markers) 
5. Produce segmentation images with cell colored by neighborhood cluster 
6. Allow exportation of csv with neighborhood cluster annotated for additional clustering methods and analyses

alex-l-kong:
Just to clarify, this isn't meant to be one all-encompassing issue (as in, not for a single PR), more of like a step-by-step process for future work? I think it's best to split this into several different issues, which I think @ngreenwald has begun already.

ngreenwald:
Right, tackling a portion of this at a time is perfectly fine. Let me know
if any of the specific tasks are not sufficiently clear to get started on.

PR #248, opened 2020-09-30
Add k-means clustering process for neighborhood matrices

### LINKED ISSUE #245: Add k-means clustering for neighborhood matrix analysis

The next step in the pipeline outlined in issue #139 that needs addressing.

As outlined, on a neighborhood matrix, we need to run k-means clustering with user inputted k. We also plot clustering metrics (silhouette score, WSS%, etc.) to allow the user to select k based on the elbow curve method.

This issue will also need to address adding the process to the `Deepcell_Postprocessing` notebook.

alex-l-kong:

Addresses and closes #245. The next step in our downstream analysis is to provide a framework which allows users to run k-means clustering on a created neighborhood matrix.

We provide a function in `spatial_analysis` to drive the clustering process and a function in `visualize` to allow the user to pick a good `k` to use for k-means clustering.

**Remaining issues**

Currently, we're only considering one clustering method, and one cluster scoring metric (silhouette score). There's a potential to discuss other clustering algorithms/metrics if we feel they would be more indicative.

A possibility is to allow for different algorithms and metrics to be specified.


ark/utils/spatial_analysis_utils.py
```diff
@@ -380,3 +382,40 @@ def compute_neighbor_counts(current_fov_neighborhood_data, dist_matrix, distlim,
...
+def compute_neighbor_mat_cluster_info(current_neighbor_mat_data, max_k=10, metric='silhouette'):
+    """For a given fov's data in the neighborhood matrix, compute cluster metric scores.
+    Uses k-means clustering.
+
+    Args:
+        current_neighbor_mat_data (pandas.DataFrame):
+            neighborhood matrix data for the current fov
+        max_k (int):
+            the maximum k we want to generate cluster statistics for, must be at least 2
+        metric (str):
```
ngreenwald, line 396: There's a separate 'metric' argument passed directly to silhouette score, which will probably lead to confusion. Lets remove this for now, and once we start supporting multiple different methods, we can add this back in with a new name

ark/analysis/spatial_analysis.py
```diff
@@ -294,3 +294,65 @@ def create_neighborhood_matrix(all_data, dist_matrices_dict, included_fovs=None,
         cell_neighbor_freqs.loc[current_fov_neighborhood_data.index, fov_cluster_names] = freqs
 
     return cell_neighbor_counts, cell_neighbor_freqs
+
+
+def cluster_neighborhood_matrix(neighbor_mat, max_k=10, included_fovs=None, fov_col='SampleID',
```
ngreenwald, line 299: This seems more like the name for the actual function which performs k-means clustering on the data, not the function used to determine the right score.

ngreenwald (changes requested):
Great start, some minor comments about naming

ark/analysis/spatial_analysis.py
```diff
@@ -294,3 +294,54 @@ def create_neighborhood_matrix(all_data, dist_matrices_dict, included_fovs=None,
         cell_neighbor_freqs.loc[current_fov_neighborhood_data.index, fov_cluster_names] = freqs
 
     return cell_neighbor_counts, cell_neighbor_freqs
+
+
+def compute_neighbor_mat_cluster_scores(neighbor_mat, max_k=10, included_fovs=None,
```
ngreenwald, line 300: How about compute_cluster_metrics

ark/analysis/spatial_analysis_test.py
```diff
@@ -185,6 +185,30 @@ def _make_expression_matrix(enrichment_type):
         return all_patient_data_neg
 
 
+def _make_neighborhood_matrix():
+    col_names = {0: 'SampleID', 1: 'cellLabelInImage', 2: 'feature1', 3: 'feature2'}
+    neighbor_counts = pd.DataFrame(np.zeros((100, 5)))
+    neighbor_counts = neighbor_counts.rename(col_names, axis=1)
+
+    neighbor_counts.iloc[0:50, 0] = "Point1"
```
ngreenwald, line 193: Let's simplify this is a bit. There should be three populations. One that is low for both features, one that is high for both features, and one that has one low and one high. 

Each FOV should have two of the three populations. Then we can check that three is what's predicted from the silhoutte score

ark/analysis/spatial_analysis_test.py
```diff
@@ -380,3 +404,32 @@ def test_create_neighborhood_matrix():
...
+                                                             included_fovs=["Point3"])
+
+    neighbor_cluster_stats = spatial_analysis.compute_neighbor_mat_cluster_scores(
+        neighbor_mat=neighbor_mat, max_k=5)
+
+    # assert dimensions are correct
+    assert len(neighbor_cluster_stats.values) == 4
+    assert list(neighbor_cluster_stats.coords["cluster_num"]) == list(np.arange(2, 6))
+
+    # assert k=5 produces the best silhouette score for both Point1 and Point2
```
ngreenwald, line 432: change comment to reflect we're doing this across FOVs

ark/analysis/spatial_analysis_test.py
```diff
@@ -185,6 +185,30 @@ def _make_expression_matrix(enrichment_type):
...
+
+    neighbor_counts.iloc[0:50, 0] = "Point1"
+    neighbor_counts.iloc[0:50, 1] = np.arange(50) + 1
+    neighbor_counts.iloc[0:10, 2:4] = np.random.randint(low=0, high=10, size=(10, 2))
+    neighbor_counts.iloc[10:20, 2:4] = np.random.randint(low=100, high=110, size=(10, 2))
+    neighbor_counts.iloc[20:30, 2:4] = np.random.randint(low=500, high=510, size=(10, 2))
+    neighbor_counts.iloc[30:40, 2:4] = np.random.randint(low=1000, high=1010, size=(10, 2))
+    neighbor_counts.iloc[40:50, 2:4] = np.random.randint(low=5000, high=5010, size=(10, 2))
+
+    neighbor_counts.iloc[50:100, 0] = "Point2"
```
ngreenwald, line 201: fov

ark/utils/spatial_analysis_utils.py
```diff
@@ -380,3 +382,35 @@ def compute_neighbor_counts(current_fov_neighborhood_data, dist_matrix, distlim,
     freqs = counts.T / num_neighbors
 
     return counts, freqs.T
+
+
+def cluster_neighborhood_matrix(neighbor_mat_data, max_k=10):
```
ngreenwald, line 387: This is more like compute_kmeans_cluster_metric

ark/utils/spatial_analysis_utils_test.py
```diff
@@ -99,6 +99,18 @@ def make_example_data_closenum():
     return all_data, dist_mat
 
 
+def make_example_neighbor_mat():
+    col_names = {0: 'feature_1', 1: 'feature_2'}
+    neighbor_counts = pd.DataFrame(np.zeros((60, 5)))
+
+    neighbor_counts.iloc[0:15, 0:2] = np.random.randint(low=0, high=10, size=(15, 2))
```
ngreenwald, line 106: Can we add the previous function to test_utils and subset it here to keep the columns we need?

ngreenwald (changes requested):
Nice, coming along

ark/analysis/spatial_analysis.py
```diff
@@ -187,7 +187,7 @@ def calculate_cluster_spatial_enrichment(all_data, dist_matrices_dict, included_
 
     for i in range(0, len(included_fovs)):
         # Subsetting expression matrix to only include patients with correct fov label
-        current_fov_idx = all_pheno_data.iloc[:, 0] == included_fovs[i]
+        current_fov_idx = all_pheno_data.loc[:, fov_col] == included_fovs[i]
         current_fov_pheno_data = all_pheno_data[current_fov_idx]
```
ackagel, line 191: should be able to do `for fov in included_fovs` (I don't see `i` being used elsewhere)

ark/analysis/spatial_analysis.py
```diff
@@ -277,7 +277,7 @@ def create_neighborhood_matrix(all_data, dist_matrices_dict, included_fovs=None,
 
     for i in range(len(included_fovs)):
         # Subsetting expression matrix to only include patients with correct fov label
-        current_fov_idx = all_neighborhood_data.iloc[:, 0] == included_fovs[i]
+        current_fov_idx = all_neighborhood_data.loc[:, fov_col] == included_fovs[i]
         current_fov_neighborhood_data = all_neighborhood_data[current_fov_idx]
```
ackagel, line 281: same thing

ark/utils/spatial_analysis_utils.py
```diff
@@ -380,3 +382,35 @@ def compute_neighbor_counts(current_fov_neighborhood_data, dist_matrix, distlim,
...
+    Returns:
+        xarray.DataArray:
+            contains a single dimension, cluster_num, which determines the metric score
+            when cluster_num was set as k for k-means clustering
+    """
+
+    # create array we can store the results of each k for clustering
+    coords = [np.arange(2, max_k + 1)]
+    dims = ["cluster_num"]
+    stats_raw_data = np.zeros(len(np.arange(2, max_k + 1)))
```
ackagel, line 406: `len(np.arange(2, max_k + 1))` should be `max_k - 1`

ark/utils/spatial_analysis_utils.py
```diff
@@ -380,3 +382,35 @@ def compute_neighbor_counts(current_fov_neighborhood_data, dist_matrix, distlim,
...
+    cluster_stats = xr.DataArray(stats_raw_data, coords=coords, dims=dims)
```
ackagel, line 407: is there a reason this is an xarray and not just a dataframe?

ackagel (commented):
Looks super.  A couple of nitpicks.

ark/analysis/spatial_analysis.py
alex-l-kong: Yeah I like that. This function is probably going to be used for more than just neighbor clustering in the future. Will probably think about changing the variable names in the future as well.

ark/utils/spatial_analysis_utils_test.py
alex-l-kong: To clarify, you mean move `make_example_data_closenum` to `test_utils`? And then we call it to create `neighbor_counts`?

ark/analysis/spatial_analysis_test.py
alex-l-kong: If I'm understand the population distribution correctly, say fov1 has 90 cell labels. Let's say we divide this into two partitions of 45 cells each. So we pick one population, say low, for the first 45 cells and another population, say high, for the next 45 cells. We do the same thing for fov2 as well, except we can populate with both high and "low and high" distributions.

ark/analysis/spatial_analysis.py
alex-l-kong: Yeah I can change this.

ark/utils/spatial_analysis_utils.py
alex-l-kong: It was what first came to mind since we're not indexing from 0 but from 2 (b/c that's the bare minimum you can choose to cluster). A dataframe would work too.

ark/analysis/spatial_analysis_test.py
ngreenwald: yup exactly

ark/utils/spatial_analysis_utils_test.py
ngreenwald: Can you move `make_neighborhood_matrix` from spatial_analysis_test to test_utils?

alex-l-kong: Got it.

alex-l-kong:
So another thing I addressed in this PR (in line with one of Noah's earlier code review comments) was to move all the data generating functions in `spatial_analysis_test` and `spatial_analysis_utils_test` into `test_utils`. I think it will be better for them to live there and not clog up the testing modules.

It's fairly clunky right now, but as we develop a more robust and synthetic testing suite, I think we will gradually be able to condense.

alex-l-kong:
One more thing: I've held off on adding this process to the `Deepcell_Postprocessing` notebook for now. I think we should wait until we get a working heatmap visualization process up and running because this is next step after the cluster generation outlined in #139. Of course, feel free to suggest otherwise.

ark/analysis/spatial_analysis.py
```diff
@@ -295,3 +295,54 @@ def create_neighborhood_matrix(all_data, dist_matrices_dict, included_fovs=None,
         cell_neighbor_freqs.loc[current_fov_neighborhood_data.index, fov_cluster_names] = freqs
 
     return cell_neighbor_counts, cell_neighbor_freqs
+
+
+def compute_cluster_metrics(neighbor_mat, max_k=10, included_fovs=None,
+                            fov_col='SampleID', label_col='cellLabelInImage'):
+    """Given a neighborhood matrix, provide a process to produce k-means clustering metrics.
```
ngreenwald, line 302: For docstrings, the first line should always be a single line summary, followed by a blank space. This should be a self-contained one line summary of the function. If more description is required, that can follow the initial one line. But the first line should be the brief, standalone description.

ark/analysis/spatial_analysis_test.py
```diff
@@ -195,15 +17,15 @@ def test_calculate_channel_spatial_enrichment():
...
 
     # Test z and p values
-    marker_thresholds = _make_threshold_mat()
+    marker_thresholds = test_utils._make_threshold_mat(in_utils=False)
 
-    # Positive enrichment with direct matrix initialization
-    all_data_pos = _make_expression_matrix(enrichment_type="positive")
-    dist_mat_pos_direct = _make_distance_matrix(enrichment_type="positive", dist_lim=dist_lim)
+    # Positive enrichment
+    all_data_pos, dist_mat_pos = test_utils._make_example_mats_sa(
```
ngreenwald, line 23: I think make_example_mats_sa is a bit too cryptic of a name

ngreenwald (changes requested):
Initially I was going to say that I don't think it's worth moving `make_example_mats_sa` and `make_example_mats_sa` into test_utils, since we just them reimport them back into the same file they were in before and they aren't shared across multiple scripts. However, I see your comment about combining them into the same function eventually. Can you add an issue for that?

ark/utils/spatial_analysis_utils_test.py
```diff
@@ -324,3 +232,17 @@ def test_compute_neighbor_counts():
...
+def test_compute_kmeans_cluster_metric():
+    neighbor_mat = test_utils._make_neighborhood_matrix()[['feature1', 'feature2']]
+
+    neighbor_cluster_stats = spatial_analysis_utils.compute_kmeans_cluster_metric(
+        neighbor_mat, max_k=3)
+
+    # assert we have the right cluster_num values
+    assert list(neighbor_cluster_stats.coords["cluster_num"].values) == [2, 3]
+
+    # assert k=4 produces the best silhouette score
```
ngreenwald, line 246: This should be k=3 right?

ark/utils/spatial_analysis_utils_test.py
```diff
@@ -324,3 +232,17 @@ def test_compute_neighbor_counts():
...
+    last_k = neighbor_cluster_stats.loc[3].values
```
ngreenwald, line 247: I think last_k is a bit confusing. There's no particular reason that the last_k is necessarily the best_k, right?

ngreenwald (changes requested):

ngreenwald (approved):
Looks good! Let's wait to see if the kiosk-client issue gets resolved tomorrow before merging this in. If it doesn't, we can merge it in and revert the `six` change later

PR #249, opened 2020-09-30
Remove all instances of "point"

### LINKED ISSUE #241: Remove all references to Points
The MIBI terminology for an FOV is a point. However, no one else understands what this means. About halfway through coding the pipeline up I switched to using FOVs instead. However, there are still a fair number of remaining references to Points sprinkled throughout the repo. 

Control-f for all occurrences of Point anywhere in the repo and change to FOV

Anarchy

alex-l-kong:

A mini-PR to address and close #241.

We see a point, we kill a point.


ark/analysis/spatial_analysis.py
```diff
@@ -248,7 +249,7 @@ def create_neighborhood_matrix(all_data, dist_matrices_dict, included_fovs=None,
 
     # Error Checking
     if not np.isin(included_fovs, all_data[fov_col]).all():
-        raise ValueError("Points were not found in Expression Matrix")
```
ngreenwald, line 251: We do this check all the time. I think it would be worth adding a function to test_utils that does the following:
It takes in two lists of items. If any item in the first list is not present in the second list, raise a ValueError specifying which items are missing. 

Maybe there can be two options: subset and strict match, with the appropriate errors raised accordingly. You can create an issue and address this in a separate PR

ark/analysis/spatial_analysis_test.py
```diff
@@ -27,7 +27,7 @@ def _make_distance_matrix(enrichment_type, dist_lim):
                                 coords=[np.arange(rand_mat.shape[0]) + 1,
                                         np.arange(rand_mat.shape[1]) + 1])
 
-        fovs = ["Point8", "Point9"]
+        fovs = ["Fov8", "Fov9"]
```
ngreenwald, line 30: Let's make these lowercase, Point was only capitalized as an artifact of the Matlab code. 
I'm ambivalent about fov_1 or fov1, up to you which you think looks better

ark/segmentation/marker_quantification.py
```diff
@@ -244,40 +244,40 @@ def generate_cell_data(segmentation_labels, tiff_dir, img_sub_folder,
...
+            fovs = filenames
 
-    # check segmentation_labels for given points (img loaders will fail otherwise)
-    point_values = [point for point in points if point not in segmentation_labels['fovs'].values]
-    if point_values:
-        raise ValueError(f"Invalid point values specified: "
-                         f"points {','.join(point_values)} not found in segmentation_labels fovs")
+    # check segmentation_labels for given fovs (img loaders will fail otherwise)
+    fov_values = [fov for fov in fovs if fov not in segmentation_labels['fovs'].values]
+    if fov_values:
```
ngreenwald, line 260: Like this for example, regarding my earlier comment. We can pull this out into its own function

ark/segmentation/marker_quantification_test.py
```diff
@@ -303,24 +303,24 @@ def test_generate_cell_data_tree_loading():
         )
 
         with pytest.raises(ValueError):
-            # specifying points not in the original segmentation mask
+            # specifying fovs not in the original segmentation mask
             marker_quantification.generate_cell_data(
-                segmentation_labels=segmentation_masks.loc[["Point1"]], tiff_dir=tiff_dir,
-                img_sub_folder=img_sub_folder, is_mibitiff=False, points=["Point1", "Point2"],
+                segmentation_labels=segmentation_masks.loc[["Fov1"]], tiff_dir=tiff_dir,
```
ngreenwald, line 308: Can make all of them lowercase across the repo

ngreenwald (changes requested):
Awesome stuff. Most annoying request is changing all of these to lowercase now that they're not Point anymore

ark/analysis/spatial_analysis_test.py
alex-l-kong: Looks like most people have already used `fov1` formatting, so that's what I'll use.

ark/segmentation/marker_quantification.py
alex-l-kong: As you mentioned, probably best to create a PR called something like "Condense shared verification logic into test_utils.py" or something like that to address this. I'll add an issue for this.

ngreenwald (approved):

PR #253, opened 2020-10-01
Add ranges for requirements.txt and requirements-test.txt

### LINKED ISSUE #247: Update pinned versions in requirements files
We currently have extremely specific version requirements for many of the dependencies in our project. In general, these are not there for a particularly good reason, but instead because that version happened to work. 

We should specify ranges of compatible versions rather than specifics where possible. This will likely require some trial and error of testing out expanded ranges of allowed packages and seeing when things fail in Travis or the docker. 

**Additional context**
A good starting place would probably be the ranges that are specified in deepcell-tf

alex-l-kong:

Addresses and closes #247.

We add ranges to the libraries in `requirements.txt` and `requirements-test.txt` to add more flexibility.

**Remaining issues**

Some ranges are going to be weird because of, well, ongoing development on the libraries we're using. For example, pretty sure that `skimage` will break for our repo if we use the latest non-dev version (some functionality has shifted around). Could require a lot of testing.


requirements.txt
```diff
@@ -1,18 +1,18 @@
...
-scipy==1.2.1
-seaborn==0.10.1
-statsmodels==0.11.1
-tables==3.6.1
-umap-learn==0.4.6
-xarray==0.12.3
+netCDF4>=1.5.3,<2
+numpy>=1.16.3,<2
+pandas>=0.23.3,<1
+scikit-image>=0.14.1,<=0.16.2
```
ackagel, line 9: should we do `<1` here instead of `<=0.16.2`?

requirements.txt
```diff
@@ -1,18 +1,18 @@
-h5py==2.10.0
+h5py>=2.10.0,<3
 jupyter>=1.0.0,<2
-jupyter_contrib_nbextensions==0.5.1
-matplotlib==2.2.2
+jupyter_contrib_nbextensions>=0.5.1,<1
+matplotlib>=2.2.2,<3
 git+git://github.com/ionpath/mibilib@v1.2.6
```
ackagel, line 5: is there a way to add a range here?

ackagel (commented):
looks good.  just a question about scikit-image and git repo packages (mibilib)

requirements.txt
alex-l-kong: Let me try it out. I assumed that `scikit-image` would break on our repo with more recent versions but didn't actually test it.

alex-l-kong: Yeah, confirmed that it won't work for anything greater than `0.16`. Problem is `skimage` discontinued support for `external` after `0.16`, but some of our code still relies on that.

requirements.txt
alex-l-kong: For now, I do not see a way. It would seem like a perk that only comes if you've integrated your project with PyPI. If anyone knows something about this, would really appreciate it!

requirements-test.txt
```diff
@@ -1,6 +1,6 @@
 pytest==5.4.0
 pytest-cov==2.5.1
 pytest-pycodestyle
-coveralls
+coveralls>=1.8.2,<2.0.0
```
ngreenwald, line 4: pytest can be <6.0

ngreenwald (changes requested):

ngreenwald (changes requested):
Can you also confirm that all of these are still needed?

alex-l-kong:
> Can you also confirm that all of these are still needed?

Good catch, we could remove a lot of the libraries in `requirements.txt`.

On a side note, can you configure the version of `six` on Travis to be `1.15.0`? I mentioned this in Slack, but right now, Travis defaults to `1.11.0`, and it's throwing errors. If we can't do anything on Travis' end, I think we should get this branch merged in ASAP so that the rest of the team can begin testing normally.

ngreenwald (approved):
Same thing here, we'll see if the `six` issue gets addressed by Will tomorrow, if not we can merge this

PR #254, opened 2020-10-01
Refactor verification logic into test_utils

### LINKED ISSUE #250: Condense shared verification logic into test_utils.py

We have a lot of shared logic used for verification. For example:

```
if not np.isin(included_fovs, all_data[fov_col]).all():
    raise ValueError(...)
```

It would be best to condense this into `test_utils.py` so that all it takes is a call to, say, `test_utils.verify_fovs(included_fovs, all_data[fov_col])`. There are various other examples found in our codebase that could use this treatment as well.

### LINKED ISSUE #264: Make error messages for mismatched columns, fovs, and markers for spatial analysis more descriptive

The exception messages we raise for checking `all_data` column, fov, and marker validity are very vague. For an inexperienced user with a very wide dataframe, it can be confusing to simply be told, for example, that "Fovs were not found in Expression Matrix" without also being told which ones were not found.

We should print out the exact mismatched values to help guide the user to adjust their csv/dataframe accordingly.

**Additional context**

We hadn't got too many complaints about this so far, but I ran into a lot of problems with this when given a sample dataset from Tyler. There were several column, fov, and marker mismatches and it was hard to pinpoint exactly which ones were missing without print debugging inside `spatial_analysis.py` (which also required restarting the kernel several times in `example_spatial_analysis_script.ipynb`).

While it is true that we're working on enforcing strict standardization guidelines for expression CSV's, I still think this is an issue worth looking into.

alex-l-kong:

Addresses and closes #250. Also addresses and closes #264. We have a lot of brief verification procedures that are annoying to repeat, especially in `analysis` (ex. verifying and saving to location, checking if `fovs` in `all_data`). We need to move these to `test_utils` so we don't duplicate logic.

As described, condense shared logic into `test_utils`.

**Remaining issues**

Let me know if there are more things we can move to `test_utils`.


ark/analysis/dimensionality_reduction.py
```diff
@@ -55,13 +57,7 @@ def plot_dim_reduced_data(component_one, component_two, fig_id, hue, cell_data,
...
 
     if save_dir is not None:
-        if not os.path.exists(save_dir):
-            raise ValueError("save_dir %s does not exist" % save_dir)
-
-        if save_file is None:
-            raise ValueError("save_dir specified but no save_file specified")
-
-        plt.savefig(os.path.join(save_dir, save_file))
+        test_utils.save_figure(save_dir, save_file)
```
ngreenwald, line 60: Nice, good idea, we do this all the time

ark/analysis/spatial_analysis.py
```diff
@@ -44,11 +45,7 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds,
     """
 
     # Setup input and parameters
-    if included_fovs is None:
-        included_fovs = list(set(all_data[fov_col]))
-        num_fovs = len(included_fovs)
-    else:
-        num_fovs = len(included_fovs)
+    included_fovs, num_fovs = test_utils.get_fov_info(all_data, fov_col, included_fovs)
```
ngreenwald, line 48: I think this is over-engineering. Especially given that we'd only use this inside this script, I think it adds clutter to have a function that does something this simple

ark/utils/test_utils.py
```diff
@@ -564,3 +565,78 @@ def create_test_extraction_data():
...
+    if included_fovs is None:
+        included_fovs = all_data[fov_col].unique()
+        num_fovs = len(included_fovs)
+    else:
+        num_fovs = len(included_fovs)
+
+    return included_fovs, num_fovs
+
+
+def verify_spatial_info(all_data, fov_col, included_fovs, excluded_colnames=None):
```
ngreenwald, line 618: The reason I described this in more general terms in my comment was because there are other places in the repo where we do this. 
For example [here](https://github.com/angelolab/ark-analysis/blob/cd51002e622c802fc44114ac6dd3431edf806682/ark/segmentation/marker_quantification.py#L259]). 

We can turn this into a function that checks if membership of one list with another, and reports errors according to whether all of list A have to be in list B, or only if parts of list A aren't in list B, depending on the context.

ngreenwald (changes requested):
Good start, some ways to further expand in the comments

ark/utils/test_utils.py
```diff
@@ -580,6 +581,52 @@ def create_test_extraction_data():
...
+            should be similar to the name of the variable associated with list_verify
+        name_values_bucket (str):
+            should be similar to the name of the variable associated with values_bucket
+
+    Raises:
+        ValueError:
+            if not all values in list_verify exist in values_bucket
+    """
+
+    if not np.isin(list_verify, values_bucket).all():
```
ngreenwald, line 623: Nice, this is much closer to what I had in mind. This will be a super useful, general function to give more descriptive error messages. 

Right now, it only supports checking for subsets of list 2 being present in list 1. Another use case would be adding checks for lists that should have all of the same items. 
For example, here: https://github.com/angelolab/ark-analysis/blob/526ea6d2e7d6446b618e52cc8fabed302b24fd9f/ark/segmentation/marker_quantification.py#L167

I think adding a `mode` flag that controls what type of check is being performed would accomplish this.

ark/utils/test_utils.py
```diff
@@ -580,6 +581,52 @@ def create_test_extraction_data():
...
+        bad_vals = [str(val) for val in list_verify if val not in values_bucket]
+
+        raise ValueError("%s value(s) %s not found in %s" %
```
ngreenwald, line 626: This error message could probably be a bit more verbose

ark/utils/test_utils.py
```diff
@@ -580,6 +581,52 @@ def create_test_extraction_data():
...
+        raise FileNotFoundError("save_dir %s does not exist" % save_dir)
+
+    # verify that if save_dir specified, save_file must also be specified
+    if save_file is None:
+        raise FileNotFoundError("save_dir specified but no save_file specified")
+
+    plt.savefig(os.path.join(save_dir, save_file))
+
+
+def verify_in_list(list_verify, values_bucket, name_list_verify, name_values_bucket):
```
ngreenwald, line 605: I think changing the variables to something like `test_list` and `good_values` would make this more clear. I also think it would be helpful to call these arguments by name when using the function. Otherwise, it's fairly hard to figure out what's going on when someone is reading through the code and just sees 4 different lists/names being passed to a verify function

ngreenwald (changes requested):
Much closer. One high level comment. These aren't actually test_utils, since they're not used for testing. They're more similar to `validate_paths`, which Adam put in io_utils. 

I'm open to suggestions of where their new home should be, but I don't think it should be test_utils, which is reserved for helper functions used during test functions

ark/utils/test_utils.py
alex-l-kong: My concern with this approach is that `verify_in_list` can get very confusing the more types of checks we add. In this case, for example, we're not really verifying whether something exists in another list, rather, it's more like if both lists exist inside each other. Even if we choose a different name for the function and the arg names to make things clearer (which I will be addressing), it can be really easy to forget which mode to set for this function, especially if we need to add more to it.

I do like the idea of having this check, but I think it's worth looking into making this into its own separate function if we expect set equality to pop up more often (same goes for other list-based verifications that pop up). I actually think checking if a value is a proper index into a list of keys (ex. val in dict.keys(), xarray.coords, DataFrame.columns or DataFrame.index) should also be its own separate function (we do that plenty throughout the repo). This would have the benefit of not only making it clear which function to call, but also prevent having too many mixed type arguments and allow for the customization of error messages without needing a ton of control statements.

It's probably not a big deal if we do choose to put everything into the same function, I just think it might make life a bit easier for us and for later contributors if we split things up a bit.

alex-l-kong:
> Much closer. One high level comment. These aren't actually test_utils, since they're not used for testing. They're more similar to `validate_paths`, which Adam put in io_utils.
> 
> I'm open to suggestions of where their new home should be, but I don't think it should be test_utils, which is reserved for helper functions used during test functions

I actually think this (or these, if we choose to split up the functions) can go in `misc_utils`. Doesn't really fall under any umbrella really, so I think it fits the mold.

ark/utils/test_utils.py
ngreenwald: Totally agree that checking if two lists are sets with the same items is a different functionality. Lets put that in a differently named function to make our life easier. 

I'm not sure what the difference is between the function that you currently have written and one that would check if an item is a valid index into a list of keys; isn't that just checking if it's present in the list?

alex-l-kong: @ngreenwald I can keep the valid index checking in `verify_in_list`.

ark/analysis/dimensionality_reduction.py
```diff
@@ -78,15 +74,15 @@ def visualize_dimensionality_reduction(cell_data, columns, category, color_map="
         color_map (str):
             Name of MatPlotLib ColorMap used, default is Spectral
         algorithm (str):
-            Name of dimensionality reduction algorithm, default is UMAP
+            Name of dimensionality reduction algorithm, must be UMAP, PCA, or tSNE, default is UMAP
```
ngreenwald, line 77: If the default is present in as a default in the definition, I don't think we need to redefine it in the docstring. If the default isn't immediately obvious (i.e., it results from None being assigned something), then it's worth having in the docstring

ark/analysis/dimensionality_reduction.py
```diff
@@ -78,15 +74,15 @@ def visualize_dimensionality_reduction(cell_data, columns, category, color_map="
...
             Directory to save plots, default is None
     """
+
     cell_data = cell_data.dropna()
+    dim_reduction_algos = ["UMAP", "PCA", "tSNE"]
 
-    if algorithm not in ["UMAP", "PCA", "tSNE"]:
-        raise ValueError(f"The algorithm specified must be one of the following: "
-                         f"{['UMAP', 'PCA', 'tSNE']}")
+    misc_utils.verify_in_list(algorithm=algorithm, dim_reduction_algos=dim_reduction_algos)
```
ngreenwald, line 85: I'm on the fence about this one. It's nice only having to provide 2 args instead of 4, especially when the extra args were basically just the string version of the variable. However, the downside is I think it's slightly more confusing to figure out what this function is doing, and it doesn't give us the flexibility to provide a different name for the variables.

ngreenwald: For example, if it was instead 
```
verify_in_list(values=algorithm, valid_values=dim_reduction_algos, values_names='algorithm', valid_names='dimensionality reduction algorithms')
```

It's longer and more annoying to write, but easier to read and slightly more descriptive error message. I'm fine with whatever you and @vacuousplanet think is better

ngreenwald (changes requested):
Overall really nice, it's clear how widely useful this function will be

ark/analysis/dimensionality_reduction.py
alex-l-kong: I'm also on the fence about this one. @vacuousplanet what do you think?

alex-l-kong: Yeah, I'm also on the fence. @ackagel what do you say?

ackagel: You could allow a custom error string to be passed in as a kwarg:
```
err_str = '%s doesn't accept the value: %s.  Valid dimensionality reduction algorithms are: %s'
misc_utils.verify_in_list(algorithm=algorithm, dim_reduction_algos=dim_reduction_algos, err_str=err_str)
```
Then, just check for provided err_str, and otherwise use the default, in `verify_in_list`.

One extra way you could go with this, is for a provided `err_str` to use all of the other provided kwargs in order.  The default `err_str` should stay tho; it looks like it works pretty well for most cases.

ackagel (commented):
looks mega clean

ark/analysis/dimensionality_reduction.py
ngreenwald: That made me realize something: we don't *have* to name the args the same as the variable. 

For example:
```
verify_in_list(algorithm=algorithm, dimesionality_reduction_algorithms=dim_reduction_algos)
```

This would work with Alex's current version, and allows us to basically make custom error messages, as long as we're okay with any new stuff being a single string. We could optionally split the arg_names based on _ before putting them into the error string for extra readability. 

Or, we could make it a dict, then we could have the keys be strings and values the appropriate values. But it wouldn't be ordered, so I think that would cause more of a headache

alex-l-kong: @ngreenwald I'm not a big fan of the dict method personally because of the reason you mentioned (unordered) and also because I think it's just a little bit more work than necessary (`**kwargs` basically accomplish the same thing and don't require dicts to be passed). As you mention, having flexible `**kwargs` also allows the variable names passed to be different.

@ackagel I do like the idea of custom error messages, but I think the original idea of this function was to eliminate the need of the user to format/print the error message on their end. The concern is that if we'll require a ton of custom error messages, it could defeat the purpose of having an easy-to-use function in `misc_utils` which allows a simple list in list/set equality check and an appropriate error message displayed if necessary.

There are definitely a lot of tradeoffs we'll have to make here, but unless you guys have extra concerns, I'll move forward with Noah's most recent suggestion of leaving the `**kwargs` as is and allowing the user to specify the arg names.

alex-l-kong:
I'm moving forward with Noah's suggestions for leaving the `**kwargs` as is, though I have changed a few of them to make them clearer and added underscore splitting as well.

ark/analysis/dimensionality_reduction.py
ngreenwald: I think we can remove "default is spectral" and "default is UMAP" from docstring, since they're both defined at the top

ark/analysis/spatial_analysis.py
```diff
@@ -58,12 +59,13 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds,
                              "eccentricity", "major_axis_length", "minor_axis_length",
                              "perimeter", "fov"]
 
-    # Error Checking
-    if not np.isin(excluded_colnames, all_data.columns).all():
-        raise ValueError("Column names were not found in Expression Matrix")
+    # check if included fovs found in fov_col
+    misc_utils.verify_in_list(included_fovs=included_fovs,
+                              all_data_fov_col_values=all_data[fov_col].unique())
```
ngreenwald, line 64: I think we could change the variable name be more interpretable, something like `fov_names`, `unique_fovs`, etc

ark/analysis/spatial_analysis.py
```diff
@@ -58,12 +59,13 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds,
...
 
-    if not np.isin(included_fovs, all_data[fov_col]).all():
-        raise ValueError("Fovs were not found in Expression Matrix")
+    # check if all excluded column names found in all_data
+    misc_utils.verify_in_list(excluded_colnames=excluded_colnames,
```
ngreenwald, line 67: Same here, something like `column_names`

ark/analysis/spatial_analysis.py
```diff
@@ -164,9 +167,9 @@ def calculate_cluster_spatial_enrichment(all_data, dist_matrices_dict, included_
 
     values = []
 
-    # Error Checking
-    if not np.isin(included_fovs, all_data[fov_col]).all():
-        raise ValueError("Fovs were not found in Expression Matrix")
+    # check if included fovs found in fov_col
+    misc_utils.verify_in_list(included_fovs=included_fovs,
+                              all_data_fov_col_values=all_data[fov_col].unique())
```
ngreenwald, line 172: Same

ark/utils/misc_utils.py
```diff
@@ -89,3 +92,86 @@ def combine_fov_directories(dir_path):
...
+    """
+
+    if len(kwargs) != 2:
+        raise ValueError("You must provide 2 arguments to verify_same_elements")
+
+    list_one, list_two = kwargs.values()
+
+    if not np.all(set(list_one) == set(list_two)):
+        bad_vals = ','.join(list(set(list_one) ^ set(list_two)))
+        list_one_name, list_two_name = kwargs.keys()
```
ngreenwald, line 174: I think you can add underscore splitting here too

ngreenwald (changes requested):
Overall looks good, I think this is the right approach. 

I left the same comment in a couple places, but I think it's worth modifying some of the variable names that we provide the function. The goal is for this error message to be super interpretable to a random user. The kind of cryptic variable shorthand we use (fov_col_all_vals) is not going to be sure useful. Making these more verbose and a bit simpler will ensure that when someone sees this error message, they don't still need to come bother us with a question.

ark/analysis/visualize.py
```diff
@@ -37,8 +38,8 @@ def draw_boxplot(cell_data, col_name, col_split=None, split_vals=None, save_dir=
             raise ValueError("If split_vals is set, then col_split must also be set")
 
         # all the values in split_vals must exist in the col_name of cell_data
-        if not all(val in cell_data[col_split].unique() for val in split_vals):
-            raise ValueError("Some values in split_vals not found in col_split column of data")
+        misc_utils.verify_in_list(split_vals=split_vals,
+                                  cell_data_col_split_values=cell_data[col_split].unique())
```
ngreenwald, line 42: This one too

ark/analysis/visualize.py
```diff
@@ -23,12 +25,11 @@ def draw_boxplot(cell_data, col_name, col_split=None, split_vals=None, save_dir=
...
 
     # the col_name must be valid
-    if col_name not in cell_data.columns.values:
-        raise ValueError("col_name specified does not exist in data provided")
+    misc_utils.verify_in_list(col_name=col_name, cell_data_columns=cell_data.columns.values)
 
     # if col_split is not None, it must exist as a column in cell_data
     if col_split is not None and col_split not in cell_data.columns.values:
-        raise ValueError("col_split specified does not exist in data provided")
+        misc_utils.verify_in_list(col_split=col_split, cell_data_columns=cell_data.columns.values)
```
ngreenwald, line 32: I think cell_data_columns or column_names are both fine, whichever you prefer

ark/segmentation/marker_quantification.py
```diff
@@ -161,8 +159,13 @@ def create_marker_count_matrices(segmentation_labels, image_data, nuclear_counts
         raise ValueError("Incorrect data type for image_data, expecting xarray")
 
     if nuclear_counts:
-        if 'nuclear' not in segmentation_labels.compartments:
-            raise ValueError("Nuclear counts set to True, but not nuclear mask provided")
+        misc_utils.verify_in_list(
+            nuclear_label='nuclear',
+            segmentation_labels_compartments=segmentation_labels.compartments.values
```
ngreenwald, line 164: `compartment_names` is probably more clear

ark/utils/spatial_analysis_utils.py
```diff
@@ -27,9 +29,8 @@ def calc_dist_matrix(label_maps, path=None):
 
     # Check that file path exists, if given
 
-    if path is not None:
-        if not os.path.exists(path):
-            raise FileNotFoundError("File path not valid")
+    if path is not None and not os.path.exists(path):
+        raise FileNotFoundError("File path not valid")
```
ngreenwald, line 33: `validate_path` here as well

ngreenwald (changes requested):
Just a few more examples where I think we can make the arg name more clear

ngreenwald (changes requested):
Looks good, once the marker_threshold PR is merged in you can update that here and we should be good to go

ark/analysis/spatial_analysis.py
```diff
@@ -58,21 +59,20 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds,
...
     # Subsets the expression matrix to only have channel columns
     all_channel_data = all_data.drop(excluded_colnames, axis=1)
 
-    # this will get refactored once the verification refactor PR gets merged in
-    if not np.all(set(marker_thresholds.iloc[:, 0]) == set(all_channel_data.columns)):
-        raise ValueError(
-            "The same markers must be found in marker thresholds and expression matrix columns"
-        )
+    # check that the markers are the same in marker_thresholdsa and all_channel_data
+    misc_utils.verify_same_elements(markers_to_threshold=marker_thresholds.iloc[:, 0].values,
```
ngreenwald, line 74: `markers_to_threshold` is a perfect name. `all_data_marker_columns` not as great. Maybe `all_markers`, 'marker_names`, `column_names`, etc

ngreenwald (changes requested):
Just a quick name change and should be good to go

ngreenwald (approved):

PR #268, opened 2020-10-06
Generate k means clustering count/mean value matrices by cell type and marker

### LINKED ISSUE #255: Generate k-means matrix for clustering

Now that we have a method for generating a method to run k-means cluster statistics, we should now implement the actual clustering results and create a matrix containing cell types, clusters, and mean marker expressions. This will address the data generating portion of the fourth point in #139.

alex-l-kong:

Addresses and closes #255. Now that we have a method to determine the optimal `k` for k-means clustering of a neighborhood matrix, we add functionality to generate count and mean value matrices by cell type and marker respectively for each cluster. This finishes up point 3 of #139.

We add a function to generate the actual k-means clusters and another to generate the corresponding count/mean value matrices.


ark/analysis/spatial_analysis.py
```diff
@@ -297,6 +297,95 @@ def create_neighborhood_matrix(all_data, dist_matrices_dict, included_fovs=None,
...
+    if included_fovs is None:
+        included_fovs = neighbor_mat[fov_col].unique()
+
+    if excluded_colnames is None:
+        excluded_colnames = ["cell_size", "Background", "HH3",
+                             "summed_channel", "label", "area",
+                             "eccentricity", "major_axis_length", "minor_axis_length",
+                             "perimeter", "fov"]
+
+    # make sure the specified
```
ngreenwald, line 350: Fix comment

ark/analysis/spatial_analysis.py
```diff
@@ -297,6 +297,95 @@ def create_neighborhood_matrix(all_data, dist_matrices_dict, included_fovs=None,
...
+
+    if cluster_num < 2:
+        raise ValueError("Invalid k provided for clustering")
+
+    # subset neighbor mat
+    neighbor_mat_data = neighbor_mat[neighbor_mat[fov_col].isin(included_fovs)]
+
+    # drop non-channel columns
+    col_to_drop = [fov_col, label_col]
+    neighbor_mat_data = neighbor_mat_data.drop(col_to_drop, axis=1)
```
ngreenwald, line 365: We'll need to do some error checking to make sure remaining columns are only numeric

ark/analysis/spatial_analysis.py
```diff
@@ -297,6 +297,95 @@ def create_neighborhood_matrix(all_data, dist_matrices_dict, included_fovs=None,
...
+
+    # generate cluster labels
+    cluster_labels = spatial_analysis_utils.generate_cluster_labels(
+        neighbor_mat_data, cluster_num)
+
+    # add labels to all_data
+    all_data[cluster_label_col] = cluster_labels
```
ngreenwald, line 372: Does this modify the all_data in place? can we return modified version instead?

ngreenwald (changes requested):
Looks good! Some more error checking and documentation mostly

ark/analysis/spatial_analysis.py
alex-l-kong: I'm not sure if this will work because `label_col` will also be a numeric type, but we don't want that in the clustering process. I think we should simply include a cell in the `Deepcell_Postprocessing` notebook that handles dropping these columns prior to passing them into `compute_cluster_metrics`. I will adjust the tests accordingly.

ark/analysis/spatial_analysis.py
```diff
@@ -297,6 +297,95 @@ def create_neighborhood_matrix(all_data, dist_matrices_dict, included_fovs=None,
...
+            the name of the column in all_data and neighbor_mat determining the fov
+        label_col (str):
+            the name of the column in all_data and neighbor_mat determining the label
+        cell_type_col (str):
+            the name of the column in all_data determining the cell type
+
+    Returns:
+        tuple (pandas.DataFrame, pandas.DataFrame):
+
+        - a matrix listing the number of cell types per cluster
```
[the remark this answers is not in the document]
ngreenwald: Yes, that's what Erin had in mind. My comment is just related to the actual docstring description. "A matrix listing the number of cell types per cluster" sounds like an `ax1` matrix, where `a` is the number of clusters, and entry has a single value representing the number of clusters. Your description here does a great job of explaining, I would use that instead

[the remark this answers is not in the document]
alex-l-kong: Ah got it.

ackagel (commented):
Looks great!

ngreenwald (changes requested):
Looks good, just a couple questions about naming/descriptions

ark/analysis/spatial_analysis.py
```diff
@@ -297,21 +297,115 @@ def create_neighborhood_matrix(all_data, dist_matrices_dict, included_fovs=None,
...
+    """Generate the cluster info on all_data using k-means clustering on neighbor_mat.
+
+    cluster_num has to be picked based on visualizations from compute_cluster_metrics.
+
+    Args:
+        all_data (pandas.DataFrame):
+            data including fovs, cell labels, and cell expression matrix for all markers
+        neighbor_mat (pandas.DataFrame):
+            a neighborhood matrix, created from create_neighborhood_matrix
+            the matrix should have the label col dropped
```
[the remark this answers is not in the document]
alex-l-kong: Oh right, forgot to remove that from the docstring.

ngreenwald (approved):
Looks good, lets see if Adam has any feedback before merging

ackagel (approved):
Looks great!  Once this is merged, I'll need to revise some lines for the global-settings pr, but it doesn't look like that'll be difficult.

Issue #242, opened 2020-09-27
Run setup in .travis.yml to properly install dependencies of ark-analysis on pip install


alex-l-kong:
**Describe the bug**

Our current library on PyPI does not install the required dependencies on `pip install ark-analysis`. When a user runs this command, they will only get `ark-analysis`, but none of the other dependencies will be installed.

To fix, we need to add a command to run `travis_retry python setup.py test` at the very least. Most likely, we'll be piggybacking off of VVL again.

This issue may require multiple PRs to fix. Currently, the only way we can see the final result is on a merge into master with a tagged commit and a new release.

Not a high priority issue right now, I would hold off on it until someone actually complains (don't expect too many people to be `pip` installing our library now).

alex-l-kong:
This issue has been resolved: `pip` installs all the dependencies properly when we hard code the list in.

PR #273, opened 2020-10-14
Bump version number and ensure requirements.txt are installed properly on pip install

alex-l-kong:

Addresses #242. This PR may require more work after it is merged in because we can only see the final results after a merge into master.

Now that Omer has completely integrated deepcell with ark-analysis, we need to bump the version number. However, another issue needs to be addressed: running `pip install ark-analysis` will only install `ark-analysis`, but not any of the other dependencies.

We use `deepcell-toolbox` as a guiding point for our changes.

**Remaining issues**

It seems like `deepcell-toolbox` specifically implemented some of their changes because they need to install `deepcell` as well. I'm not sure how much is actually applicable to us, it may potentially not work. If so, we'll need to open another PR to address the issues that come up.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from os import path
 from distutils.command.build_ext import build_ext as DistUtilsBuildExt
 from setuptools import setup, find_packages
 
-VERSION = '0.3.6'
+VERSION = '0.2.7'
 VERSION_MIBILIB = '1.3.0'
 
 
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ from distutils.command.build_ext import build_ext as DistUtilsBuildExt
 from setuptools import setup, find_packages
 
 VERSION = '0.3.6'
+VERSION_MIBILIB = '1.3.0'
 
 
 # define a parsing function for requirements.txt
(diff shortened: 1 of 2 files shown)


alex-l-kong:
So I literally just copied most everything that Will did on `deepcell-toolbox`, which I confirmed does install dependencies correctly. If this still doesn't fix our `pip install` issue with dependencies, then I think we need to ask Will for some extra help.

ngreenwald:
Did you take this from `deeepcell-tf` or `deepcell-toolbox`? You mentioned this could be due to a requirements issue, but the toolbox doesn't have `deepcell-tf` as a requirement

alex-l-kong:
> Did you take this from `deeepcell-tf` or `deepcell-toolbox`? You mentioned this could be due to a requirements issue, but the toolbox doesn't have `deepcell-tf` as a requirement

I took this from `deepcell-toolbox`. But now that you mention it, perhaps I should've been looking at `kiosk-client`. Without any complicated `setup.py` jazz, `pip` installed it fine: the requirements all downloaded properly. I am curious about the extra Docker commands in `.travis.yml` are doing though. Wonder if they help install the requirements in `requirements.txt`.

ngreenwald:
Maybe ask Will? If we can keep things simpler that would be better

alex-l-kong:
> Maybe ask Will? If we can keep things simpler that would be better

Ah OK, now I know why I used `deepcell-toolbox` as a reference. Will told me that its `setup.py` was configured properly, as well as the `.travis.yml` file which explicitly ran it. I decided to copy the `BuildExtension` class as well because `deepcell-toolbox` does that in their `setup.py` as well.

I'll ask Will what that function does, and if he says its purely for building `deepcell` on their end, then I think we can safely remove it and hope that we only need to run `setup.py` in Travis for it to properly build the extensions we need.

alex-l-kong:
Just heard back from Will. He said that the `BuildExtension` class was only if you needed to compile C files for `cython`. This is required for `deepcell-toolbox` because they actually need this, but we don't. So we're in the clear to remove it.

The key is to have the `travis_retry python setup.py` in `.travis.yml` so Travis actually runs `setup.py` (which includes the dependencies) on their end before deploying to PyPI.

ngreenwald (changes requested):
Sounds good, let me know once it's ready

setup.py
```diff
@@ -1,7 +1,10 @@
+import setuptools
 from os import path
+from distutils.command.build_ext import build_ext as DistUtilsBuildExt
 from setuptools import setup, find_packages
 
-VERSION = '0.2.6'
+VERSION = '0.3.6'
```
ngreenwald, line 6: If we're bumping this, it should 0.2.7 right?

ngreenwald (changes requested):

setup.py
alex-l-kong: Ah got it, will fix.

ngreenwald (approved):

PR #274, opened 2020-10-15
Fix marker ordering errors and verification for channel enrichment

### LINKED ISSUE #265: Marker threshold matrix validation does not throw proper exception if markerThresholds.csv has extra markers
**Describe the bug**

Many times, the user will specify `markerThresholds.csv` with more markers than are actually used for analysis. The current process for validating markers cannot handle this case properly, erroring when computing `not (list(marker_thresholds.iloc[:, 0]) == channel_titles).any()` before getting to `raise ValueError`. The only way to fix this currently is to require the user to remove the offending markers from `markerThresholds.csv`.

For the record, this bug was also encountered from running on Tyler's data.

**To Reproduce**

See above.

**Expected behavior**

We should be raising proper exceptions for this type of error at the very least (using `np.isin`, not `==`), but I would strongly argue that we should not raise an exception in this case. The reason is because we don't raise an exception if the user specifies a `dist_matrices_dict` with more point values than found in the `fov_col` of `all_data`. To keep things simple and consistent, I think we should replicate a similar methodology for validating markers.

### LINKED ISSUE #266: compute_close_cell_num assumes same indexing for thresh_vec and current_fov_channel_data for channel-based analysis
**Describe the bug**

First encountered with Tyler's data. Related to #265 but requires much more attention.

Our current marker validation scheme in `calculate_channel_spatial_enrichment` completely fails to check if the order of `marker_thresholds.iloc[:, 0]` and `all_channel_data.columns` are the same (the latter is generated by subsetting only the marker columns of the expression matrix). This is bad because in `compute_close_cell_num`, it is assumed for channel-based analysis that `thresh_vec` (analogous with `marker_thresholds.iloc[:, 0]`) and the columns for `current_fov_channel_data` (subsetting the expression matrix for only the desired fov) are indexed the same. If they are in actuality not, our thresholds would be completely wrong and could potentially produce extremely misleading results.

**To Reproduce**

Specify the markers in `markerThresholds.csv` in a different order than in `example_expression_matrix.csv`.

**Expected behavior**

Obviously, we should be thresholding properly, I think this should be done regardless of the order the markers are specified in `markerThresholds.csv` and `example_expression_matrix.csv` (or whatever the user names them).

We should change our marker validation process to only check `isin` without checking order, and change the `current_marker` arg in the call to `get_pos_cell_labels_channel` accordingly.

**Screenshots**

Screen Shot 2020-10-05 at 11 17 58 PM

Tyler's example data exhibits this exact same problem. I printed out the markers in `markerThresholds.csv` first, then `all_channel_data`. As you can see, the order in which the columns appear is not the same between the two. And explicitly printing out `list(marker_thresholds.iloc[:, 0]) == channel_titles` (which is used in our current logic for marker validation, sorry printed it twice) reveals problems. However, we do not actually raise an exception in this case.

I got fortunate in that Tyler specified the same threshold for each marker, but this is definitely not to be assumed, and it would have absolutely failed if Tyler specified different thresholds for each marker.

alex-l-kong:

Addresses and closes #265. Also addresses and closes #266. We add a more robust method to verify that the markers in `all_channel_data` and `marker_thresholds` are the same between the two, and we ensure that they are ordered consistently as well. This only applies for channel enrichment.

Use set equality to check if the markers are the same, and reindexing to enforce order.

**Remaining issues**

This is an extremely strict way of handling the various different types of `marker_thresholds` and `all_channel_data` values the user may pass in. From personal experience, expect all sorts of messes in this regard. This method may be a bit too strict, so I'm open to other ideas.

### DIFF
--- a/ark/analysis/spatial_analysis.py
+++ b/ark/analysis/spatial_analysis.py
@@ -69,9 +69,6 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds, 
     all_channel_data = all_data.drop(excluded_colnames, axis=1)
 
     # this will get refactored once the verification refactor PR gets merged in
-    # TODO: this is the really easy way to to this, but is it worth enforcing such strict
-    # guidelines (every marker present in marker_thresholds must be found in channel_titles
-    # and vice versa)
     if not np.all(set(marker_thresholds.iloc[:, 0]) == set(all_channel_data.columns)):
         raise ValueError(
             "The same markers must be found in marker thresholds and expression matrix columns"
--- a/ark/analysis/spatial_analysis.py
+++ b/ark/analysis/spatial_analysis.py
@@ -67,14 +67,22 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds, 
 
     # Subsets the expression matrix to only have channel columns
     all_channel_data = all_data.drop(excluded_colnames, axis=1)
+
+    # this will get refactored once the verification refactor PR gets merged in
+    # TODO: this is the really easy way to to this, but is it worth enforcing such strict
+    # guidelines (every marker present in marker_thresholds must be found in channel_titles
+    # and vice versa)
+    if not np.all(set(marker_thresholds.iloc[:, 0]) == set(all_channel_data.columns)):
+        raise ValueError(
+            "The same markers must be found in marker thresholds and expression matrix columns"
+        )
+
+    # reorder all_channel_data's marker columns the same as they appear in marker_thresholds
+    all_channel_data = all_channel_data[marker_thresholds.iloc[:, 0].values]
     # List of all channels
     channel_titles = all_channel_data.columns
     # Length of channels list
     channel_num = len(channel_titles)
-
-    # Check to see if order of channel thresholds is same as in expression matrix
-    if not (list(marker_thresholds.iloc[:, 0]) == channel_titles).any():
-        raise ValueError("Threshold Markers do not match markers in Expression Matrix")
 
     # Create stats Xarray with the dimensions (fovs, stats variables, num_channels, num_channels)
     stats_raw_data = np.zeros((num_fovs, 7, channel_num, channel_num))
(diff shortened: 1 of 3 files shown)


ark/utils/test_utils.py
```diff
@@ -640,7 +640,11 @@ def _make_threshold_mat(in_utils):
     thresh.iloc[:, 1] = .5
 
     if not in_utils:
-        thresh.iloc[:, 0] = np.arange(20) + 2
+        # print(len(np.concatenate([np.arange(2, 14), np.arange(15, 23)])))
```
ngreenwald, line 643: So this function currently isn't used for anything, but will be in the future?

ngreenwald (changes requested):

ark/utils/test_utils.py
alex-l-kong: Forgot to remove the print statement, sorry.

If you're talking about `_make_threshold_mat` as a whole, it is used in `spatial_analysis_test.py` and `spatial_analysis_utils_test.py`.

ark/analysis/spatial_analysis.py
```diff
@@ -67,15 +67,20 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds,
 
     # Subsets the expression matrix to only have channel columns
     all_channel_data = all_data.drop(excluded_colnames, axis=1)
+
+    # this will get refactored once the verification refactor PR gets merged in
+    if not np.all(set(marker_thresholds.iloc[:, 0]) == set(all_channel_data.columns)):
```
ngreenwald, line 72: Lets add a brief error check in spatial_analysis_test for this as well

ngreenwald (changes requested):
Ah right, got confused by the comment, I see now. Just a one-liner check for new test and we should be good

ark/analysis/spatial_analysis.py
alex-l-kong: Cool, I basically took the last test in for `calculate_channel_spatial_enrichment` and added a little bit on to that.

ackagel (approved):
`marker_thresholds` is growing up so fast 😭 ; so much tougher and robust than they once were

ngreenwald (approved):

PR #275, opened 2020-10-15
Add empty list validation for mibitiff loading

### LINKED ISSUE #271: Error checking for mibi-tiff loading
I think we're missing some error checking for the load_mibitiffs function. Specifically, as noted in #270, we're getting too deep into the stack trace with what appears to be an empty list of filenames

Earlier checking to ensure that a valid list has been supplied. Depending on what the specific issue is, this could either be a one-line fix to ensure that `list is not None`, an additional validation function for `misc_utils`, or a call to validate_urls

alex-l-kong:

Addresses and closes #271. Someone recently encountered a bug in `load_imgs_from_mibitiff` in which an empty list of `mibitiff_files` (which subsequently leads to an empty list of `fovs`) caused the loading process to freak out. We need to properly address this so the error message is clear.

In addition to adding a basic check for an empty `mibitiff_files` list in the function, also add `data_dir` validation where needed.

### DIFF
--- a/ark/utils/load_utils.py
+++ b/ark/utils/load_utils.py
@@ -35,8 +35,14 @@ def load_imgs_from_mibitiff(data_dir, mibitiff_files=None, channels=None, delimi
             xarray with shape [fovs, x_dim, y_dim, channels]
     """
 
+    if not os.path.isdir(data_dir):
+        raise ValueError(f"Invalid value for data_dir. {data_dir} is not a directory.")
+
     if not mibitiff_files:
         mibitiff_files = iou.list_files(data_dir, substrs=['.tif'])
+
+    if len(mibitiff_files) == 0:
+        raise ValueError("No mibitiff files specified in the data directory %s" % data_dir)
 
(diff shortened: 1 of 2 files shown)


alex-l-kong:
I verified that not sorting by channel has no effect on the `Segment_Image_Data` notebook nor the other functions in `ark`. I'll open a PR to address this.

ark/utils/load_utils.py
```diff
@@ -35,9 +35,15 @@ def load_imgs_from_mibitiff(data_dir, mibitiff_files=None, channels=None, delimi
             xarray with shape [fovs, x_dim, y_dim, channels]
     """
 
+    if not os.path.isdir(data_dir):
```
ngreenwald, line 38: We should use validate_path here instead, since it produces a more verbose error message.

ngreenwald (changes requested):
Looks good! We can switch all of the directory checks to use the validate_paths function, since it identifies where the typo is

ngreenwald (approved):
Looks good

PR #277, opened 2020-10-15
Remove double sort by channel from load_imgs_by_tree

### LINKED ISSUE #262: load_imgs_from_tree always sorts channels
https://github.com/angelolab/ark-analysis/blob/eaf429e6efa25e29babf463c89d6a3de52410a3f/ark/utils/load_utils.py#L247

The current implementation now always sorts fovs and channels. These means it's not possible to specify the order of channels in the xarray by passing in a list, since they'll get resorted anyway. I think before we had it set up s.t. sorting only happened if None was provided. Is there a reason that we need to sort in all cases?

alex-l-kong:

Addresses and closes #262.

We need to remove the extra `.sorted('channels')` at the end of `load_imgs_by_tree`. This means the testing function will also need to be slightly altered.


ark/utils/load_utils.py
```diff
@@ -172,7 +172,7 @@ def load_imgs_from_tree(data_dir, img_sub_folder=None, fovs=None, channels=None,
                           dims=["fovs", "rows", "cols", "channels"])
 
     # sort by fovs and channels for deterministic result
-    img_xr = img_xr.sortby('fovs').sortby('channels')
+    img_xr = img_xr.sortby('fovs')
```
ngreenwald, line 175: Same thing for FOVs: I think we should sort once we list all of the FOVs in the folder if None is provided. If a custom FOV list is provided, then we shouldn't sort. If order matters for some reason, I think it makes sense to keep it in the order that the items were provided in the function call.

ark/utils/load_utils_test.py
```diff
@@ -96,15 +96,22 @@ def test_load_imgs_from_tree():
             load_utils.load_imgs_from_tree(temp_dir, img_sub_folder="TIFs", dtype="int16",
                                            channels=some_chans)
 
-        assert loaded_xr.equals(data_xr[:, :, :, :2], )
+        # need to reindex columns first before asserting equality
+        # cannot use xarray.equals for this because reindexing string coords changes their dtype
+        assert np.all(loaded_xr.values == data_xr[:, :, :, :2].loc[
```
ngreenwald, line 101: Why is this change necessary? We were sorting before and we're sorting now, so what's the difference?

ngreenwald (changes requested):
Nice start, see comments below

ark/utils/load_utils_test.py
alex-l-kong: So the problem seems to happen when `channels` is unordered (ex. `['chan1', 'chan0']`). This can either happen because the argument to `channels` itself is unordered or becomes unordered when not all of the channels have a proper extension (in which `iou.list_files` is run, and that potentially can return an unordered list as well).

In this case, the `channel` coords in `loaded_xr` will likely be ordered differently than it would be in the test data (`data_xr`) created by `create_paired_xarray_fovs`. If we want to test that the corresponding values are the same, we need to reindex one of these arrays prior to testing equality because `.equals` in `xarray` will complain that the `channel` coords do not match between the two xarrays.

ark/utils/load_utils.py
ackagel: I think @ngreenwald was suggesting to remove this line. fovs and channels are both already sorted if they are `None`, and this line always sorts by fovs, even with user specified order.

ackagel (commented):
Not much to add.  Any reason those big binary files got added in?

`git add *` is a gift and a curse

ark/utils/load_utils.py
alex-l-kong: Yeah, forgot to do that. Will do for the next commit.

alex-l-kong:
> Not much to add. Any reason those big binary files got added in?
> 
> `git add *` is a gift and a curse

Screwed up using Sublime Merge. Will remove for the next commit.

alex-l-kong:
Oh whoops, sorry premature review request here. @ngreenwald perhaps we could use this to discuss the comment I left above about why we need to reindex `data_xr`.

ark/utils/load_utils_test.py
ngreenwald: Okay cool. In that case, maybe it would be worthwhile to add a few lines around 134 so that the channel names, with extensions, are in the same order as the supplied channels. I think it would be good for the channels, and fovs, to appear in the xarray in the order the user specified them.

alex-l-kong:
Ok, this PR is finally ready for a real review!

ark/utils/load_utils_test.py
```diff
@@ -131,6 +131,8 @@ def test_load_imgs_from_tree():
 
         assert loaded_xr.equals(data_xr)
 
+        # assert loaded_xr.equals(data_xr.loc[:, :, :, loaded_xr.coords['channels'].values])
```
ngreenwald, line 134: What's going on with this guy

ngreenwald (changes requested):
Awesome. Can you doublecheck if any of the other data loaders do the same thing?

alex-l-kong:
> Awesome. Can you doublecheck if any of the other data loaders do the same thing?

Yeah for sure. Looks like both `load_imgs_from_mibitiff` and `load_imgs_from_dir` also use an extra sort at the end. Time to nuke that.

ngreenwald (approved):

PR #281, opened 2020-10-16
Remove .tif/.tiff extension from list_files for MIBItiff reading in Segment_Image_Data.ipynb

### LINKED ISSUE #278: create_deepcell_output in Segment Image Data notebook generates error on multichannel .tiffs
The "create_deepcell_output" command in the Segment Image Data notebook produces an error. I have multiple MIBI multi channel .tif files in my deepcell_input folder, but once I get to the upload step, it seems to be looking for files ending in ".tiff.tif" files instead.

**To Reproduce**
1. In the Segment_Image_data.ipynb file, set MIBItiff = True and point the tiffdir variable to a folder with multichannel .tiffs 
2. Run the notebook through all of the preprocessing steps (this will create .tif files in the deepcell_input folder, as expected)
3. Run the line with "create_deepcell_output"

**Expected behavior**

The program should find the .tif files in the deepcell_input folder, zip and upload them.

**Error trace**

```
---------------------------------------------------------------------------
<ipython-input-13-2b23a959c361> in <module>
----> 1 deepcell_service_utils.create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=fovs)

/usr/local/lib/python3.6/site-packages/ark/utils/deepcell_service_utils.py in create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs, suffix, host, job_type)
     52             if not os.path.isfile(filename):
     53                 raise ValueError('Could not find .tif file for %s. '
---> 54                                  'Invalid value for %s' % (fov, filename))
     55             zipObj.write(filename, os.path.basename(filename))
     56 

ValueError: Could not find .tif file for fov6_LichenoidDysplasia-MassCorrected-Filtered.tiff. Invalid value for ../data/HNDysplasia/input_data/deepcell_input/fov6_LichenoidDysplasia-MassCorrected-Filtered.tiff.tif
```

### LINKED ISSUE #279: Generating fov names with MIBItiff keeps .tiff extension in Segment_Image_Data
**Describe the bug**

When `MIBItiff` is set to`True` in `Segment_Image_Data.ipynb`, the `fov` names that are generated from `list_files` does not properly remove the `.tiff` extension afterward. This is a problem because `generate_deepcell_output` assumes that the `fovs` provided are just the names of the `fovs` and not corresponding file names.

**To Reproduce**

Run `Segment_Image_Data.ipynb` with example data such that `MIBItiff = True`.

**Expected behavior**

We should either be removing the `.tiff` extension in `Segment_Image_Data` or in `generate_deepcell_output`, whichever one works better for us.

alex-l-kong:

A mini-PR that addresses and closes #279. By extension, addresses and closes #278.

We simply add a call to `extract_delimited_names` in `io_utils` to fix this.


alex-l-kong:
We still may need to address other locations where we need to remove extensions.

ngreenwald (changes requested):
As I mentioned in my comment on the issue, this will fail because FOVs needs to have the .tiff extension for loading mibi-tiff files. 
You should change the mibi-tiff flag to be true while making these changes to make sure we fix it

ngreenwald (changes requested):
Rather than adding more stuff to the notebook, we could probably put this in the deepcell function itself. For example, this exact cleanup already happens if FOVs are set to None:

https://github.com/angelolab/ark-analysis/blob/b7d7314b9043c82a33d410352e6566c17e9df244/ark/utils/deepcell_service_utils.py#L43

alex-l-kong:
> Rather than adding more stuff to the notebook, we could probably put this in the deepcell function itself. For example, this exact cleanup already happens if FOVs are set to None:
> 
> https://github.com/angelolab/ark-analysis/blob/b7d7314b9043c82a33d410352e6566c17e9df244/ark/utils/deepcell_service_utils.py#L43

Oh wow, totally forgot about that. Let's do that instead.

alex-l-kong:
I'm not a huge fan of the logic I'm using to handle the removal of the `.tif` extension and the verification of all the `fovs` in `deepcell_input_dir`, so would definitely appreciate a review on that.

ngreenwald (changes requested):
It's a bit messy, but I think it gets the job done. I think 2 things would help:
1) A few comments describing what's going on
2) A test case for input names with/without extension to make sure it handles properly

ark/utils/deepcell_service_utils_test.py
```diff
@@ -31,6 +31,12 @@ def test_create_deepcell_output(mocker):
         pathlib.Path(os.path.join(input_dir, 'fov1.tif')).touch()
         pathlib.Path(os.path.join(input_dir, 'fov2.tif')).touch()
 
+        with pytest.raises(ValueError):
+            # fail if non-existant fovs are specified
```
ngreenwald, line 35:
```suggestion
            # fail if non-existent fovs are specified
```

ark/utils/deepcell_service_utils_test.py
```diff
@@ -41,9 +47,30 @@ def test_create_deepcell_output(mocker):
         assert os.path.exists(os.path.join(output_dir, 'fov1_feature_0.tif'))
         assert os.path.exists(os.path.join(output_dir, 'fov2_feature_0.tif'))
 
-        # if fovs is None, all .tif files in input dir should be taken
+        # /fovs.zip file should be removed from input folder after completion
```
ngreenwald, line 50:
```suggestion
        # fovs.zip file should be removed from input folder after completion
```

ngreenwald (changes requested):
A few typos. We'll see if @vacuousplanet has any additional comments about how to handle the file suffixes

ark/utils/deepcell_service_utils_test.py
alex-l-kong: Aw man, gotta take spelling lessons again?

ark/utils/deepcell_service_utils.py
```diff
@@ -37,25 +39,39 @@ def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
...
+        fovs = input_files
+
+    # now extract only the names of the fovs without the file extension
+    fovs = io_utils.extract_delimited_names(fovs, delimiter='.')
 
+    # make sure that all fovs actually exist in the list of input_files
+    misc_utils.verify_in_list(
+        fovs=fovs,
+        deepcell_input_files=[in_file.replace('.tif', '') for in_file in input_files])
+
```
ackagel, line 58: I think you can just do:
```
in_fovs = io_utils.extract_delimited_names(tifs, delimiter='.')
...
misc_utils.verify_in_list(
    fovs=fovs,
    deepcell_input_files=in_fovs
)
```
Regardless, currently, if someone has `fov1.tiff`, the `replace` line will turn that into `fov1f` not `fov1`.

ackagel (changes requested):
So, the messiness of this code depends on how tolerant of user input we are.  If we mandate fovs have their file extensions (or mandate they don't have them), then the code is much less messy.  However, if we keep fovs file-extension agnostic, this 'clean-up' code it about as optimized as it can be imo.

I'm in favor of keeping file-extension agnosticism, in the hopes of a cleaner issues page 😉 .  

Anyways, just one fix and I think we're gtg here

ackagel (approved):
👍

ark/utils/deepcell_service_utils.py
```diff
@@ -37,25 +39,39 @@ def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
...
             filename = os.path.join(deepcell_input_dir, fov + '.tif')
-            if not os.path.isfile(filename):
-                raise ValueError('Could not find .tif file for %s. '
-                                 'Invalid value for %s' % (fov, filename))
             zipObj.write(filename, os.path.basename(filename))
 
-    print('Uploading files to DeppCell server.')
+    # pass the zip file to deepcell.org, then remove it afterwards
+    print('Uploading files to DeepCell server.')
+
```
ngreenwald, line 74: Rather than removing this zip file (line 76), I think we should check if it already exists when creating the zip file above, and then give a warning that it's being overwritten. given the current issues with zip file sizes, and purely for debugging purposes, I think it's useful to keep this zip file around.

ark/utils/deepcell_service_utils_test.py
```diff
@@ -41,14 +47,35 @@ def test_create_deepcell_output(mocker):
...
 
-        # if fovs is None, all .tif files in input dir should be taken
+        # fovs.zip file should be removed from input folder after completion
+        assert not os.path.isfile(os.path.join(input_dir, 'fovs.zip'))
+
+        os.remove(os.path.join(output_dir, 'fov1_feature_0.tif'))
+        os.remove(os.path.join(output_dir, 'fov2_feature_0.tif'))
+
+        # test with mixed fov/file list
+        create_deepcell_output(deepcell_input_dir=input_dir, deepcell_output_dir=output_dir,
```
ngreenwald, line 57: Can you add one image that has .tiff as well? We want to catch the error Adam pointed out

ngreenwald (changes requested):
Look good, two small tweaks

ark/utils/deepcell_service_utils.py
alex-l-kong: We already have this warning, so I'll just remove the line that removes the zip file.

ngreenwald (approved):

PR #288, opened 2020-10-23
Add status updates to deepcell_service_utils

### LINKED ISSUE #283: Add status updates to deepcell_service_utils
We have built-in logging already for the signal_extraction portion of the segmentation pipeline. However, we don't have this for the code that sends the data to deepcell.org

We should add logging that describes that stage of the pipeline is currently running. For example, creating zip file, sending to server, downloading from server, etc.

omerbt:

To add logging for `deepcell_service_utils` to describe the current stage of the pipeline. closes #283 

Added logging when: creating the zip file, calling kiosk-client, and extracting the tif files from the output of deepcell. 

**Remaining issues**

kiosk-client writes its own logs (using Python `logging` module). Maybe it is worth thinking if/how to integrate with it so we can show these logs in the pipeline (for example, currently there will be no indication in our pipeline if there was a problem when uploading the files to deepcell server).

### DIFF
--- a/ark/utils/deepcell_service_utils.py
+++ b/ark/utils/deepcell_service_utils.py
@@ -46,6 +46,7 @@ def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
     if os.path.isfile(zip_path):
         warnings.warn(f'{zip_path} will be overwritten.')
 
+    print('Zipping preprocessed tif files.')
     with ZipFile(zip_path, 'w') as zipObj:
         for fov in fovs:
             filename = os.path.join(deepcell_input_dir, fov + '.tif')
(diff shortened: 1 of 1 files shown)


ngreenwald (changes requested):
That's a good point about passing errors from the deepcell server. I've written a test so that it will fail if you provide images of the wrong shape. Can you try sending an image that's the wrong shape (i.e. 512x512x4 or something) to the server and see what that looks like in the jupyter notebook? If it shows up then we should be fine, but if it doesn't then we might need to investigate what the best way is to propagate that error message to the notebook.

omerbt:
We get a UserWarning (that is generated in `create_deepcell_output`) saying that the tif file was not found for that fov.
('Deep Cell output file was not found for {fov}.' )

ngreenwald (changes requested):
Can you ask Will if there's a way for us to display the ValueError that is getting raised on the server?

That's what is preventing the FOV from being generated in the first place, so it would be nice if we could just show whatever internal error is being raised by the kiosk.

ngreenwald (approved):

PR #291, opened 2020-10-23
Fix Segment_Image_Data plot overlay

### LINKED ISSUE #284: Error in Segment Image Data notebok when making overlay using > 3 total nuclear and membrane channels
**Describe the bug**
Using more than 3 total nuclear and membrane channels for segmentation produces an error in the overlay visualization step in Segment_Image_Data notebook.

**To Reproduce**
1. Set mibitiff = true, though I don't know if this is necessary to reproduce
2. I added 2 nuclear channels and 3 membrane channels
3. Run notebook until cell 9 (titled "We can also then save the segmented mask overlaid on the imaging data")
4. Try to generate overlays
5. See error below

**Expected behavior**
Expect to get an overlay with the proper addition of the nuclear and membrane channels , as referenced in #272 .

I can get around this error by manually feeding the plot_overlay function just a selection of the nuclear and membrane channels, as long as I just feed it 3 or les.

**Error Trace**
```
<ipython-input-9-f71583686a6c> in <module>
      2     plot_utils.plot_overlay(segmentation_labels.loc[fov, :, :, "whole_cell"].values,
      3                             data_xr.loc[fov, :, :, :].values,
----> 4                             path=os.path.join(deepcell_output_dir, f'{fov.values}_overlay.tif'))

/usr/local/lib/python3.6/site-packages/ark/utils/plot_utils.py in plot_overlay(predicted_contour, plotting_tif, alternate_contour, path)
     56             else:
     57                 raise ValueError("only 3 channels of overlay supported, got {}".
---> 58                                  format(plotting_tif.shape))
     59         else:
     60             raise ValueError("plotting tif must be 2D or 3D array, got {}".

ValueError: only 3 channels of overlay supported, got (2048, 2048, 5)

```

### LINKED ISSUE #285: Save combined nuclear and membrane channels for visualization
As described in #284, our current visualization fails when multiple markers have been summed together to create the input for deepcell. This is because the `data_xr`, which is used to load the markers in cell 5, doesn't contain the summed channels. Instead, this summation happens within the `generate_deepcell_input` function, and the summed markers are saved directly to disk as .tiff files

We should store the summed versions of these markers for easy visualization. There are two ways to accomplish this. 
- The first would be to separately load the tiffs which are created by `generate_deepcell_input` into a different data_structure, and use that for visualization. For example, `load_utils.load_imgs_from_dir` should work directly on those .tiff files. 
- The second would be to break `generate_deepcell_input` into two functions. One function would create (and return) the summed xarray. The next function would take that xarray and save it to disk as individual tiffs.

alex-l-kong:

Addresses and closes #285. By extension, addresses and closes #284. When more than 3 channels are specified, plot overlaying will fail. This is because we're trying to visualize the un-summed arrays, but we need to visualize the summed arrays across nuclear and membrane channels.

`generate_deepcell_input` already handles the summing, so we just need to read that data from the directory `deepcell_input` and pass that to our overlay instead.

### DIFF
--- a/templates/Segment_Image_Data.ipynb
+++ b/templates/Segment_Image_Data.ipynb
@@ -9,7 +9,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 2,
    "metadata": {
     "tags": []
    },
@@ -19,10 +19,6 @@
     "import os\n",
     "import skimage.io as io\n",
     "import matplotlib.pyplot as plt\n",
-    "\n",
-    "import sys\n",
-    "sys.path.append('..')\n",
-    "sys.path.append('../ark')\n",
     "\n",
     "from ark.utils import data_utils, load_utils, io_utils, plot_utils, deepcell_service_utils\n",
     "from ark.segmentation import marker_quantification"
@@ -37,7 +33,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -64,7 +60,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 4,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -100,7 +96,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 5,
    "metadata": {
     "tags": []
    },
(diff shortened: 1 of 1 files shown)


ngreenwald (changes requested):
How come you have `force_ints` = True?

alex-l-kong:
> How come you have `force_ints` = True?

Oh oops, let me remove that.

ngreenwald (approved):

PR #292, opened 2020-10-23
ignore low_contrast warning

### LINKED ISSUE #260: low contrast image warning
Getting this warning message when loading images into notebook:

/usr/local/lib/python3.6/site-packages/ark/utils/data_utils.py:39: UserWarning: ../data/hippocampus/input_data/deepcell_input/Point102.tif is a low contrast image
  io.imsave(save_path, out, plugin='tifffile')

omerbt:

To avoid the repeating warning: 'X.tif is a low contrast image'. closes #260 
This warning is raised when saving the preprocessed .tif images to the deepcell_input dir.

Setting the optional parameter `check_contrast` of skimage.io.imsave to false.

### DIFF
--- a/ark/utils/data_utils.py
+++ b/ark/utils/data_utils.py
@@ -38,7 +38,7 @@ def generate_deepcell_input(data_xr, data_dir, nuc_channels, mem_channels):
             out[1] = np.sum(data_xr.loc[fov, :, :, mem_channels].values, axis=2)
 
         save_path = os.path.join(data_dir, f'{fov}.tif')
-        io.imsave(save_path, out, plugin='tifffile')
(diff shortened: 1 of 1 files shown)


ngreenwald (changes requested):
Great! Not sure if you did this already, but can you do a search through the repository to see if there are any other spots where we save images that this same warning might come up?

omerbt:
We save images also in `plot_overlay`. The images that are saved in this function are not supposed to be with low contrast. I'm not sure if we should turn off this flag there, because in case we do get this warning when saving images in this function, it might be because of a real unexpected problem.

ngreenwald (approved):
Okay great, thanks

PR #298, opened 2020-10-26
Remove unnecessary headings from documentation

### LINKED ISSUE #296: Tweaks to RTD display
A couple things that I think would make RTD easier to use/more visually pleasing:

Rather than libraries, I think we should name the top-level modules, since these aren't actually different libraries
image

Within each module, we should remove the `Submodules` and `Module contents` headers, which don't have any content associated with them
image

Each script currently has `module` displayed after it, both in the menu on the left and within the documentation itself. We should remove this
image

alex-l-kong:

Addresses and closes #296. Our documentation has a few textual artifacts that don't need to be there and are confusing because they are empty headings. This PR focuses on nuking all of that.

We add a function in `conf.py` to make the documentation pages prettier and easier to follow along.

**Remaining issues**

If there is anything else we need to nuke from these pages (ex. removing all the repetitive `ark.analysis`, `ark.segmentation`, and `ark.utils`), I can further edit `conf.py` to do that.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -149,8 +149,7 @@ intersphinx_mapping = {
 intersphinx_cache_limit = 0
 
 
-# this function appends the proper lines and formatting to landing.md
-# so it can get displayed properly
+# appends README.md to landing.md and adds proper formatting
 def append_readme():
     with open(os.path.join('..', 'README.md')) as fin, open(os.path.join('_rtd', 'landing.md'), 'a') as fout:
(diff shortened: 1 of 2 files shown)


ngreenwald (approved):
Awesome, this looks great! I took a look at some other RTD pages, and it seems like the convention is to keep the full function name (i.e. `ark.utils.segmentation_utils.find_nuclear_mask`), so I take back my original suggestion.

PR #300, opened 2020-10-28
Address coverage issues

### LINKED ISSUE #256: Check for important missed test coverage
There is undoubtedly a number of lines that are not currently being tested. Some of these are likely not worth spending our time on. However, in order to avoid users discovering bugs by changing things, it would be great to test the ones that are likely to lead to errors. Especially uncovered lines that depend on variable names or variable formats that seem likely to change over time. 

Take a look through the coveralls output, either online or via the HTML output generated locally, to identify uncovered lines, and add in tests to cover those corner cases. 

Bug-report driven coverage increases

alex-l-kong:

Addresses and closes #256. We have quite a few lines that run important validation checks prior to performing an operation. While it's not feasible to test all of these, we should certainly try to do so where we can.

We improve our testing functions to cover more cases, mostly that we're throwing errors properly.

**Remaining issues**

Is there a need to address the coverage issues in `test_utils.py`? I don't think so since these are mostly for test functions. We should tell `coveralls` to disregard this file when computing coverage.

Also, is it worth writing test functions for `plot_utils.py`?

### DIFF
--- a/ark/segmentation/marker_quantification_test.py
+++ b/ark/segmentation/marker_quantification_test.py
@@ -211,16 +211,13 @@ def test_create_marker_count_matrices_base():
         marker_quantification.create_marker_count_matrices(segmentation_labels,
                                                            channel_data.values)
 
-    channel_data_bad = channel_data.copy()
-    channel_data_bad = channel_data_bad.reindex({'fovs': [0, 1]})
-
     segmentation_labels_bad = segmentation_labels.copy()
     segmentation_labels_bad = segmentation_labels_bad.reindex({'fovs': [1, 2]})
 
     with pytest.raises(ValueError):
         # attempt to pass segmentation_labels and channel_data with different fovs
         marker_quantification.create_marker_count_matrices(segmentation_labels_bad,
-                                                           channel_data_bad)
+                                                           channel_data)
 
 
 def test_create_marker_count_matrices_multiple_compartments():
--- a/ark/segmentation/marker_quantification_test.py
+++ b/ark/segmentation/marker_quantification_test.py
@@ -201,6 +201,27 @@ def test_create_marker_count_matrices_base():
     assert np.array_equal(normalized['chan0'], np.repeat(1, len(normalized)))
     assert np.array_equal(normalized['chan1'], np.repeat(5, len(normalized)))
 
+    # error checking
+    with pytest.raises(ValueError):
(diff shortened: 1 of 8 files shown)


ark/segmentation/marker_quantification_test.py
```diff
@@ -201,6 +201,27 @@ def test_create_marker_count_matrices_base():
...
+
+    with pytest.raises(ValueError):
+        marker_quantification.create_marker_count_matrices(segmentation_labels,
+                                                           channel_data.values)
+
+    channel_data_bad = channel_data.copy()
+    channel_data_bad = channel_data_bad.reindex({'fovs': [0, 1]})
+
+    segmentation_labels_bad = segmentation_labels.copy()
+    segmentation_labels_bad = segmentation_labels_bad.reindex({'fovs': [1, 2]})
```
ngreenwald, line 218: Only one of these changes should be necessary to create different FOVs, right?

ark/utils/spatial_analysis_utils_test.py
```diff
@@ -29,6 +32,21 @@ def test_calc_dist_matrix():
     assert np.array_equal(distance_mat["1"].loc[range(1, 4), range(1, 4)], real_mat)
     assert np.array_equal(distance_mat["2"].loc[range(1, 4), range(1, 4)], real_mat)
 
+    # file save testing
+    with pytest.raises(ValueError):
+        # trying to save to a non-existant directory
```
ngreenwald, line 37: :)

ngreenwald (changes requested):
Looks great! I think we should also test load_utils line 299, since that's the only place indices are actually used in that function

ark/segmentation/marker_quantification_test.py
alex-l-kong: I'll remove one of these reindexes.

ark/utils/spatial_analysis_utils_test.py
ngreenwald, line 37:
```suggestion
        # trying to save to a non-existent directory
```

ngreenwald (changes requested):

ark/utils/spatial_analysis_utils_test.py
alex-l-kong: 😅

ngreenwald (approved):

PR #305, opened 2020-10-28
Make sure Docker selects an open port if the user already has a Jupyter session open

### LINKED ISSUE #276: Running bash start_docker.sh will automatically try to use port 8888
**Describe the bug**

When a user tries to run `bash start_docker.sh`, the Jupyter server is automatically assigned port 8888. This is not usually a problem, but if the user already has a Jupyter server open that is using port 8888, you will not be able to open the notebook (technically, you will be taken to a page where you need to validate your token, but that will always fail because you need the token for the existing server using port 8888, and then you will be taken to that server and not the Deepcell notebooks).

This can be a problem if, say, a user has multiple important Jupyter servers open and one of them is using the default port 8888. It would be very inconvenient to tell the user that they have to kill all of the work they've been doing on the server using port 8888 before they can properly start the Docker.

**To Reproduce**

Open a Jupyter server using the default port 8888, then try running `bash start_docker.sh`.

**Expected behavior**

The proper functionality would be to test if port 8888 is being used first, and if so, then assign it to the next available port (so if 8889 is also taken, then 8890, and so forth).

alex-l-kong:

Addresses and closes #276. If the user is an active Python user, there's a good chance they will have an existing Jupyter session open when they run `bash start_docker.sh`. This will be a problem because there's a good chance that existing session will be on port 8888, and Docker as it stands will also start Jupyter on port 8888. This PR will fix this.

There's no real easy or clean way to fix this, unfortunately (Jupyter on Docker is a real pain to deal with), but I was able to get a solution going that requires a couple of more prompts to follow should the user encounter this. I don't think it will be a huge problem because I don't expect too many people to be in this situation (and I expect a person who is in this situation to have a level of comprehension that would allow him to decipher what's going on), but it is nonetheless an important edge case to address.

Think of this case like handling a merge conflict. It sucks to do, and you hope it doesn't come up often, but you'll be glad you put in a little more effort to get it going.

**Remaining issues**

Let me know what you think of the prompts and the updated `README.md`.


ngreenwald (commented):
Looks good to me. Maybe Adam has more useful feedback. Obviously we can't really test this programmatically, did you try this with a couple different instances simultaneously?

alex-l-kong:
> Looks good to me. Maybe Adam has more useful feedback. Obviously we can't really test this programmatically, did you try this with a couple different instances simultaneously?

Testing it on my end with one and two open Jupyter servers worked.

start_docker.sh
```diff
@@ -3,9 +3,24 @@
...
-# run docker and start notebook server
+# find lowest open port available
+PORT=8888
+
+until [[ $(lsof -i -P -n | grep 127.0.0.1:$PORT | wc -l) -eq 0 ]]
+  do
+    ((PORT=PORT+1))
+done
+
+echo $PORT
```
ackagel, line 14: shouldn't need this echo statement

start_docker.sh
```diff
@@ -3,9 +3,24 @@
...
+
+if [ $PORT -ne 8888 ]
+then
+  echo "Enter this URL instead to access the notebooks: http://127.0.0.1:$PORT/"
+  echo "In the URLs below, copy the token after \"token=\", paste that into the password prompt, and log in"
```
ackagel, line 19: maybe add another all caps 'WARNING' echo statement to draw attention to this.

ackagel (commented):
looks good.  some comments on the echo statements.

also, kind of a side note, but make sure that the lsof flags are 'posix'-compliant, in that most terminal sessions support them

ackagel (approved):
looks good

ngreenwald (approved):

PR #306, opened 2020-10-28
compression parameter

### LINKED ISSUE #301: Zipfile for deepcell upload is very large
**Describe the bug**
The process of creating a zip file appears to generate files that are larger than they should be. As mentioned in #297, 
> If I zip the 6 files manually (on Linux), the resulting zip file is 30.2MB. The one generated by the jupyter notebook is 201.3MB.  

**Expected behavior**
We should confirm that the zip file created by the notebook is the same size as one created manually, and identify what is causing the issue. This matters because of the current bug with the deepcell-kiosk, which fails for large zip files.

omerbt:
To change the implementation in`create_deepcell_output` so that the generated zip file will have a smaller size.
[Bug fix (closes #301)]

`create_deepcell_output`generates the zip file using the zipfile module (https://docs.python.org/3/library/zipfile.html#zipfile-objects). the default value for the `compression` parameter in zipfile.ZipFile is `zipfile.ZIP_STORED` (uncompressed archive member). I changed the value to `zipfile.ZIP_DEFLATED` (the usual ZIP compression method). 
After changing this parameter, the zip file created by the notebook is the same size as one created manually.

### DIFF
--- a/ark/utils/deepcell_service_utils.py
+++ b/ark/utils/deepcell_service_utils.py
@@ -3,7 +3,7 @@ from twisted.internet import reactor
 from kiosk_client import manager
 import os
 import glob
-from zipfile import ZipFile
+from zipfile import ZipFile, ZIP_DEFLATED
 import warnings
 
 from ark.utils import misc_utils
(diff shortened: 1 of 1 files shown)


ngreenwald (approved):
The default behavior is that there is no compression applied?!?!? Who thought that was a good idea??

PR #311, opened 2020-10-28
Remove file extensions from fovs in generate_cell_data

### LINKED ISSUE #303: generate_cell_table fails if .tiff extensions are in supplied FOV names
Screen Shot 2020-10-27 at 5 33 31 PM
I tried to segment a multi-channel tiff.  I was able to input  the tif file and create segmented mask xarray and overlay file.   I am trying  to get the expression matrices .csv file

**What is the problem?**
I  run into an error saying: Invalid value(s) provided for fovs variable: value(s) 2020-10-24T19-45-39_Run-255_FOV1_Breast2-mid-left-2-MassCorrected-Filtered.tiff not found in segmentation labels fovs list

Screen Shot 2020-10-27 at 5 33 31 PM

alex-l-kong:

Addresses and closes #303. As it stands, the user cannot specify fovs with file extensions in `generate_cell_data`. This PR will fix that.

We simply remove the file extensions as a precaution before running the rest of `generate_cell_data`. Because we're seeing this functionality being used in several places, we'll branch this off from `extract_delimited_names` into a separate function. By extension, this means we'll need to update the places in the repo where we call this.


ark/segmentation/marker_quantification.py
```diff
@@ -255,13 +255,16 @@ def generate_cell_table(segmentation_labels, tiff_dir, img_sub_folder,
     if fovs is None:
         # handle mibitiffs with an assumed file structure
         if is_mibitiff:
-            filenames = io_utils.list_files(tiff_dir, substrs=['.tif'])
+            filenames = io_utils.list_files(tiff_dir, substrs=['.tif', '.tiff'])
```
ngreenwald, line 258: I think the substring .tif will already match .tiff?

ngreenwald (changes requested):

ark/segmentation/marker_quantification.py
```diff
@@ -263,6 +263,9 @@ def generate_cell_table(segmentation_labels, tiff_dir, img_sub_folder,
             filenames = io_utils.list_folders(tiff_dir)
             fovs = filenames
 
+    # ensures removal of file extensions, happens if fovs is not originally None
+    fovs = io_utils.extract_delimited_names(fovs, delimiter=None)
```
ngreenwald, line 267: Okay, I couldn't figure out why this worked, but then I saw that the "default" for delimiter is actually the default for the second split, not the first split. It first splits off anything with a period, and then applies the specified user delimiter. I think this is unnecessarily complicated, and means that people can never specify a custom delimiter that doens't split based on period. 

I think we should change the default of this function to be splitting on a period, and remove the double split call. 
Is there any reason this would mess things up?

ngreenwald (changes requested):

ark/segmentation/marker_quantification.py
alex-l-kong: @ngreenwald I think the problem is that this function needs to handle both `fov1.tif` and `fov1_Row0Depth0....tif` types of files. The former doesn't need a double split, but the latter will.

I think we're better off setting the default delimiter to `None` because in that way, the user doesn't need to pass in any delimiter if they only want to split by period, but can additionally specify `_` if they need to. Also, we should change `delimiter` to something like `extra_delimiter` to avoid confusion. I don't see a use case where the user needs to specify multiple `extra_delimiters`, but that can be easily handled if we see this case pop up more often.

ngreenwald: If they have `fov1_Row0Depth0....tif`, won't it work if they supply that entire string, including .tiff, as delimiter?

alex-l-kong: @ngreenwald So I think this basically affirms the fact that `extract_delimited_names` is really confusing.

Since this function is used only for extracting text from file names, I was under the assumption that by default, the user would want the extension (`.tif`, `.tiff`, etc.) removed. In other words, calling split('.')[0] is always necessary and we don't need an extra argument for that.

The `extra delimiter` argument would be if another delimiter is needed to extract just the `fovn` text from the file name. That would apply for a file name like `fov1_Row0Depth0....tif`.

Yes, you could supply the entire string, including `.tiff`, as delimiter and it would work, but that's not a very intuitive way of doing it. I feel like throughout the repo, we're setting some crazy stuff as delimiter and while it works, it leads to general confusion overall.

I feel like we should stick to passing single character delimiters to simplify our calls to `extract_delimited_names`. It would definitely help make it clearer what this function should be doing.

ngreenwald: Good point. I think the main use case Adam wrote this for originally was to simplify loading of MIBI-Tiffs in our repo, since they had that extra metadata. However, we're now using it for general extension cleanup as well. 

How about you create a separate function, `remove_file_extension`, that will take a list and remove .tiff, .tif, .jgp, etc from a list of names. Then we can remove the first split from `extract_delimited_names`, and call  `remove_file_extensions` everywhere in the repo where that's really the only thing we want

alex-l-kong: @ngreenwald that sounds like a good idea. I'll adjust the PR description to include this cleanup task as well.

ark/segmentation/marker_quantification.py
```diff
@@ -266,14 +266,17 @@ def generate_cell_table(segmentation_labels, tiff_dir, img_sub_folder,
...
-            filenames = io_utils.list_folders(tiff_dir)
-            fovs = filenames
+            fovs = io_utils.list_folders(tiff_dir)
+
+    # drop file extensions
+    fovs = io_utils.remove_file_extensions(fovs)
+
+    # if mibitiff, also need to remove delimiters
+    if is_mibitiff:
+        fovs = io_utils.extract_delimited_names(fovs)
```
ngreenwald, line 279: I think this is going to lead to unexpected results. There's no particular reason that people using MIBITIFF files will definitely want any underscores removed from their FOV names, especially if it's buried inside this function and can't be modified. 

In fact, I think if you set MIBITIFF=True in the notebook, this will fail, since we don't do this delimiter removal when making the segmentation_labels.xr, which is what the FOVs are being compared to.

ark/utils/io_utils.py
```diff
@@ -72,18 +72,44 @@ def list_files(dir_name, substrs=None):
...
+        return
+
+    # only get the file name and not the directory path leading up to it
+    names = [os.path.split(name)[1] for name in files]
 
-    Examples:
-         - 'fov2_restofthefilename.tiff' becomes 'fov2'
-         - 'fov1.tiff' becomes 'fov1'
+    # remove anything past and including the first '.' in each entry
+    names = [name.split('.')[0] for name in names]
```
ngreenwald, line 96: This will fail if the filename has periods in it, which while not ideal, does happen. You can use os.path.splitext

ngreenwald (changes requested):
See my comments about automatically removing delimiters when not set to None; I think it's going to cause problems.

If you didn't already, can you do a control for extract_delimited_names and see if there are any other locations where we use it?

ark/utils/io_utils.py
alex-l-kong: Got it. By the way, this method will fail if the user passes in a file with more than 1 `.` in the extension (ex. `my_sample_tar.tar.gz`). Do we expect this case to pop up often, or can we safely assume that we won't be seeing it?

ark/segmentation/marker_quantification.py
alex-l-kong: @ngreenwald confirmed that setting `MIBITiff = True` will fail because the actual `fov` coordinate being set is that of the entire `fov8_Row8Depth8...`.

I can remove these two lines, but as a follow-up question, for MIBITiff inputs, can we safely assume that the coordinates  will always be that of the entire fov name (as in, `fov8_Row8Depth8...`)?

ngreenwald: No, I think people will have all kinds of creative ways of naming their mibitiffs. Hence, defaulting the delimiter to None will just pass it through unmodified. If people want to clean up their FOV names, they can either specify the correct, custom delimiter at each step in the pipeline, or clean them up prior to starting.

ark/utils/io_utils.py
ngreenwald: Agreed, if their image data is still in a tarball nothing is going to work anyway. Maybe add a warning message that gets raised if the cleaned up version of the filename still has a period in it.

alex-l-kong:
Verified that this passes wtih `MIBITiff = True` in `Segment_Image_Data.ipynb`.

ark/utils/io_utils.py
```diff
@@ -72,23 +72,61 @@ def list_files(dir_name, substrs=None):
...
+    # remove the file extension
+    names = [os.path.splitext(name)[0] for name in names]
+
+    # identify names with '.' in them: these may not be processed correctly
+    bad_names = [name for name in names if '.' in name]
+    if len(bad_names) > 0:
+        print(f"These files still have \".\" in them after file extension removal: "
+              f"{','.join(bad_names)}, "
+              f"please double check that these are the correct names")
+        warnings.warn("remaining periods in file names")
```
ngreenwald, line 108: Why have both a print statement and a warning? Does one not show up in certain locations?

ngreenwald (changes requested):
Looks good. I think it makes sense to hold off on integrating this until the jupyter notebook testing is up and running. Partly to give the high school students more time, and also because of all the different possible permutations of mibi-tiffs, multiple channels, etc, that aren't currently tested

ark/utils/io_utils.py
alex-l-kong: `warnings.warn` doesn't actually print the desired warning message, only provides something that you can catch and do something with. I didn't know if we were planning to write code to explicitly catch these in the future, so I left them in.

You're right, though: we need to stick to one or the other. I should open an issue that addresses these inconsistencies. My opinion is that we should drop `warnings.warn` and simply do `print` statements. Normally, I would go the other route, but these functions that throw warnings are mostly used directly in the notebooks, and we don't want to handle the warnings in the notebooks themselves.

alex-l-kong:
The functionality in this branch has moved over to #318, closing.

PR #314, opened 2020-10-29
Specify explicit requirements list for setup.py

alex-l-kong:

Another PR to try and address #242. Again, does not close until we can guarantee `pip install` works correctly.

Will from VVL recommends that we specify a manual requirements list for the `install_requires` parameter in our `setup` call, so we attempt to do that and see if it changes.

See above.

**Remaining issues**

If this doesn't work, there are a few more things we can try out:

1. Remove `dependency_links`. This one would technically be required to install `mibilib` correctly, and I confirmed that the link does indeed get version 1.3.0 of `mibilib`. But if it causes any noticeable problems, best to remove and have the user manually install `mibilib` separately.
2. Attempt to install from `tarball` instead of `archive`. I'm pretty sure this won't be the way to go: I read somewhere that `tarball` has been deprecated. VVL uses it but I think that's because they are eligible for this legacy `tarball` path since their repo is much older than ours.
3. See if certain packages are the ones causing the issue. So we can have a beta release that only installs a few packages such as `jupyter`, `matplotlib`, `numpy`, `pandas`, and `xarray`. If that works properly, then it would be pretty clear that `setup.py` doesn't like certain modules.


ngreenwald (changes requested):
We'll need to bump the version for a new release

PR #317, opened 2020-11-03
Overlay segmentation with neighborhood clusters

### LINKED ISSUE #295: Overlay segmented images with neighborhood clusters

The final visualization in `example_neighborhood_analysis_script.ipynb` (corresponding to the last TODO) should provide the user an image showing which cells were assigned to which neighborhood cluster generated. Before writing the generated DataFrame to .csv, this will serve as the check that everything went according to plan.

omerbt:

closes #295. Add a step in `example_neighborhood_analysis_script` to display images showing which cells were assigned to which neighborhood cluster generated.

Created a dictionary to map cell labels (`all_data_cluster_labeled`) to cluster labels (`label_maps`). Using this dictionary, the function `plot_clustering_result` will translate cell labels to cluster labels and display the resulted image.


ark/utils/plot_utils.py
```diff
@@ -6,13 +6,55 @@
...
+        figsize (tuple):
+            Size of the image that will be displayed.
+    """
+    if not default_val:
+        default_val = max(labels_dict.values()) + 1
+    if default_val in labels_dict.values():
+        raise ValueError('Invalid value for default_val. '
+                         'default_val should not appear as a value in labels_dict.')
+    # copy the given image so that it will not be overwritten
+    img = np.copy(labeled_image)
```
ngreenwald, line 44: This works, but a double nested for loop is going to be slow, especially if people are doing this across all of the images in their cohort. Check out how we used regionprops [here](https://github.com/angelolab/ark-analysis/blob/91610fa6a260d9b315eb273ebace2dc7fe561a94/ark/segmentation/marker_quantification.py#L87) to get all of the coordinates associated with each cell at once

ark/utils/plot_utils.py
```diff
@@ -6,13 +6,55 @@
...
+    for i in range(img.shape[0]):
+        for j in range(img.shape[1]):
+            img[(i, j)] = labels_dict.get(img[(i, j)], default_val)
+
+    plt.figure(figsize=figsize)
```
ngreenwald, line 49: It's generally good practice to split up plotting related functionality into a separate function from actually data manipulation. This will also make it easier to test the code. We can have one function whose only job is to relabel all of the segmentation_labels in the xarray, and returns the relabeled version. Then we can have a second function that takes the relabeled segmentation_labels, and plots them.

templates/example_neighborhood_analysis_script.ipynb
```diff
@@ -228,10 +228,42 @@
...
+     "metadata": {
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
    "source": [
-    "TODO: add visualization"
+    "fovs = ['fov6', 'fov7']\n",
+    "for fov in fovs:\n",
```
ngreenwald, line 262: Let's try and remove some of these details from the user. From my comment about, the for loop should get integrated into the relabeling fucntion, so that it takes an xarray and relabels all of them. Then the plotting function can take the name of a single FOV within the xarray, and produce the visualization of just that FOV

ngreenwald (changes requested):
Great start! Alex can help if there are issues with RTD

templates/example_neighborhood_analysis_script.ipynb
```diff
@@ -268,9 +300,9 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.10"
+   "version": "3.8.5"
```
alex-l-kong, line 303: Can you try running the notebook with Python 3.6? I don't think anything will change but since our repo uses Python 3.6, it would be good to make sure this code runs on that version just in case.

ark/utils/plot_utils.py
```diff
@@ -6,13 +6,55 @@
...
+            The title of the image that will be displayed.
+        default_val (int):
+            Cells that were not assigned to any cluster (according to labels_dict
+            will be translated to this value. This value should not appear in
+            labels_dict.
+        cmap (str):
+            Cmap to use for the image that will be displayed.
+        figsize (tuple):
+            Size of the image that will be displayed.
+    """
```
alex-l-kong, line 37: I generally like to leave a newline in between the docstring and the first line of code in a function, but I'm not sure if we're strictly enforcing this. @ngreenwald any comments?

ark/utils/plot_utils.py
```diff
@@ -6,13 +6,55 @@
...
+    and display an image showing which cells were assigned to which cluster.
+
+    Args:
+        labeled_image (numpy.ndarray):
+            2D numpy array of labeled cell objects.
+        labels_dict (dict):
+            a mapping between labeled cells and their clusters.
+        title (str):
+            The title of the image that will be displayed.
+        default_val (int):
```
alex-l-kong, line 29: `default_val` is slightly confusing: something more specific would be better.

ark/utils/plot_utils.py
```diff
@@ -6,13 +6,55 @@
...
+            Cells that were not assigned to any cluster (according to labels_dict
```
alex-l-kong, line 30:
```suggestion
            Cells that were not assigned to any cluster (according to labels_dict)
```

ark/utils/plot_utils.py
```diff
@@ -6,13 +6,55 @@
...
+from mpl_toolkits.axes_grid1 import make_axes_locatable
 
 from skimage.segmentation import find_boundaries
 from skimage.exposure import rescale_intensity
 
 
 # plotting functions
 
+def plot_clustering_result(labeled_image, labels_dict, title,
+                           default_val=None, cmap='tab20', figsize=(10, 10)):
```
alex-l-kong, line 18: @ngreenwald should we be setting the default `figsize` of the image to be `(1024, 1024)`?

alex-l-kong (changes requested):
Just a few more little things to add on top of Noah's comments.

ark/utils/plot_utils.py
ngreenwald: No, it's the size in inches or something, not pixels. 10x10 is a good size for plotting

ark/utils/plot_utils.py
omerbt: I'm not quite sure actually how to make it more efficient. It is a nested for loop, but it goes through each pixel exactly once (n^2 iterations, but actually it is linear in the size of the image). 
If we go in a loop over all cell labels, and then for each label use np indexing to find all entries of `img` equal to that label and change the value to the corresponding cluster label, it shouldn't be faster.

ngreenwald: The skimage.regionprops function does a lot of optimization under the hood to speed things up. Using the indices produced by regionprops ends up being a lot faster than looping over pixels. I have to admit I'm not actually sure why this is the case; maybe or @alex-l-kong or @vacuousplanet have a more satisfying explanation.

ark/utils/plot_utils.py
omerbt: How about `default_mapping` ?

ark/utils/plot_utils.py
alex-l-kong: @omerbt looking at the source code for `skimage.measure.regionprops` and `find_objects` in `scipy` (which `regionprops` relies on), it looks like the reason why `regionprops` runs faster is because the meat of `find_objects` is optimized to run in C.

templates/example_neighborhood_analysis_script.ipynb
omerbt: Checked it now, it works as expected

ark/utils/plot_utils.py
omerbt: I created another function: `relabel_img_array` in data_utils

ark/utils/plot_utils.py
omerbt: changed :)

ark/utils/plot_utils.py
omerbt: changed

ark/utils/plot_utils.py
omerbt: I changed the implementation (it is now in `relabel_img_array`)

ark/utils/plot_utils.py
```diff
@@ -6,11 +6,42 @@
...
 import matplotlib.pyplot as plt
+from mpl_toolkits.axes_grid1 import make_axes_locatable
 
 from skimage.segmentation import find_boundaries
 from skimage.exposure import rescale_intensity
 
 
+# plotting functions
+
+def plot_clustering_result(img_list, titles, cmap='tab20', figsize=(10, 10)):
```
alex-l-kong, line 17: Add an argument called `save_dir` if the user also wants to save each image into a directory (default argument should be `None`). You can use `misc_utils.save_figure` to do this:

```
if save_dir is not None:
    misc_utils.save_figure(save_dir, "file_name.png")
```

alex-l-kong (changes requested):
Looks awesome! Just a couple docstring checks and image saving needed. By the way, DeepCell is up again, so if you push another commit, given nothing else is broken, it should work!

alex-l-kong (approved):
Looks awesome!

ark/utils/data_utils.py
```diff
@@ -5,6 +5,58 @@
...
+        numpy.ndarray:
+            The relabeled array.
+    """
+
+    img = np.copy(labeled_image)
+    unique_cell_ids = np.unique(labeled_image.values)
+    unique_cell_ids = unique_cell_ids[np.nonzero(unique_cell_ids)]
+
+    for cell_id in unique_cell_ids:
+        img[img == cell_id] = labels_dict.get(cell_id, -1)
```
ngreenwald, line 55: The way this is currently implemented won't work as expected due to overlap between cluster labels and cell_id for small cell_ids.

Let's add some test functions for this and the function above. You can talk to @alex-l-kong and @vacuousplanet to see if any of our current test functions produce test data that makes sense here. Otherwise, to start off you can create a labeled array with say 20 different cells, sequentially labeled, and a corresponding dataframe with 3 different cluster types.

ark/utils/plot_utils.py
```diff
@@ -6,10 +6,48 @@
...
 
 from skimage.segmentation import find_boundaries
 from skimage.exposure import rescale_intensity
 
+from ark.utils import misc_utils
+
+# plotting functions
+
+
+def plot_clustering_result(img_list, titles, save_dir=None, cmap='tab20',
```
ngreenwald, line 19: Given that this is going to be run in a jupyter notebook, I don't think sequentially plotting all (potentially 100+) labels makes sense. Instead, let's allow the user to pass in the name of a specific FOV with the relabel_xr (see my comment from above) and plot that one

ngreenwald (changes requested):
Looks great, this logic makes a ton more sense. Next step is to add test functions that check the correctness of the functions in data_utils.

ark/utils/data_utils.py
omerbt: oh wow can't believe I missed that! I should definitely take the indices from the input image and not from the output

ark/utils/plot_utils.py
omerbt: changed

ark/utils/data_utils.py
```diff
@@ -5,6 +5,58 @@
...
+            The relabeled arrays (representing images).
+    """
+
+    images = []
+    for fov in fovs:
+        df = all_data.loc[all_data['SampleID'] == fov]
+        labels_dict = dict(zip(df.label, df.cluster_labels))
+        images.append(relabel_img_array(label_maps.loc[label_maps['fovs'] == fov].squeeze(),
+                                        labels_dict))
+    return images
```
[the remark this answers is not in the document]
omerbt: changed

ark/utils/data_utils.py
```diff
@@ -5,6 +5,58 @@
...
+        all_data (pandas.DataFrame):
+            data including fovs, cell labels, and cell expression matrix for all markers.
+        label_maps (xr.DataArray):
+            xarray of label maps for multiple fovs
+    Returns:
+        list:
+            The relabeled arrays (representing images).
+    """
+
+    images = []
```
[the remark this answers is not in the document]
omerbt: changed

omerbt:
I added tests for `relabel_img_array`. @ngreenwald are there things I should test in `relabel_images`? (I wasn't sure because it doesn't have much logic and basically calls `relabel_img_array`)

ngreenwald:
Agreed, I don't think you need to check the correctness of the `relabel_images` output, since that will be done in the tests for `relabel_img_array`. Instead, what would be useful is to make sure the output it produces makes sense, and to check corner cases. 

For example, if only a single FOV is specified, does the xarray have the right shape? If multiple FOVs are specified, are the non-zero pixels in each xarray the same as the non-zero pixels in label map? That should mostly cover it.

ark/utils/data_utils.py
```diff
@@ -4,6 +4,74 @@
...
+        fov_col (str):
+            column with the fovs names in all_data.
+    Returns:
+        xr.DataArray:
+            The relabeled images (dims: ["fovs", "rows", "cols"]).
+    """
+
+    # check if included fovs found in fov_col
+    verify_in_list(fov_names=fovs, unique_fovs=all_data[fov_col].unique())
+    verify_in_list(fov_names=fovs, unique_fovs=label_maps[cell_label_col])
```
ngreenwald, line 36:
```suggestion
    verify_in_list(fov_names=fovs, unique_fovs=label_maps.fovs.values)
```

ngreenwald: No need to have `fovs` as a modifiable parameter, this is used throughout the repo

ark/utils/data_utils.py
```diff
@@ -4,6 +4,74 @@
...
+
+    img_data = []
+    for fov in fovs:
+        df = all_data[all_data[fov_col] == fov]
+        labels_dict = dict(zip(df.label, df.cluster_labels))
+        img_data.append(
```
ngreenwald, line 42: I think this line is a bit too long, hard to see what's going on. I think it would be better to split it into one line that calls `relabel_img_array`, and a second line that appends the output.

Also, you can directly index into xarrays, which will simplify the expression:
`label_maps[fov, :, :, 0]`

ark/utils/data_utils.py
```diff
@@ -4,6 +4,74 @@
 import numpy as np
 import xarray as xr
 
+from ark import settings
+from ark.utils.misc_utils import verify_in_list
+
+
+def relabel_images(fovs, all_data, label_maps, cell_label_col='fovs',
```
ngreenwald, line 11:
```suggestion
def label_cells_by_cluster(fovs, all_data, label_maps, cell_label_col='fovs',
```

ark/utils/data_utils.py
```diff
@@ -4,6 +4,74 @@
...
+    """
+
+    # check if included fovs found in fov_col
+    verify_in_list(fov_names=fovs, unique_fovs=all_data[fov_col].unique())
+    verify_in_list(fov_names=fovs, unique_fovs=label_maps[cell_label_col])
+
+    img_data = []
+    for fov in fovs:
+        df = all_data[all_data[fov_col] == fov]
+        labels_dict = dict(zip(df.label, df.cluster_labels))
```
ngreenwald, line 41: df.label and df.cluster_labels should be changed to use the relevant global variables and pulled out as arguments to the function; these can change from one person's dataset to another

ark/utils/data_utils.py
```diff
@@ -4,6 +4,74 @@
...
+        img_data.append(
+            relabel_img_array(label_maps.loc[label_maps[cell_label_col] == fov].squeeze().values,
+                              labels_dict))
+    np.stack(img_data, axis=0)
+    return xr.DataArray(img_data, coords=[fovs, range(img_data[0].shape[0]),
+                                          range(img_data[0].shape[1])],
+                        dims=["fovs", "rows", "cols"])
+
+
+def relabel_img_array(labeled_image, labels_dict):
```
ngreenwald, line 51:
```suggestion
def relabel_segmentation(labeled_image, labels_dict):
```

ark/utils/data_utils_test.py
```diff
@@ -134,3 +138,54 @@ def test_split_img_stack():
 
         assert np.array_equal(sample_chan_1, data_xr[0, :, :, 0].values)
         assert np.array_equal(sample_chan_2, data_xr[0, :, :, 1].values)
+
+
+def test_relabel_img_array():
+    x = y = 5
+    img_arr = np.arange(1, x * y + 1).reshape((x, y))
```
ngreenwald, line 145: Great, let's add a case where multiple pixels have the same label to more accurately simulate. You can np.repeat to modify the above code, or make a new test

ngreenwald (changes requested):
Awesome work, see below for some more tweaks

ark/utils/data_utils.py
omerbt: done

ark/utils/data_utils.py
omerbt: I split it to multiple lines to keep it more clear. I tried `label_maps[fov, :, :, 0]` (and some variants) and it didn't work, is it the exact syntax?

ark/utils/data_utils.py
omerbt: changed

ark/utils/data_utils.py
omerbt: changed

ark/utils/data_utils.py
omerbt: changed

ark/utils/data_utils_test.py
omerbt: I added another test

ark/utils/data_utils.py
```diff
@@ -4,6 +4,77 @@
...
+        cluster_column (str):
+            column with the cluster labels in all_data.
+    Returns:
+        xr.DataArray:
+            The relabeled images (dims: ["fovs", "rows", "cols"]).
+    """
+
+    # check if included fovs found in fov_col
+    verify_in_list(fov_names=fovs, unique_fovs=all_data[fov_col].unique())
+    verify_in_list(fov_names=fovs, unique_fovs=label_maps.fovs.values)
```
ngreenwald, line 39: The way the verify function works is that when it fails, it reports the name of the argument that didn't match. So it helps to give the variables more informative names, so that the error message will help to identify the problem area. Right now, the same error message will be raised if the FOV isn't in label_map, or if the FOV isn't in all data. It would be better to rename `unique_fovs` so that it reflects where the fovs are coming from, i.e. `label_map_fovs`, `extracted_data_fovs`, or something like that

ark/utils/data_utils.py
ngreenwald: No I'm an idiot, It's `label_maps.loc`. Sorry!

ngreenwald (changes requested):
Just one small change for better input validation

templates/example_neighborhood_analysis_script.ipynb
```diff
@@ -12,7 +12,7 @@
     "import pandas as pd\n",
```
ngreenwald, line 12: <p>One small change here, I think giving the labeled output a more informative name (cluster_labels_xr, or something like that) would be useful to help distinguish the many different xarrays that will be produced. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/317/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

ark/utils/data_utils.py
omerbt: Changed to `label_map_fovs` and `all_data_fovs`

templates/example_neighborhood_analysis_script.ipynb
omerbt: changed

ngreenwald (approved):

ngreenwald:
Looks great!

PR #318, opened 2020-11-03
Extend notebook testing

### LINKED ISSUE #303: generate_cell_table fails if .tiff extensions are in supplied FOV names

### LINKED ISSUE #307: Better testing of notebooks
We've had a number of issues pop up with the jupyter notebooks. These basically boil down to the fact that the jupyter notebook is run as-is by Travis. That means that only the specific default parameters for each variable are being tested, and not the different ways that these can potentially be modified. This leads to holes in our test coverage, and hence bugs that are only discovered when users modify these parameters. 

We want to avoid this happening, by ensuring that the code in the notebooks is being more thoroughly tested. There are two options for this. 

1) Abstract more of the code currently in these cells into a function, and the test the function itself
2) Use a more intelligent testing framework than just running the notebook from start to finish. 

I think we've done a fairly good at number 1, and going further down this route is going to lead to extraneous functions. 

The question then is what are the options for option 2. I did some googling, and found this pretty informative discussion [thread](https://discourse.jupyter.org/t/testing-notebooks/701/18). I think it would be great if you guys could take a look through. What popped out to me as potentially the most promising is [this tool](https://test-book.readthedocs.io/en/latest/), which allows you to interact with and write tests for individual cells, variables, and functions within a notebook. 

We could use this to basically parameterize the cells of the notebook, generating fake input data of different shapes, modifying boolean flags, etc etc. 

@vacuousplanet @alex-l-kong, let me know what you think. Any other suggestions also welcome.

alex-l-kong:

Addresses and closes #307. By extension, addresses and closes #303. Users have found several bugs in `Segment_Image_Data` over the past couple weeks, especially when it comes to `MIBItiff` loading, which the notebook doesn't explicitly test. It's not feasible to duplicate this notebook several times over to handle various inputs, so we programmatically add them in.

We use a library called `testbook`, which provides a nice interface into Jupyter notebooks for testing. We programmatically can add cells to override default settings and test for errors popping up.

**Remaining issues**

Currently, this PR is only for addressing the testing framework. I'll include a couple of basic tests, but we'll open a separate PR after this one is done to address the addition of more comprehensive tests.


alex-l-kong:
Probably not ready to go yet, but here's an initial pass. Let me know what you think!

ark/utils/io_utils.py
```diff
@@ -72,23 +72,58 @@ def list_files(dir_name, substrs=None):
...
+        list:
+            List of files without file extensions
+    """
+
+    # make sure we don't try to split on a non-existent list
+    if files is None:
+        return
+
+    # only get the file name and not the directory path leading up to it
+    names = [os.path.split(name)[1] for name in files]
```
ngreenwald, line 95: I think we originally used this for the MIBITIFFs. Is there anywhere in the repo where we're still passing in complete paths rather than file names?

ark/utils/data_utils.py
```diff
@@ -22,9 +22,10 @@ def generate_deepcell_input(data_xr, data_dir, nuc_channels, mem_channels):
         mem_channels (list):
             membrane channels to be summed over
     Raises:
-            ValueError:
-                Raised if nuc_channels and mem_channels are both None or empty
+        ValueError:
```
ngreenwald, line 25: So does this work with one of them set to None, or it fails?

ark/utils/load_utils.py
```diff
@@ -65,6 +66,9 @@ def load_imgs_from_mibitiff(data_dir, mibitiff_files=None, channels=None, delimi
         channel_tuples = tiff.read(mibitiff_files[0]).channels
         channels = [channel_tuple[1] for channel_tuple in channel_tuples]
 
+    if len(channels) == 0:
```
ngreenwald, line 69: We've already checked that mibitiff_files has at least one valid entry, I don't think this error makes sense?

ark/utils/notebooks_test.py
```diff
@@ -23,3 +37,234 @@ def test_example_spatial_analysis():
...
+def segment_notebook_setup(tb, deepcell_tiff_dir="test_tiff", deepcell_input_dir="test_input",
+                           deepcell_output_dir="test_output",
+                           single_cell_dir="test_single_cell",
+                           is_mibitiff=False, mibitiff_suffix="-MassCorrected-Filtered",
+                           num_fovs=3, num_chans=3, dtype=np.uint16):
+    # import modules and define file paths
+    tb.execute_cell('import')
+
+    # create the path to the directory containing the input data
+    tiff_path = os.path.join(os.path.dirname(os.path.realpath(__file__)),
```
ngreenwald, line 52: Is there a reason we need this complicated setup rather than overwriting the path like you defined above?

ark/utils/notebooks_test.py
```diff
@@ -23,3 +37,234 @@ def test_example_spatial_analysis():
...
+
+    # define custom mibitiff paths
+    define_mibitiff_paths = """
+        base_dir = "../data/example_dataset"
+        input_dir = os.path.join(base_dir, "input_data")
+        tiff_dir = "%s/"
+        deepcell_input_dir = os.path.join(input_dir, "%s/")
+        deepcell_output_dir = os.path.join(base_dir, "%s/")
+        single_cell_dir = os.path.join(base_dir, "%s/")
+    """ % (tiff_path, deepcell_input_dir, deepcell_output_dir, single_cell_dir)
```
ngreenwald, line 90: What is this syntax with triple quotes followed by percent sign doing?

ark/utils/notebooks_test.py
```diff
@@ -23,3 +37,234 @@ def test_example_spatial_analysis():
 
 def test_example_neighborhood_analysis():
     _exec_notebook('example_neighborhood_analysis_script.ipynb')
+
+
+# testing specific inputs for Segment_Image_Data
+def segment_notebook_setup(tb, deepcell_tiff_dir="test_tiff", deepcell_input_dir="test_input",
```
ngreenwald, line 43: This looks great. I think for a first pass, we should move all of the helper functions into a separate file. Then this test file will just have the code where we run these tests.

ngreenwald (changes requested):
Awesome start. I just left high-level comments for now on the testbook stuff, since the organization was making it hard for to understand the exact logic.

ark/utils/io_utils.py
alex-l-kong: Let me double check.

ark/utils/data_utils.py
alex-l-kong: It works if one of them is set to `None`. I realize there was an obscure issue with my testing that was causing this to fail, not Omer's logic.

ark/utils/load_utils.py
alex-l-kong: Even if `mibitiff_files` has at least one valid entry, it will fail if the user passes `[]` for `channels`. This will happen in `Segment_Image_Data` if both `nucs` and `mems` are set to `None`.

ark/utils/notebooks_test.py
alex-l-kong: I don't know why I did it this way anymore, can definitely make this simpler.

ark/utils/notebooks_test.py
alex-l-kong: It's replacing each `%s` with the corresponding variable name. So the first `%s` gets set to `tiff_path`, the second one to `deepcell_input_dir`, etc.

ark/utils/notebooks_test.py
alex-l-kong: @ngreenwald so it looks like the other reason we need to do this is because currently, `validate_paths` requires a `../data` prefix. Unfortunately, this means we need to hack this into the path when creating the `deepcell_tiff_dir` directory. Is there a reason we need this `../data` prefix still?

ark/utils/load_utils.py
ngreenwald: Got it, in that case we should change the error message

ark/utils/notebooks_test.py
alex-l-kong: @ngreenwald OK, so it looks like the reason we have to do it this way is because of the way `pytest` looks at the directory structure. This is the same reason why we have to set `path` in `exec_notebook` using `os.path.dirname(os.path.realpath(__file__))` as the root as well. I understand it's confusing though, because it looks like `testbook` has a different root than `pytest`.

One way around this is to use `tb.inject` to create `tiff_path` inside the notebook, but I don't think it's worth having to format a long string to do this. I'm fine with leaving `os.path.dirname(os.path.realpath(__file__))` in just that once, but if you see a reason we need to remove it then I can go with the `tb.inject` method instead.

ark/utils/load_utils.py
```diff
@@ -142,7 +146,7 @@ def load_imgs_from_tree(data_dir, img_sub_folder=None, fovs=None, channels=None,
         channels = [chan for _, chan in sorted(zip(channels_indices, channels))]
 
     if len(channels) == 0:
-        raise ValueError("No images found in designated folder")
+        raise ValueError("No channels specified")
```
ngreenwald, line 149: So the reason this one says no images found in directory is because when channels = None, channels get populated by whatever images are in each tree. If there are none, its empty, hence the warning. The one I commented on above doesn't follow that logic, so the warning should be modified

ark/utils/notebooks_test.py
```diff
@@ -13,6 +21,7 @@ def _exec_notebook(nb_filename):
         subprocess.check_call(args)
 
 
+# test runs with default inputs
```
ngreenwald, line 24: I don't think we need the mocker decorator here

ark/utils/notebooks_test.py
```diff
@@ -23,3 +32,93 @@ def test_example_spatial_analysis():
...
+                                                single_cell_dir="test_mibitiff_single_cell",
+                                                is_mibitiff=True)
+
+    # default fov setting, standard nucs/mems setting
+    notebooks_test_utils.fov_channel_input_set(tb,
+                                               nucs_list=['chan0'],
+                                               mems_list=['chan1', 'chan2'])
+
+    # default fov setting, nucs set to None
+    notebooks_test_utils.fov_channel_input_set(tb,
```
ngreenwald, line 55: I don't think this is necessary. We already test `create_deepcell_input` with all permutations of blank and non-blank channel inputs, right? Or is this covering something different?

ngreenwald (changes requested):
Okay great, mostly minor comments

ark/utils/notebooks_test.py
alex-l-kong: You're right, we already test that. I'll remove those tests and just leave the one where we hard code a `fovs` list in.

alex-l-kong:
Looks like all that needs to be changed when the high schoolers' PR gets merged in is to add a tag for their `segmentation_overlay` function cell and make sure that integrates well with `notebooks_test.py`.

ark/utils/notebooks_test_utils.py
```diff
@@ -0,0 +1,224 @@
...
+    """
+
+    remove_dirs = """
+        from shutil import rmtree
+        rmtree(tiff_dir)
+        rmtree(deepcell_input_dir)
+        rmtree(deepcell_output_dir)
+        rmtree(single_cell_dir)
+    """
+    tb.inject(remove_dirs)
```
ackagel, line 224: weird nitpick, but I'm thinking that temp folder creation + clean up should maybe happen outside of the notebook.

I think all that would need to happen, is to skip the execution of the 'create_dirs' cell, make all the directories from within a tempfolder, and then let python do the clean up afterwards.

The only reason I bring this up, is because if someone runs this test locally, and has data in folders whose names happen to match those of the test folders, all that data would be deleted after the test runs.

ackagel (changes requested):
Looks great!  Everything looks pretty clean.  Just one point about creating/cleaning-up directories.

ark/utils/notebooks_test_utils.py
alex-l-kong: Great suggestion, I'll put that in!

alex-l-kong:
@ngreenwald @ackagel the notebook testing suite has been updated to handle the high schooler's PR!

ark/utils/notebooks_test.py
```diff
@@ -23,3 +33,66 @@ def test_example_spatial_analysis():
...
+        # hard coded fov setting
+        notebooks_test_utils.fov_channel_input_set(
+            tb,
+            fovs_to_load=['fov0_otherinfo-MassCorrected-Filtered.tiff',
+                          'fov1-MassCorrected-Filtered.tiff'],
+            nucs_list=['chan0'],
+            mems_list=['chan1', 'chan2'])
+
+        # generate the deepcell output files from the server
+        tb.execute_cell('create_output')
```
ngreenwald, line 62: I think sending data to the server is leading to slow testing; this now accounts for the majority of the Travis build time. Can we replace this with a test function to perform the same thing? 
image

ngreenwald (changes requested):
Looks great. One more speedup: it would be great if we could avoid sending to the server for the individual calls

ark/utils/notebooks_test.py
alex-l-kong: Yeah definitely agree. So basically a function that creates a dummy `segmentation_labels` value? Would be really easy to do.

ngreenwald: My assumption is that this is the slowest step; it could be something else in the notebook. Whatever it is, we should speed it up.

alex-l-kong: Just realized that this method would also force us to bypass the segmentation label loading step. We'll need to re-create a random `_feature_0.tif` for each `fov` specified. Shouldn't be much harder though.

alex-l-kong:
Tests runs much faster now with `create_deepcell_output` bypassed for the notebook unit tests.

ngreenwald (approved):
Looks good!

PR #323, opened 2020-11-05
Add scale argument to create_deepcell_output and run_deepcell_task functions

### LINKED ISSUE #322: Changing the rescaling value for deepcell input
I would like to change the rescaling value for the deepcell input programaticallyu

**What is the problem?**
I vaguely remember being able to change the rescaling value in the pre-processing notebook, but I can't seem to find how to do that.

**What have you tried so far?**
Looked through the code and went back to look at the pre-processing notebook.

awedwards:

Addresses and closes #322 . This change allows users to add a scale value as input to create_deepcell_output, which rescales the data before prediction.

I added ```scale``` as an input argument to the ```create_deepcell_output``` and ```run_deepcell_task``` functions in the util module. The default value is 1.0 for both functions. Finally, before the value is submitted to the server, it is converted from float to string, as required by the deepcell API. I figured that most users would intuitively provide the scale value as a float.

### DIFF
--- a/ark/utils/deepcell_service_utils.py
+++ b/ark/utils/deepcell_service_utils.py
@@ -10,7 +10,7 @@ from ark.utils import misc_utils
 
 
 def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
-                           suffix='_feature_0', host='https://deepcell.org', job_type='multiplex'):
+                           suffix='_feature_0', host='https://deepcell.org', job_type='multiplex', scale=1.0):
     """ Handles all of the necessary data manipulation for running deepcell tasks.
 
         Creates .zip files (to be used as input for DeepCell),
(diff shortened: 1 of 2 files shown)


ngreenwald (changes requested):
Looks good! Two things. 
1. Can you update the description of the PR to close your issue?
2. Can you test this out with some data? We don't do any correctness checks for the kiosk-client, so maybe you could take a file and run it using rescale of 2.0 and 0.5 to make sure it modifies predictions appropriately

awedwards:
I tested this with the data in the example_dataset folder, setting the rescale to 2.0 and 0.5. Both looked reasonable: at 2.0, the predicted objects were much larger than at 1.0, and at 0.5 the objects were smaller. I can upload some screenshots if that is necessary.

ngreenwald:
Awesome!

PR #324, opened 2020-11-05
Bump kiosk-client requirement

alex-l-kong:

Because the `attrs` library updated 12-13 hours ago, `treq` also had to be updated, and by extension, `kiosk-client`. `master` is currently failing because we need to update our `requirements.txt` to match the version of `kiosk-client`.

Simply bump the requirement in `requirements.txt` and `setup.py`.

**Remaining issues**

Is it best that we don't assert requirement ranges for `kiosk-client` for the time being?

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,5 +1,7 @@
+attrs>=20.3.0
 coveralls>=1.8.2,<2.0.0
 pytest<6.0.0
 pytest-cov<3.0.0
 pytest-mock<4.0.0
 pytest-pycodestyle<3.0.0
+six>=1.13.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-attrs>=20.3.0
 jupyter>=1.0.0,<2
 jupyter_contrib_nbextensions>=0.5.1,<1
 kiosk-client>=0.8.4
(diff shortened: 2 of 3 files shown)


ngreenwald:
Why would removing the requirements range for `kiosk-client` resolve this issue?

I think the underlying issue is that Will is switching from Travis to Github Actions, and some of the requirements used for testing have now been removed (https://github.com/vanvalenlab/kiosk-client/pull/72/files)

alex-l-kong:
> Why would removing the requirements range for `kiosk-client` resolve this issue?
> 
> I think the underlying issue is that Will is switching from Travis to Github Actions, and some of the requirements used for testing have now been removed (https://github.com/vanvalenlab/kiosk-client/pull/72/files)

OK, I'll leave the range in right now. The problem, which you may be hinting at, is that since Will remove the explicit requirements for `six` and `attrs` in his `setup.py`, it means we no longer get those libraries installed and we're now reverting back to the old versions. Technically, these should get installed by `treq`, but it looks like `treqs` requirement for `six` is getting overridden by `matplotlibs` requirement for `six (we use a very old version of `matplotlib`, consider updating?) and the requirement for `attrs` is ill-defined on `treqs` `setup.py` (it only requires a generic `attrs`, not the most recent version which was updated just 12 hours ago, thus we're defaulting to Travis' built in `attrs` which is incompatible with that of `kiosk-client`).

ngreenwald (approved):

PR #328, opened 2020-11-10
Changed arg scale to data_scale in run_deepcell_task BatchProcessingJob

### LINKED ISSUE #327: scale arg name needs to be changed to data_scale in the call to BatchProcessingJobManager
**Describe the bug**
Changing scale doesn't do anything -- This is because the BatchProcessingJobManager is looking for keyword "data_scale" and not "scale".

**To Reproduce**
Changing the scale to anything except default (1.0) and comparing to scale=1.0.

awedwards:
This PR will fix and close #327 . The deepcell ```BatchProcessingJobManager```` is expecting the keyword "data_scale" and not "scale". It is also expecting a float and not a string. This PR changes both of these issues.

I changed the argument name scale to data_scale in the ```run_deepcell_task``` method in ```utils.deepcell_service_utils.py```. I tested the changes on two different images by changing and comparing the scale to 0.5, 1 and 2.0. Not sure why I saw differences when I originally added this functionality -- when I re-ran the same code I got different results (in other words, the deepcell output wasn't changing even if I changed the scale).


ngreenwald:
What happens if you pass a string instead of a float?

awedwards:
It gives the correct output if you input a string. In other words, if you pass something like ```scale="2.0"``` you get the same output as ```scale=2.0```.

ngreenwald (changes requested):
Okay, in that case, can you add a step within `create_deepcell_output` that attempts to cast scale to a float if it isn't already one, and raises an informative value error if the input can't be cast to a float? And add a check to the test function that the appropriate value error is raised. You can take a look at some of our other functions where we do input validation and pytest.raises() checks in the test functions as an example

awedwards:
I added a try/except clause to check whether or not the scale arg can be cast in ```deepcell_service_utils``` and added a test in the ```deepcell_service_utils_test``` script.

ngreenwald (approved):

PR #329, opened 2020-11-11
Fix contributing.md headings so they don't all display in sidebar

alex-l-kong:

Fixes the headings in `contributing.md`. They are all h2 headings, meaning that RTD will format all of them to appear in the `Getting Started` section, which we don't want. We need to change all but the top heading to h3 to prevent this.

See above.


ngreenwald:
I was just about to open an issue for the same thing! We'll need to remember in the future to do a quick visual inspection of RTD before updating documentation to make sure everything looks okay

PR #331, opened 2020-11-11
Add section to add conda path in virtualenv.md

alex-l-kong:

Addresses and closes #331. Some users have reported issues with both the graphical installer of Anaconda (specifically for Mac OS X Catalina) and `PATH` issues. We need to address these in `virtualenv.md` since it seems like these are fairly common issues that have popped up with recent OS updates.

We now recommend the command line installer since it appears that this one consistently works. We also add commands the user should run to test if `conda` is in their path and if it isn't, commands to help them add it to their path.


docs/_rtd/virtualenv.md
```diff
@@ -8,7 +8,11 @@ For a step-by-step guide of how to install Anaconda, please refer to these links
 * https://docs.anaconda.com/anaconda/install/mac-os/ for Mac users
 * https://docs.anaconda.com/anaconda/install/windows/ for Windows users
 
-We recommend using the graphical installer for ease of use.
+We recommend following the command line installer instructions as users have reported recent issues with the graphical installer. 
+ 
+To test if `conda` has been added to your path, run `conda info` in your Terminal. If you get an error message, it means `conda` has not been added to your `PATH` environment variable yet. To fix this:
+* Run `export PATH="/Users/yourname/anaconda3/bin:$PATH"` for Mac users
+* Run `set PATH=%PATH%;C:\Anaconda3;C:\Anaconda3\Scripts\` for Windows users
```
ngreenwald, line 15: I think @vacuousplanet uses Windows? Or he did at some point? Can you confirm that this works?

ngreenwald (changes requested):

docs/_rtd/virtualenv.md
ackagel: I would never use Windows.  Not sure if this works

alex-l-kong: @ngreenwald @ackagel I'll change the template to address just Mac issues. If anyone runs into Windows issues we can address that at a later time, but it seems like the vast majority of our users will be Mac/Linux users.

ngreenwald (approved):

PR #333, opened 2020-11-12
Add exact pattern matching to list_files

### LINKED ISSUE #332: list_files in io_utils needs better substr matching
**Describe the bug**

It's finally time to tackle this bug! We have mentioned this in passing before, but it never caused a real issue until Candace's data came in.

The problem with the current matching scheme is that we simply test `if any([substr in file for substr in substrs])`. However, this can match more files than intended, which in ceratin cases, will break our code.

**To Reproduce**

The golden example is if the user includes `'C'` in the desired subset of channels, which is exactly what Candace is doing as a preprocessing step for her FlowSOM pipeline. `list_files` will pull _anything_ with a `C` in it, which includes every `CD` marker. If the user explicitly wanted to drop some `CD` markers, that could at best be an annoyance to clean out the residue.

At worst, it could completely break `load_imgs_from_tree` because in a certain case (if `channels` contains elements without file extensions), it will rightfully expect the list of channels provided to be equal to the desired file listing returned by `list_files`.

**Expected behavior**

In this example, `'C'` should only match `C.tiff` and nothing else. It's probably best to use a method like `os.path.splitext` for this.

alex-l-kong:

Addresses and closes #332. Our current substr matching scheme doesn't work if we pass in file names for exact matching (as opposed to file extensions, which we do most but not all of the time). This has caused problems when implementing FlowSOM preprocessing, so we patch this up.

We add a control flag `extensions` in `list_files` which will handle both file extension generic matching and file name exact matching.

We also patch up bugs in `load_imgs_from_tree` when the user passes in some channel names without file extensions and make the variable names more distinguished from each other.


ark/utils/io_utils.py
```diff
@@ -36,19 +36,20 @@ def validate_paths(paths):
...
+def list_files(dir_name, substrs=None, is_extensions=True):
     """ List all files in a directory containing at least one given substring
 
     Args:
         dir_name (str):
             Parent directory for files of interest
         substrs (str or list):
             Substring matching criteria, defaults to None (all files)
+        is_extensions (bool):
+            If True, will match on file extensions, if False, will match exact file names
```
ngreenwald, line 48: This isn't exactly correct. It won't just match extensions if is_extensions is true. That's why I'm not sure I like the name_change for this flag

ark/utils/load_utils.py
```diff
@@ -130,16 +130,16 @@ def load_imgs_from_tree(data_dir, img_sub_folder=None, fovs=None, channels=None,
         # need this to reorder channels back because list_files may mess up the ordering
         channels_no_delim = [img.split('.')[0] for img in channels]
 
-        channels = iou.list_files(
+        all_channels = iou.list_files(
```
ngreenwald, line 133: Do we need to modify load_imgs_from_dir as well?

ngreenwald (changes requested):

ark/utils/io_utils.py
alex-l-kong: Changed to `exact_match`, which I think is more informative.

ark/utils/load_utils.py
alex-l-kong: I don't think so. In `load_imgs_from_dir`, all we use `list_files` for is to get the images we need to load in the directory. We don't need to read in specific channel files because the channels exist in a dimension in each image to load. Whereas in `load_imgs_from_tree`, the channel data is contained in separate files which we have to read in manually.

ngreenwald: Should we set exact_match=True for load_imgs_from_dir?

ngreenwald: I think you could make a reasonable argument for either case

alex-l-kong: @ngreenwald I think we should just use `exact_match=False` since we're only doing file extension substr matching from what I can see.

ngreenwald (approved):

PR #340, opened 2020-11-16
Optimize data types in spatial analysis utils

### LINKED ISSUE #337: Memory Optimizations for Spatial Analysis
## **Description**
Suppose that someone were to have a FOV with ~20,000 cells.  The distance matrix generated for this FOV will take up a lot of memory.  In particular, for float64 (worst case), this distance matrix will take up 20,000^2 * 64 / 8 = **3.2 GB** of memory.  This isn't terrible on its own, but `compute_close_cell_num_random` is also fairly memory intensive, scaling linearly in memory with bootstrap number.  This leads to some users encountering 'out of memory errors' and dying kernels.

## **Proposed Solutions**
There a few optimizations (some easier than others) that can reduce the memory requirements:
 - The distance matrix is computed using `cdist` from scipy, which outputs float64.  We can shave off half of the occupied memory by changing this to float32.
 - Output array, `close_num_rand` doesn't need to be int32 (uint16 is probs sufficient)
 - Needless creation of xarray in compute_close_cell_num_rand
 - Both the distance matrix and `close_num_rand` are symmetric.  Numpy doesn't have a memory efficient data structure for exploiting symmetry.  Depending on the efficacy of other fixes, a quick class/object could be written for symmetric data structures, with numpy under the hood.

## **Alternate Solution**
Dynamic distance matrix computation.  Instead of storing the distance matrix in memory, it can be computed only when it's needed.  `cdist` is very fast, so recomputing the full distance matrix isn't a big deal.  This does create some code gymnastics on the backend, but user experience could improve.

ackagel:
## **Purpose**

Optimizes data types in `spatial_analysis_utils.py` for memory usage.  Closes #337  

## **Changes**

 - distance matrices where float64, and are now float32
 - `close_num` and `close_num_rand` were long's (int32 on most hardware), and are now uint16
 - binarized distance matrices where int8's, and are now uint8's
 - `np.sum()`'s on integer types where given `dtype` arguments to prevent default promotion to int64

## **Remaining issues**

There are probably more memory optimizations that could be performed, but let's deal with those as needed.

### DIFF
--- a/ark/utils/spatial_analysis_utils.py
+++ b/ark/utils/spatial_analysis_utils.py
@@ -218,7 +218,7 @@ def compute_close_cell_num(dist_mat, dist_lim, analysis_type,
                 mark1poslabels[j].values,
                 mark1poslabels[k].values
             ].values
-            count_close_num_hits = np.sum(dist_mat_bin_subset)
+            count_close_num_hits = np.sum(dist_mat_bin_subset, dtype=np.uint16)
 
             close_num[j, k] = count_close_num_hits
             # symmetry :)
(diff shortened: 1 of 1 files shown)


alex-l-kong (approved):
Looks nice!

ngreenwald (approved):

Issue #287, opened 2020-10-23
Add clustering to pipeline


ngreenwald:
**Is your feature request related to a problem? Please describe.**
Currently clustering must take place outside of the context of the pipeline. We should add in functionality to perform clustering so that we can have a seamless, integrated user-experience. 

**Describe the solution you'd like**
This is a high-level issue to track the progress towards the eventual end goal. Each bullet below should be its own PR. Please create a separate issue for each bullet once you are going to tackle it, and link it here.
- [x] Implement basic SOM in Python (#325). It looks like the most popular SOM package in python is [miniSOM](https://github.com/JustGlowing/minisom), but open to other options. We'll want to add basic functionality to load images, train the SOM, classify all pixels, and produce an output.

- [x] Write testing for SOM code (#344) . This should first involve running some of the sample data from the miniSOM package to ensure the outputs are the same, and then building out an actual test suite.

- [x] Benchmark current implementation to find slowest parts and most memory intensive processes (#345)

- [x] Add option to select random subset of pixels and use those for training (#362) , rather than the entire image

- [x] Implement pixel metaclustering (#364) to aggregate the 100 pixel clusters which are produced. 

- [x] Implement cell assignment (#365) to identify the cell type of each cell based on the pixel clusters of the corresponding pixels within that cell

- [x] Create skeleton Jupyter notebook for clustering pipeline

- [x] Reproduce code for heatmap visualizations in current pixel-clustering pipeline. This may overlap with #205. In particular, we'll need to add hierarchical clustering and a dendrogram display

- [x] Create interactive interface to manually re/assign cluster labels (#350)

- [x] Create visualization tools to overlay pixel clusters over image data (#351)

ngreenwald:
These are all wrapped up now

PR #342, opened 2020-11-16
Add SOM implementation 

### LINKED ISSUE #325: Create basic SOM implementation

alex-l-kong:

Addresses and closes #325. Candace's pixel clustering algorithm needs to be integrated with `ark-analysis` now. This PR begins to add this functionality in, starting with a basic SOM implementation. We need to verify that this PR produces the same or at least very similar results.

We piggyback off of MiniSOM (https://github.com/JustGlowing/minisom), a Python-based SOM implementation that should work for our purposes. It shouldn't be hard to translate most of the code into our project, but optimizing the process may be difficult.

Once we've verified our outputs, we then need to implement downsampling. A SOM would take a very long time to train on hundreds of thousands of pixel data from various images. We don't need all of these pixel entries.

We also need to implement 

**Remaining issues**

Preprocessing takes a very long time. Unfortunately, it seems like `apply_along_axis` doesn't offer much of a speedup. Alternatives would be greatly welcomed!


alex-l-kong:
I'm holding off on the tests for now (which explains the coverage failures). I'm still putting the SOM implementation in, but I would very much welcome a review of the preprocessing step and the notebook setup as it stands.

ark/flowsom/preprocess.py
```diff
@@ -0,0 +1,96 @@
+import os
+
+import numpy as np
```
ngreenwald, line 3: Lets change the directory name to be phenotyping, rather than FlowSOM, since we won't actually be using the package directly

ark/flowsom/preprocess.py
```diff
@@ -0,0 +1,96 @@
...
+    # set channels to all if None
+    if channels is None:
+        channels = img_xr.coords['channels'].values
+
+    # verify that the fovs and channels provided are valid
+    misc_utils.verify_in_list(fovs=fovs, image_fovs=img_xr.coords['fovs'].values)
+    misc_utils.verify_in_list(channels=channels, image_channels=img_xr.coords['channels'].values)
+
+    # delete any fovs and channels we don't need
+    # NOTE: no idea why Candace didn't use xarrays
```
ngreenwald, line 47: Lets try to avoid (publicly) code shaming our lab members :)

ackagel, line 47: :(

ark/flowsom/preprocess.py
```diff
@@ -0,0 +1,96 @@
...
+    if fovs is None:
+        fovs = img_xr.coords['fovs'].values
+
+    # set channels to all if None
+    if channels is None:
+        channels = img_xr.coords['channels'].values
+
+    # verify that the fovs and channels provided are valid
+    misc_utils.verify_in_list(fovs=fovs, image_fovs=img_xr.coords['fovs'].values)
+    misc_utils.verify_in_list(channels=channels, image_channels=img_xr.coords['channels'].values)
```
ackagel, line 44: `img_xr.coords['fovs'].values` -> `img_xr.fovs.values`
same with channels

ngreenwald (changes requested):
I think before we do any optimizing, we should confirm that this produces the same results as Candace's current preprocessing. Can you get the input CSV from her, as well as the example data, and add a simple assert that 1,000 randomly selected pixels from the first 10 images are exactly equivalent to hers?

I think for this project, it makes sense to keep this branch open for a while, and create/merge pull requests directly into the branch rather than master. We're going to get annoying questions from people as soon as a new notebook pops up, so I think it makes more sense to keep this on the SOM branch until we have a complete first pass at the whole pipeline.

ark/flowsom/preprocess.py
```diff
@@ -0,0 +1,96 @@
...
+            A matrix with pixel-level channel information for non-zero pixels in img_xr
+    """
+
+    # set fovs to all if None
+    if fovs is None:
+        fovs = img_xr.coords['fovs'].values
+
+    # set channels to all if None
+    if channels is None:
+        channels = img_xr.coords['channels'].values
```
ackagel, line 40: see prior comment about `.fovs`

ark/flowsom/preprocess.py
```diff
@@ -0,0 +1,96 @@
...
+import xarray as xr
+import scipy.ndimage as ndimage
+from skimage.io import imread
+
+import ark.settings as settings
+from ark.utils import load_utils
+from ark.utils import misc_utils
+
+
+def preprocess_flowsom(img_xr, seg_labels, fovs=None, channels=None, blur_factor=2):
```
ackagel, line 14: better name for this function is probably something like `extract_pixel_table`

ark/flowsom/preprocess.py
```diff
@@ -0,0 +1,96 @@
...
+    for fov in fovs:
+        # subset img_xr with only the fov we're looking for
+        img_xr_sub = img_xr_subset.loc[fov, ...].values
+
+        # apply a Gaussian blur for each marker
+        img_data_blur = np.apply_along_axis(ndimage.gaussian_filter, axis=2,
+                                            arr=img_xr_sub, sigma=blur_factor)
+
+        # flatten each image, assuming each image is square for now
+        pixel_mat = np.reshape(img_data_blur, (img_data_blur.shape[0]**2, len(channels)))
```
ackagel, line 63: This should do the same w/o square assumption

```suggestion
        pixel_mat = img_data_blur.reshape(-1, len(channels))
```

ark/flowsom/preprocess.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # convert into a dataframe
+        pixel_mat = pd.DataFrame(pixel_mat, columns=channels)
+
+        # assign metadata about each entry
+        # NOTE: no idea why Candace didn't use pandas
+        pixel_mat['fov'] = fov
+        pixel_mat['x_coord'] = np.repeat(range(img_data_blur.shape[0]), img_data_blur.shape[0])
+        pixel_mat['y_coord'] = np.tile(range(img_data_blur.shape[0]), img_data_blur.shape[0])
```
ackagel, line 72: Removing the square assumption
```suggestion
        pixel_mat['x_coord'] = np.repeat(range(img_data_blur.shape[0]), img_data_blur.shape[1])
        pixel_mat['y_coord'] = np.tile(range(img_data_blur.shape[1]), img_data_blur.shape[0])
```

ark/flowsom/preprocess.py
```diff
@@ -0,0 +1,96 @@
...
+                                            arr=img_xr_sub, sigma=blur_factor)
+
+        # flatten each image, assuming each image is square for now
+        pixel_mat = np.reshape(img_data_blur, (img_data_blur.shape[0]**2, len(channels)))
+
+        # convert into a dataframe
+        pixel_mat = pd.DataFrame(pixel_mat, columns=channels)
+
+        # assign metadata about each entry
+        # NOTE: no idea why Candace didn't use pandas
```
ackagel, line 69: :((

ark/flowsom/preprocess.py
```diff
@@ -0,0 +1,96 @@
...
+        pixel_mat['y_coord'] = np.tile(range(img_data_blur.shape[0]), img_data_blur.shape[0])
+
+        # assign segmentation label
+        seg_labels_flat = seg_labels.loc[fov, ...].values.flatten()
+        pixel_mat['seg_label'] = seg_labels_flat
+
+        # remove any rows that sum to 0
+        pixel_mat = pixel_mat.loc[pixel_mat.loc[:, channels].sum(axis=1) != 0, :]
+
+        # NOTE: no idea why Candace didn't do these preprocessing steps in Python
```
ackagel, line 81: :(((

ackagel (changes requested):
Looks like a good start; reviewing the nb in a sec.  How well does the output match Candace's original implementation?

ark/flowsom/preprocess.py
alex-l-kong: Crud, forgot to remove those 😆

alex-l-kong:
So the basic implementation (functional) of FlowSOM is in. I'm still working a few things out, but at this point, I'll definitely branch off into preprocessing-specific and clustering-specific branches to better address the facts that they are quite different from each other.

I discussed with @ngreenwald briefly, but it looks like there may be some inherent differences in the way MiniSOM and FlowSOM are implemented in terms of learning rates and sigma decay. Not sure how much of a difference this will cause between our outputs and Candace's in terms of clustering, we'll probably just have to see.

I've stuck with the default decay function (asymptotic), neighborhood function (Gaussian), and distance metric (Euclidean). If we need different functions for these, then that would be something to discuss.

Finally, there's also a decent amount of parameter passing, a consequence of converting MiniSOM to a functional-based approach. We can consider reverting back to an OOP-based approach if need be.

ark/phenotyping/cluster.py
```diff
@@ -0,0 +1,151 @@
...
+import pandas as pd
+
+
+def winner(sample, weights):
+    """Find the coordinates to the winning neuron in the SOM
+
+    Args:
+        sample (numpy.ndarray):
+            A row in the pixel matrix identifying information associated with one pixel
+        weights (numpy.ndarray):
```
ngreenwald, line 13: This description is confusing to me. I thought the weights matrix for the SOM represented the location in n-dimensional space for each of the neurons in the SOM? Isn't this matrix `[num_x * num_y * num_channels]`?

ark/phenotyping/cluster.py
```diff
@@ -0,0 +1,151 @@
...
+            A square matrix identifying the weights between two neurons
+
+    Returns:
+        tuple:
+            A coordinate array which indicates the winning neuron's position
```
ngreenwald, line 18: Simpler would be "the coordinates of the winning neuron"

ark/phenotyping/cluster.py
```diff
@@ -0,0 +1,151 @@
...
+    # define the activation map
+    activation_map = np.zeros((x_neurons, y_neurons))
+
+    # define meshgrid coords for the weights matrix, convert to float (yikes, memory...)
+    x_mesh, y_mesh = np.meshgrid(np.arange(x_neurons), np.arange(y_neurons))
+    x_mesh = x_mesh.astype(float)
+    y_mesh = y_mesh.astype(float)
+
+    # define the iterations iterable, this is WRONG in MiniSOM
+    iterations = np.arange(num_iters) % pixel_mat.shape[0]
```
ngreenwald, line 118: I think to begin with, we should define this the way it's defined in FlowSOM, where the number of passes through the dataset is specified, and each pass through the dataset looks at all of the samples. This will make it easier to compare to what the previous implementations have done before we start changing things

ark/phenotyping/cluster.py
```diff
@@ -0,0 +1,151 @@
...
+    ay = np.exp(-np.power(y_mesh - y_mesh.T[winning_coords], 2) / d)
+    g = (ax * ay).T * decay_lr
+
+    # update the weights based on the Gaussian neighborhood
+    weights += np.einsum('ij, ijk->ijk', g, sample - weights)
+
+    return weights
+
+
+def train_flowsom(pixel_mat, x_neurons, y_neurons, num_iters,
```
ngreenwald, line 77: FlowSOM specifically refers to the R package that combines SOM with minimum spanning tree and some nice visualization. We can just use SOM in our function names

ark/phenotyping/cluster.py
```diff
@@ -0,0 +1,151 @@
...
+        numpy.ndarray:
+            The weights matrix after training
+    """
+
+    # define the random generator
+    rand_gen = np.random.RandomState(random_seed)
+
+    # initialize the weights and normalize
+    weights = rand_gen.rand(x_neurons, y_neurons, pixel_mat.shape[1]) * 2 - 1
+    weights /= np.linalg.norm(weights, axis=-1, keepdims=True)
```
ngreenwald, line 107: This is just to make sure all weights are in [0, 1] right?

ark/phenotyping/cluster.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # define the activation map
+    activation_map = np.zeros((x_neurons, y_neurons))
```
ngreenwald, line 110: looks like this isn't needed here anymore now that there's an update function

ark/phenotyping/cluster.py
```diff
@@ -0,0 +1,151 @@
...
+    # define the iterations iterable, this is WRONG in MiniSOM
+    iterations = np.arange(num_iters) % pixel_mat.shape[0]
+    # iterations = np.array_split(pixel_mat.index.values, num_iters)
+
+    for t, iteration in enumerate(iterations):
+        # find the winning neuron's coordinates
+        winning_coords = winner(pixel_mat.loc[iteration, :].values, weights)
+
+        # update the weights, learning rate, and sigma
+        weights = update(pixel_mat.loc[iteration, :].values, weights,
```
ngreenwald, line 126: in line with my above comment, we should be able to pull out the learning rate update into this main for loop, rather than passing it to update

ngreenwald (changes requested):
This looks great! See my comments below. 

I've updated the main issue for tracking (#287), and put the bullets in the order that I think they should be worked on. Let me know if anything isn't clear.

ark/phenotyping/cluster.py
alex-l-kong: @ngreenwald yeah, it was confusing the way MiniSOM defines iterations, I had to double-check with the guy who maintains the repo exactly what was going on. So as you mention, `num_iters` should now be the number of passes through the dataset, not just the number of random data points to select with replacement (a la MiniSOM's style).

ark/phenotyping/cluster.py
alex-l-kong: I can do better: a coordinate defined in the x-y plane, which should contain two values, both of which should be contained in the set of whole numbers, which defines the position of the neuron in the weights matrix which corresponds to the closest Euclidean distance to the weights in question 😝.

ark/phenotyping/cluster.py
alex-l-kong: Yes, it should be. I'll make sure to clear that up with the `cluster.py` PR off of this one.

ark/phenotyping/cluster.py
alex-l-kong: Yup, we can remove if it's not necessary, just wanted to make sure we were consistently following MiniSOM's logic.

ark/phenotyping/cluster.py
alex-l-kong: @ngreenwald ah OK, this comment is confusing and should be changed. It turns out that MiniSOM actually doesn't ever update the learning rate and sigma parameter. So the decay function (`val / (1 + t / (num_iters / 2))`) uses the same based hyperparameters (learning rate and sigma), but they are changed based on the current iteration.

ark/phenotyping/cluster.py
ngreenwald: No it's definitely useful, just wanted to make sure I understood what was going on

ark/phenotyping/cluster.py
ackagel: another way of updating the learning rate/sigma out here iteratively:
```
decay_factor = (num_iters + 2*(t - 1)) / (num_iters + 2*t)
learning_rate *= decay_factor
sigma *= decay_factor
```

ark/phenotyping/preprocess.py
```diff
@@ -0,0 +1,93 @@
...
+        # flatten each image
+        pixel_mat = img_data_blur.reshape(-1, len(channels))
+
+        # convert into a dataframe
+        pixel_mat = pd.DataFrame(pixel_mat, columns=channels)
+
+        # assign metadata about each entry
+        pixel_mat['fov'] = fov
+        pixel_mat['x_coord'] = np.repeat(range(img_data_blur.shape[0]), img_data_blur.shape[1])
+        pixel_mat['y_coord'] = np.tile(range(img_data_blur.shape[0]), img_data_blur.shape[1])
```
ackagel, line 70: should the 0 and 1 be swapped?

ark/phenotyping/cluster.py
```diff
@@ -0,0 +1,151 @@
...
+    weights /= np.linalg.norm(weights, axis=-1, keepdims=True)
+
+    # define the activation map
+    activation_map = np.zeros((x_neurons, y_neurons))
+
+    # define meshgrid coords for the weights matrix, convert to float (yikes, memory...)
+    x_mesh, y_mesh = np.meshgrid(np.arange(x_neurons), np.arange(y_neurons))
+    x_mesh = x_mesh.astype(float)
+    y_mesh = y_mesh.astype(float)
+
```
ackagel, line 116: if the floats here are a big deal, we can just use `scipy.stats.mulitvariate_distribution` to generate `g` in the update function.  This reduces from 2*100^2 floats to either that many uint8's or no meshgrid at all (read through the `multivariate_distribution_docs` to figure that part out)

ackagel (commented):
Looks great so far.  Going to go checkout the notebook real quick.  Nothing much to add to Noah's comments aside from a few minor optimizations/generalizations.

ark/phenotyping/preprocess.py
alex-l-kong: Maybe, I'll look more into it in the `som_pre` PR off of this one.

ark/phenotyping/cluster.py
alex-l-kong: Now that I think about it, there's no real reason why converting to float type is bad in terms of memory. We're only ever going to be setting x and y to 10, so 100 floats should absolutely be able to be accommodated. However, if there's no difference in correctness between int and a float, definitely better to leave it as int and not have the extra conversion step.

ark/phenotyping/cluster.py
```diff
@@ -0,0 +1,151 @@
...
+        weights (numpy.ndarray):
+            A square matrix identifying the weights between two neurons
+
+    Returns:
+        tuple:
+            A coordinate array which indicates the winning neuron's position
+    """
+
+    # get euclidean distance between the sample and the weights
+    activation_map = np.linalg.norm(np.subtract(sample, weights), axis=-1)
```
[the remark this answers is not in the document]
alex-l-kong: I think for now, it's safe to assume Euclidean distances, I think that's what FlowSOM uses as well but I'll double check.

alex-l-kong:
This PR has been royally screwed, I've restarted a separate one so this one's nuked.

Issue #345, opened 2020-11-18
Benchmark SOM codebase


ngreenwald:
Before we start optimizing a bunch of stuff, we should figure out what the current bottlenecks are, both in preprocessing, training, and running the trained SOM. 

To do this, I recommend using the `timeit` package. It allows you to create timestamps throughout your code. 

For example, 
```
from timeit import default_timer
start_time = default_timer()
# do task 1
time_1 = default_timer()
# do task 2
time_2 = default_timer()
print("task 1 took {} seconds".format(time_1 - start_time))
```

We should do this for all of the steps in the preprocessing, training, and running. I would suggest randomly generating 10 different 1024x1024x20 images to use for this.

alex-l-kong:
Haven't benchmarked the SOM clustering step yet, but in `preprocess.py`, it's clear that the call to `ndimage.gaussian_filter` across 20 1024x1024 channels is the big time eater. Even if the optimizations provided by `apply_along_axis` are used, the Gaussian filtering still takes a little over a minute to complete for each image.

Removing the `fovs` loop with another `apply_along_axis` is one option for a minor speedup, but this does not eliminate the Gaussian filtering bottleneck. Python is embarrassingly bad with multithreading, but one interesting option using the `multiprocessing` library is presented here: https://stackoverflow.com/questions/45526700/easy-parallelization-of-numpy-apply-along-axis. The issue with this one is that it could be extremely memory intensive, but it's worth a shot.

ngreenwald:
Got it, makes sense

alex-l-kong:
@ngreenwald just benchmarked the SOM weight training, each step about as optimized as they can be, only taking on average 0.00001 seconds to complete. There are a total of 4 steps I timed, meaning on average on weight update takes just 0.00004 seconds.

This means the only bottleneck is the number of weight updates that need to happen. In the case of Candace's dataset, about 400,000,000 updates need to happen, meaning this process will take several hours to complete (on average 4.5).

The only way I can see this process being optimized is to take a small subset of the dataset. Since we already have a step for that in #287, I think we can close this issue.

ngreenwald:
Before you close it, it would be useful to have a summary of how long each step takes right now. 

For example, the preprocessing, training a SOM for X steps, running the som on all of the data

alex-l-kong:
@ngreenwald OK, so here's the complete rundown of the average times for Candace's dataset on a 2019 MacBook Pro 64 GB RAM machine:

[...]

PR #346, opened 2020-11-19
Implement SOM training and cluster assignment

### LINKED ISSUE #325: Create basic SOM implementation

alex-l-kong:

Addresses and closes #325. Addresses half of #345. We add the SOM training and testing processes in this PR.

We'll be working off of the MiniSOM repo which provides a nice lightweight and intuitive implementation of SOMs. In line with the rest of our repo, we convert the process to a functional approach and only take the functions that are applicable. The default activation, neighborhood, and distance functions should correspond to the defaults for the R SOM implementation as well.

Our work will be in `cluster.py` and `cluster_test.py`.

**Remaining issues**

We try to use the same parameters that the R SOM package uses (which is what Candace calls it with), but there are a few things that we may not be able to duplicate completely (ex. the default functionality of R SOM is to also specify an end learning rate, whereas MiniSOM has no such requirement). Hopefully, this does not lead to significant differences in the clustering results.

While the original MiniSOM author seems to have done his best to speed up the process with optimized NumPy code, the biggest bottleneck right now seems to be the number of rows Candace has (almost 4 million, which means 40 million updates because R SOM defaults to 10 passes through the data). No joke: training the SOM on the entire dataset takes 3 HOURS to run. Thank goodness Tyler had some stuff to do 😅. While we will address subsetting in a future PR, for now, I think it would be best to ask Candace to run FlowSOM and return the output of SOM on a much smaller subset of her data. Using the same subset, we'll compare the outputs we get on our end. It will greatly speed up verification for this PR.


alex-l-kong:
Still need to add testing and potential optimizations in, but please let me know how this looks so far!

@ngreenwald using our minimized functional approach, I was able to perfectly replicate this example in the MiniSOM repo which most closely aligns with our goals: https://github.com/JustGlowing/minisom/blob/master/examples/Clustering.ipynb. So I think it's cool to begin writing the test functions assuming they work.

ngreenwald (approved):
This looks great. I think you should edit the description so that this only closes the first issue, and create a separate PR for implementing testing. Smaller PRs like this are much easier to review. 

For now, I don't think it's worth updating the example jupyter notebook. This is likely going to change substantially as we make more changes to the API/frontend of the SOM stuff. I'm sure it's useful for you to use to test stuff out, but I don't think these changes need to go in the PRs. 

I agree that eventually we want Candace to run the SOM on a smaller subset of the data. But the first step I think should be to run both methods on the entire dataset, since that's what she's doing. Now that you've reproduced one of the examples from the miniSOM repo, the next step is to write our own tests. Once those are done, I think we can make a version for candace to run, and compare the outputs

ackagel (approved):
Looks great!

PR #348, opened 2020-11-19
Add testing for SOM clustering

### LINKED ISSUE #344: Testing for SOM implementation
Now that we have ported over the SOM code, we'll need to test it. 

The first step will be to try and recreate one of the basic analyses from the miniSOM repo. This might require a few tweaks to the code base, and likely a new jupyter notebook for testing. I don't think this specific test needs to be checked in to the repo, this will just be to weed out any obivous errors before we start writing tests. 

Once you've successfully recreated some of their analyses, we can start testing. We should add **very basic**  tests for all of the new functions. This will include correctness tests for the helper functions, along with a simple execution check for the training function.

alex-l-kong:

Addresses half of #344. Now that our SOM approach has proven to pass on a concrete example, we can begin adding tests. These will be fairly basic since it would be cumbersome to test on large datasets and assert exact equality.

We add `cluster_test.py` to handle testing `cluster.py`.


ark/phenotyping/cluster_test.py
```diff
@@ -0,0 +1,71 @@
...
+    test_t = 1
+    test_num_iters = 1
+
+    # it's impractical to check exact values, so we'll round to make life easier
+    weights = cluster.update(test_sample, test_weights, test_winning_coords,
+                             test_sigma, test_learning_rate, test_x_mesh, test_y_mesh,
+                             test_t, test_num_iters)
+    weights = np.round(weights, decimals=8)
+
+    result = np.array([[[0.00108549, 0.00514249],
```
ngreenwald, line 39: Can we also check that weights of the matching neuron are the ones that are changed the most?

ackagel (commented):
Looks good so far.  Not sure if you were planning this already, but before we close #344 , we'll probably want to add testing to the preprocessing stuff, unless that's going in a separate issue.

alex-l-kong:
@ackagel yes, that's a good point. I'll remove the closing tag on #344.

ark/phenotyping/cluster_test.py
```diff
@@ -0,0 +1,71 @@
...
+                        [0.00231832, 0.00463663]],
+                       [[0.00327853, 0.00943768],
+                        [0.00833333, 0.02583333]]])
+
+    assert np.all(weights == result)
+
+
+def test_train_som():
```
ngreenwald, line 47: I think we actually can add a correctness check. What about a 2x2 SOM with 3 channels. You can create 4 unique pixel rows, and do 100 iterations or so of training by showing those same 4 pixel rows. Then check that the SOM has converged/is close to those 4 values

ngreenwald (changes requested):
Awesome start! Looks great

ark/phenotyping/cluster_test.py
```diff
@@ -0,0 +1,71 @@
...
+    test_winning_coords = (1, 1)
+    test_sigma = 1.0
+    test_learning_rate = 0.5
+    test_x_mesh, test_y_mesh = np.meshgrid(np.arange(2), np.arange(2))
+
+    test_t = 1
+    test_num_iters = 1
+
+    # it's impractical to check exact values, so we'll round to make life easier
+    weights = cluster.update(test_sample, test_weights, test_winning_coords,
```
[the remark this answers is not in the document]
alex-l-kong: Ok cool, it's the same `asymptotic_decay` function that modifies `sigma` so it will translate easily.

ark/phenotyping/cluster_test.py
alex-l-kong: What's our definition for convergence for this test? Some pixel rows exhibit strong convergence but others are on the order of 3 times different after being trained on a 2x2 SOM with 3 channels (using 4 pixel rows, 100 iterations).

ngreenwald: I assume right now this runs extremely quickly. If we can keep this entire test under 3 seconds, we can maybe add some more iterations? Or more channels, to help make each unique pixel row more distinct. Perhaps the assertion could be that the difference between each pixel row and its closest neuron is much less (10X?) than the difference between any two neurons?

Whatever you think makes sense.

alex-l-kong: @ngreenwald I can try the other methods first, and if it doesn't work, we can settle for a not-so-specific convergence.

alex-l-kong: So it looks like no matter what we do, we're always going to get a few random weights that is out of range. For the time being, I've added a very flexible check for weight convergence (if no more than the sqrt of the number of weights are out of range).

ark/phenotyping/cluster.py
```diff
@@ -39,9 +58,9 @@ def update(sample, weights, winning_coords, sigma, learning_rate,
         winning_coords (tuple):
             A coordinate array which indicates the winning neuron's position
         sigma (float):
-            Determines the spread of the Gaussian neighborhood function
+            Determines the spread of the Gaussian neighborhood function, decayed
         learning_rate (float):
-            Determines how sensitive the weight updates will be to new data
+            Determines how sensitive the weight updates will be to new data, decayed
```
ngreenwald, line 63: can remove `t` and `num_iters` now

ark/phenotyping/cluster.py
```diff
@@ -119,9 +132,13 @@ def train_som(pixel_mat, x_neurons, y_neurons, num_passes,
         # find the winning neuron's coordinates
         winning_coords = winner(pixel_mat.loc[row, :].values, weights)
 
+        # decay both sigma and learning_rate using the decay function
+        decay_sigma = decay_function(sigma, t, num_iters)
+        decay_learning_rate = decay_function(learning_rate, t, num_iters)
```
ngreenwald, line 137: Is this the same as the LR decay in miniSOM? I would assume it only gets decayed after each pass through the dataset

ark/phenotyping/cluster_test.py
```diff
@@ -0,0 +1,97 @@
...
+    test_pixel_mat = pd.DataFrame(np.random.rand(9, 4))
+    test_x = 3
+    test_y = 3
+    test_num_passes = 1000
+
+    # only to see if it runs to completion with default sigma, learning_rate, and randomization
+    weights = cluster.train_som(test_pixel_mat, test_x, test_y, test_num_passes, random_seed=0)
+
+    # check if our weights are within 1/10 times to 10 times the original pixel values
+    threshold = np.logical_and(weights.flatten() * 10 > test_pixel_mat.values.flatten(),
```
ngreenwald, line 76: This assumes that the first row in the example data corresponds to the first neuron in the SOM, the second row in the example data corresponds to the second neuron in the SOM, etc, correct? 

The neuron corresponding to the first row in the example data could be in any of the 9 positions in the SOM. We'll need to first identify, for each neuron in the SOM, the example row that it is closest too. Then, once the closest match is found, we can check the distance

ngreenwald (changes requested):
Looks good, think I spotted why you're not getting good convergence

ark/phenotyping/cluster.py
alex-l-kong: Yup, LR needs to be decayed after each iteration in MiniSOM.

ark/phenotyping/cluster_test.py
alex-l-kong: Ah yes, totally forgot about that.

ark/phenotyping/cluster.py
ngreenwald: FlowSOM does it per pass through the dataset

alex-l-kong: @ngreenwald OK, so in a separate PR (which I'll also use to address a potential speed-up idea for training), we should probably update our decay to match that of FlowSOM so we get consistent results.

ngreenwald: Yes, I think in the long term we'll want the one you have now, but for initial testing we should switch to match FlowSOM. You can add a separate decay function that works across iterations, and one that works across dataset splits. Can you also go through and summarize what the remaining differences are between this and FlowSOM? For example I think you mentioned that the Gaussian function is different for updates?

alex-l-kong: @ngreenwald based on the documentation of `SOM` (contained here: https://www.rdocumentation.org/packages/FlowSOM/versions/1.4.0/topics/SOM), it appears that the biggest difference is in the way the learning rate parameter is being updated. It also appears that R SOM precomputes the neighborhood matrix prior to training (which I guess is a nice speedup) and uses the supremum norm to do so (rather than a Gaussian).

ark/phenotyping/cluster_test.py
```diff
@@ -0,0 +1,98 @@
...
+    test_num_passes = 1000
+
+    # only to see if it runs to completion with default sigma, learning_rate, and randomization
+    weights = cluster.train_som(test_pixel_mat, test_x, test_y, test_num_passes, random_seed=0)
+
+    # find the winning coordinates associated with the neuron weights closest to each pixel
+    cluster_coords = test_pixel_mat.apply(
+        lambda row: cluster.winner(np.array(row.values), weights), axis=1)
+
+    # now verify that the assigned weights are the ones that are the closest to each pixel
```
ngreenwald, line 75: Isn't this tautological? We already know the chosen weight is closest, that's how it was picked. 

What we should be checking is how close the weight is to the corresponding matching value. The magnitude of the difference should be small either compared to the magnitude of the weight itself, or compared to the magnitude of the smallest difference between the weights themselves

ngreenwald (changes requested):

ark/phenotyping/cluster.py
ngreenwald: Okay great, that can be in a separate PR

ark/phenotyping/cluster_test.py
alex-l-kong: @ngreenwald oh oops, let me fix that.

For this problem, it's hard to quantify exactly what "close" should mean. Generally, the difference between the weight and its matching value will have a magnitude of no more than 0.3, but there are exceptions (some can be as high as 0.5 if the weight initialization is just that awful).

I have yet to encounter any such initialization where the smallest difference between the weight and its matching value was 0.75 or even 0.6, though. That isn't to say that this case will never pop up, but if we're OK with the test very rarely failing, then I think we can use this as our benchmark.

The other alternative is to hard code `test_pixel_mat` in, but that kind of defeats the purpose of the `test_train_som`.

ark/phenotyping/cluster_test.py
```diff
@@ -0,0 +1,98 @@
...
+    for row, coord in enumerate(cluster_coords.values):
```
ngreenwald, line 76: Can you also check that the predicted coordinates are unique, i.e. each pixel was assigned to a different neuron

ngreenwald (changes requested):
looks good, one last thing to add

ark/phenotyping/cluster_test.py
alex-l-kong: Oof, from the runs I've generated it seems like that frequently will not be the case, at least for a 3x3x4 SOM and 9 pixel rows. It's likely related to the lack of training data and/or the random generation of the weights which causes the SOM to be unable to fill up each home with pixel points. I can look into this some more over the weekend.

ngreenwald: I would suggest switching to a 2x2x6 SOM. That way there are fewer nodes, and each node is in higher dimensional space. That should make it more likely that each node gets uniquely mapped to one of the four unique pixel rows

ackagel (commented):
looks great, nothing to add from me

ngreenwald (approved):

PR #349, opened 2020-11-19
Boost Gaussian blur step in preprocessing

alex-l-kong:

Addresses half of #345. Confirmed that the Gaussian blur step took the longest time for preprocessing. `np.apply_along_axis` can only get us so far in terms of optimization, but we can further boost performance by running `apply_along_axis` in parallel (because there are no dependencies across different channels).

I used an idea found here: https://stackoverflow.com/questions/45526700/easy-parallelization-of-numpy-apply-along-axis. It worked really well on my machine.

**Remaining issues**

- While this method is fast, it comes with a potentially massive memory hit.
- The performance boost I ended up getting was on the order of 6x faster, but I have 8 CPUs on my end. It may be on the order of just 2-4x faster for 2 or 4 CPUs, though that's already a massive improvement right there.
- Testing will be done in a separate PR.


ark/phenotyping/preprocess.py
```diff
@@ -10,6 +11,60 @@
...
+            The function to apply along the given axis
+        axis (int):
+            The axis to run
+
+    Returns:
+        numpy.ndarray:
+            The processed array, equivalent to what np.apply_along_axis returns
+    """
+
+    # If axis is 0 we'll effectively be working with the 1st axis
```
ngreenwald, line 47: Why?

ark/phenotyping/preprocess.py
```diff
@@ -10,6 +11,60 @@
 from ark.utils import load_utils
 from ark.utils import misc_utils
 
+from timeit import default_timer
```
ackagel, line 14: vestigial import

ark/phenotyping/preprocess.py
```diff
@@ -10,6 +11,60 @@
...
+            The processed array, equivalent to what np.apply_along_axis returns
+    """
+
+    # If axis is 0 we'll effectively be working with the 1st axis
+    effective_axis = 1 if axis == 0 else axis
+
+    # Swap the 0th and 1st axis if axis specified is 0
+    if effective_axis != axis:
+        arr = arr.swapaxes(axis, effective_axis)
+
```
ackagel, line 53: idk why we're doing this, but if we need this then we can just do:
```
if axis == 0:
    arr = arr.swapaxes(0, 1)
    axis = 1
```

ark/phenotyping/preprocess.py
```diff
@@ -10,6 +11,60 @@
...
+    # Map the array subsets
+    chunks = [(func1d, effective_axis, sub_arr, args, kwargs)
```
ngreenwald, line 55: cpu_count is a good default, but we should add in an optional parameter to the main function which allows the user to specify a smaller number of parallel threads in case they run into memory issues. 

Correct me if I'm wrong, but the actual amount of memory being used here is not that substantial, right? We're talking about at most 8 images being run simultaneously? This seems fairly insignificant when compared to loading an xarray containing an entire cohort of images

ngreenwald (changes requested):

ark/phenotyping/preprocess.py
```diff
@@ -10,6 +11,60 @@
...
+              for sub_arr in np.array_split(arr, multiprocessing.cpu_count())]
+
+    # Run the process in parallel
+    pool = multiprocessing.Pool()
+    individual_results = pool.map(_unpacking_apply_along_axis, chunks)
+
```
ackagel, line 61: Have you looked at using `concurrent.futures` instead?  I think the chunking will be the same but it might be worth comparing the performance between a threadPool vs a processPool (plus the API is a little better)

ackagel (commented):
Looks like a good start.  Made a few recommendations.

Big question for me is how massive the memory cost is.  If it's due to the copying of `arr` into `chunks`, you could instead change `_unpacking_apply_along_axis` to:

```
(func1d, axis, arr, args, kwargs, region) = func_args
return np.apply_along_axis(func1d, axis, arr[region], *args, **kwargs)
```

where `arr` is the whole array.  Then you only have to store regions.  Changes the interface a little, and it could probs be optimized more from the version above, but if that's the source of the big memory cost, then this might be a solution.

ark/phenotyping/preprocess.py
alex-l-kong: Didn't think too much about this, I can remove and if it causes problems with the 0th axis we can put it back and examine further.

ark/phenotyping/preprocess.py
ackagel: I think the `Pool` here is actually a process pool, which is slightly different than a threadpool.  Instead of everything running on a thread, everything is a different system process.  Basically, each process is slower, but you can get more than the cpu core limit I believe.

As for the memory, I think it's being duplicated (not referenced) when the chunks are made, and the float64s don't help that. Loading an entire cohort of images makes this seem not bad, but most of the time, those images are uint8 or uint16 at worse.  When you load in float64 tiffs (mibitiffs default to this 🙄), it's a different story -> ~(8*num_channels) MB per image.  If you have 20 channels, 7 images already occupy >1GB of ram.

All that to say, I'm pretty sure that the bluring defaults to outputting float64 arrays.  Rescaling and converting these into uint16s or even uint8s would probably help (either before or after bluring).

ark/phenotyping/preprocess.py
alex-l-kong: I'll probably just end up removing this. I think this was something specific the user needed but we don't.

alex-l-kong:
@ngreenwald @ackagel so in line with what Mike discussed, I think it's worth trying out OpenCV as an alternative. There are a few reasons for this:

- OpenCV is written almost completely in C/C++. I'm sure there's a fair amount of C code that `ndimage` relies on as well, but a lot of the functionality is also written in Python
- OpenCV doesn't require the use of a loop/`apply_along_axis` to apply the Gaussian filter across each channel
- OpenCV contains an optimized mode which provides a nice boost without any apparent extra work we need to do on our end (and I would imagine that the optimizations are based on multithreading done in C/C++, not multiprocessing in Python as we're doing right now)

ngreenwald (changes requested):
Not sure about the indices you're using for the pixel_mat coords, shouldn't one of them be the columns dimension?

I don't think we should spend more time optimizing this until you've done a complete run through the whole pipeline and we know how long each part takes. Then we can identify where the pain points are likely to be.

ark/phenotyping/preprocess.py
```diff
@@ -66,8 +64,8 @@ def create_pixel_matrix(img_xr, seg_labels, fovs=None, channels=None, blur_facto
 
         # assign metadata about each entry
         pixel_mat['fov'] = fov
-        pixel_mat['x_coord'] = np.repeat(range(img_data_blur.shape[0]), img_data_blur.shape[1])
-        pixel_mat['y_coord'] = np.tile(range(img_data_blur.shape[0]), img_data_blur.shape[1])
+        pixel_mat['x_coord'] = np.repeat(range(img_data_blur.shape[0]), img_data_blur.shape[0])
+        pixel_mat['y_coord'] = np.tile(range(img_data_blur.shape[0]), img_data_blur.shape[0])
```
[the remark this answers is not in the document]
alex-l-kong: I can fix this up. Basically, `x_coord` is supposed to indicate the row of the pixel, and `y_coord` is supposed to indicate the corresponding column of the pixel. However, this method will fail for non-square images (which I think Adam alluded to earlier). I can make the column names clearer and handle the non-square cases as well.

ngreenwald (approved):

PR #352, opened 2020-11-24
Verify preprocessing and add testing

### LINKED ISSUE #344: Testing for SOM implementation

alex-l-kong:

Addresses and closes #344. Candace's data allowed us to verify that our results are correct and make changes where needed. Once that was finished, we implemented basic test functions for the preprocessing step.

A few notable changes needed to be made for the output to completely match up:

- The dtype of img_xr needs to be float, if `int16` is used we lose many non-zero pixels due to conversion. Using `float32` instead of `float64` provides some memory relief.
- 99.9% normalization needs to be done column-wise, not row-wise, and at the very end of the process (not on each `fov` separately)

Testing is done in `preprocess_test.py`.


ark/phenotyping/preprocess.py
```diff
@@ -76,16 +78,16 @@ def create_pixel_matrix(img_xr, seg_labels, fovs=None, channels=None, blur_facto
 
         # turn into frequency
         pixel_mat.loc[:, channels] = pixel_mat.loc[:, channels].div(
-            pixel_mat.sum(axis=1), axis=0)
-
-        # 99.9% normalization
-        pixel_mat.loc[:, channels] = pixel_mat.loc[:, channels].div(
-            pixel_mat.quantile(q=0.999, axis=1), axis=0)
+            pixel_mat.loc[:, channels].sum(axis=1), axis=0)
```
ngreenwald, line 81: Adding a slightly more descriptive comment here would be useful. i.e. normalize each row by total marker counts to convert raw counts to frequencies

ark/phenotyping/preprocess.py
```diff
@@ -76,16 +78,16 @@ def create_pixel_matrix(img_xr, seg_labels, fovs=None, channels=None, blur_facto
...
 
         # assign to flowsom_data if not already assigned, otherwise concatenates
         if flowsom_data is None:
             flowsom_data = pixel_mat
         else:
             flowsom_data = pd.concat([flowsom_data, pixel_mat])
 
+    # 99.9% normalization
+    flowsom_data.loc[:, channels] = flowsom_data.loc[:, channels].div(
```
ngreenwald, line 90: @cliu72, do you do 99.9% normalization before or after converting to frequency?

ark/phenotyping/preprocess_test.py
```diff
@@ -0,0 +1,63 @@
...
+    misc_utils.verify_same_elements(flowsom_chans=flowsom_data.columns.values[:-4],
+                                    provided_chans=['chan0'])
+
+    # test on a subset of fovs and channels
+    flowsom_data = preprocess.create_pixel_matrix(sample_img_xr, sample_labels,
+                                                  fovs=['fov1'], channels=['chan0'])
+
+    misc_utils.verify_same_elements(flowsom_fovs=flowsom_data['fov'].unique(),
+                                    provided_fovs=['fov1'])
+    misc_utils.verify_same_elements(flowsom_chans=flowsom_data.columns.values[:-4],
```
ngreenwald, line 62: Some light correctness checks:
Make sure there are no blank rows in the output
Make sure that all of the rows sum to 1 (this will depend on when the 99.9% normalization happens, see above)
Generate five random row + col indices, extract the marker counts from those indices, convert to frequency, and check that they match the corresponding row in the output.

ngreenwald (changes requested):
Looks good, added some suggestions for additional testing for correctness

templates/example_flowsom_clustering.ipynb
```diff
@@ -104,7 +104,7 @@
     "if MIBItiff:\n",
```
ngreenwald, line 104: <p>This shouldn't be necessary now that you're casting to float32 within the body of the processing function, right?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/352/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

ark/phenotyping/preprocess_test.py
alex-l-kong: @ngreenwald does making sure that all the rows sum to 1 cover the case with no blank rows, or are you thinking of something different (ex. NaN does not appear)?

templates/example_flowsom_clustering.ipynb
alex-l-kong: <p>Are you referring to passing in <code>dtype='float32'</code> ? The reason why the images don't return the same values with <code>int16</code>  and <code>float32</code> <code>img_xr</code>  is likely in the way <code>gaussian_filter</code> works. All I can say for now is that loading <code>img_xr</code>  with <code>float32</code>  worked and loading it with <code>int16</code>  returned incorrect results and cut out a ton of pixels that shouldn't have been cut out.</p><p> </p><p>If you think it's worth researching into, I can.</p>

ngreenwald: ah, I see the issue now. I thought you were casting to float for the input for `gaussian_filter`, but it's actually for the output. If you cast the input to the function to `float32`, then we don't need to load the whole xarray as float32

ark/phenotyping/preprocess_test.py
ngreenwald: Good point, this check actually covers both cases

ark/phenotyping/preprocess.py
alex-l-kong: @ngreenwald looks like 99.9% normalization happens after converting to frequency.

ark/phenotyping/preprocess_test.py
alex-l-kong: @ngreenwald since 99.9% normalization happens after converting to frequency, I don't think the rows of the final dataframe will sum to 1 anymore.

templates/example_flowsom_clustering.ipynb
alex-l-kong: @ngreenwald good point, let me try that out.

ackagel:
quick q about the `dtype` issue.  If you round up using ceil and then convert to `int16`, do you still lose nonzero pixels?  Doesn't matter if memory isn't an issue here

ark/phenotyping/preprocess.py
ngreenwald: Candace, was there a reason you did it this way? I think it makes more sense to 99.9% threshold first and then convert to frequencies

alex-l-kong:
@ackagel so in your scenario, we'd first load in the image as `float32`, then round up with `ceil` and convert to `int16`?

ark/phenotyping/preprocess.py
cliu72: I did the frequency normalization first because it has the natural interpretation of calculating the normalized expression per pixel. Then the point of the 99.9% normalization is so that no single marker dominates the FlowSOM clustering. If we do 99.9% first, then the expression values for each pixel depend on the values of all the other pixels (which probably isn't an issue, but just has a less intuitive interpretation in my mind). Why do you think it makes more sense to apply 99.9% first? Since both of them are squishing the values to 0-1ish range, I don't know that it would make a big difference anyways.

cliu72: I can try it and see if it makes a difference.

ngreenwald: That makes sense. My thinking with 99.9% normalization is that the point is to find outliers within each channel. So first finding channel outliers, resetting them, and then using those corrected values to calculate the normalized pixel values. 

That way, each row will sum to 1 when being passed to the SOM.

Given the low dynamic range of the pixels, my suspicion is that this actually doesn't make a difference. But if you had a pixel that was super, super bright (i.e. well above the 99.9% threshold), if you row normalize and then 99.9% threshold, you're going to take a column that represents 99.99% of the data and modify it to be 99.9% of the data. 

If you did it the other way around, I think it would help to further reduce. But once again, given that most pixels are within 1-100, this likely won't have a big impact

alex-l-kong: @ngreenwald so to clarify, should `preprocess.py` now run the 99.9% normalization before the frequency conversion?

ackagel:
pretty much.  Doesn't need to be done if the memory/speed is fine.  Just mentioning it, since it seemed like you wanted to use `int16` but couldn't due to rounding errors.

alex-l-kong:
> pretty much. Doesn't need to be done if the memory/speed is fine. Just mentioning it, since it seemed like you wanted to use `int16` but couldn't due to rounding errors.

So I was able to replicate the results by reading the images in as `int16` but then doing the conversion to `float32` in `preprocess.py` instead. The good news is that `gaussian_filter` actually spits out results in the same `dtype` that was inputted. This, unfortunately, is why passing in the images as `int16` failed.

ark/phenotyping/preprocess.py
cliu72: Hm I'm not totally sure I understand your example. Perhaps I'm misunderstanding what a 99.9% normalization does. I interpret it not as resetting outliers, but as getting all the channels in the same dynamic range. Even if I have a pixel that is super bright for marker A (above 99.9% threshold), even after a 99.9% normalization, that pixel would still be an "outlier" (it would be way above 1). The 99.9% normalization depends on the 99.9th percentile for marker A across all the pixels. Why would it modify it from 99.99% to be 99.9% of the data?

cliu72: @alex-l-kong I would hold off for now. If we want to change this step, I want to first try it on my data and see if it really is making a difference.

ngreenwald: Agreed, lets not change anything yet

ngreenwald: Okay, I ran the current processing pipeline on Erin's TB data that you sent Alex. Here's what I get for the 99.9th percentiles 
image


Because we're row normalizing before percentile normalizing, many of the brightest markers have at least 0.1% of the pixels where they are the only signal. This means that the 99.9th percentile is 1. Therefore, dividing by the 99.9th percentile doesn't do anything for those markers.

It looks to me like the end result of this transformation is basically the increase the mean expression value of the dim markers. Perhaps z scoring/mean centering the raw channel data before processing would have a similar effect?

cliu72: In my experience, z-scoring isn't great because some of the lowly expressed markers have a mean of 0, so it's impossible for those z-scores to be below 0. The fact that some markers can have negative z-scores while others can't, seems to drive some weird clustering. If you think z scoring/mean centering will have a similar effect, is there a reason you want to do it instead of what we're doing now?

ngreenwald: I guess my main concern is that the current implementation of the 99.9 percent normalization isn't doing what we expect it to do, which would be to adjust the specific rows dominated by one marker. Instead, it's just multiplying all of the lowly expressed markers by a constant factor. It seems to be empirically working to resolve the shadow clusters, so I'm fine with keeping it. But if if we could figure out why it's useful, there might be a more accurate/better way to achieve the same result.

ngreenwald: @alex-l-kong we can skip a test for this part and leave it as is.

cliu72: Yeah I see what you mean. I guess in my mind, the 99.9 percent normalization isn't so much for adjusting specific rows dominated by one marker, but for adjusting the range of all the markers so they're approximately the same range. So for instance, if like Foxp3 is always 0.1 (or something super low), bring those values up so that we can pull out a Foxp3 cluster.

ark/phenotyping/preprocess.py
```diff
@@ -49,9 +49,10 @@ def create_pixel_matrix(img_xr, seg_labels, fovs=None, channels=None, blur_facto
 
     # iterate over fovs
     for fov in fovs:
-        # subset img_xr with only the fov we're looking for
-        img_data_blur = img_xr.loc[fov, ..., channels].values
+        # subset img_xr with only the fov we're looking for, and cast to float32
+        img_data_blur = img_xr.loc[fov, ..., channels].values.astype(np.float32)
 
+        # for each marker, compute the Gaussian blurå
```
ngreenwald, line 55: sounds exciting

ngreenwald (changes requested):
almost done

ark/phenotyping/preprocess.py
alex-l-kong: No idea how that made it in...

ark/phenotyping/preprocess.py
ngreenwald: For sure, sounds good!

PR #355, opened 2020-11-27
Extraction options

### LINKED ISSUE #157: Create additional signal_extraction options

ackagel:
## **Purpose**

Allows different extraction functions to be called from the notebooks.  Functions were already written, but couldn't be accessed by users.  Closes #157 

## **Implementation**

Users can pass a string denoting the type of extraction they want to use.  Additionally, for extractions which require additional arguments/parameters, kwargs have been added to facilitate generalized 'passing down' of these additional arguments.  The arguments of the extraction functions were made identical to facilitate general callability via dictionary in `signal_extraction`

## **Remaining issues**

Some more test functions could be added, along with an example in the segmentation notebook.  Lmk what your thoughts are on that front.

### DIFF
--- a/ark/segmentation/marker_quantification.py
+++ b/ark/segmentation/marker_quantification.py
@@ -15,7 +15,7 @@ import ark.settings as settings
 
 def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=False,
                           regionprops_features=None, split_large_nuclei=False,
-                          extraction='default', **kwargs):
+                          extraction='total_intensity', **kwargs):
     """Extract single cell protein expression data from channel TIFs for a single fov
 
     Args:
@@ -30,13 +30,18 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
         split_large_nuclei (bool):
             controls whether nuclei which have portions outside of the cell will get relabeled
         extraction (str):
-            extraction function used to compute marker counts. default is 'default'
+            extraction function used to compute marker counts. default is 'total_intensity'
         **kwargs:
             arbitrary keyword arguments
     Returns:
         xarray.DataArray:
             xarray containing segmented data of cells x markers
     """
+
+    misc_utils.verify_in_list(
+        extraction=[extraction],
+        extraction_options=list(extraction_function.keys())
+    )
 
     if regionprops_features is None:
         regionprops_features = ['label', 'area', 'eccentricity', 'major_axis_length',
@@ -164,7 +169,7 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
 
 
 def create_marker_count_matrices(segmentation_labels, image_data, nuclear_counts=False,
-                                 split_large_nuclei=False, extraction='default', **kwargs):
+                                 split_large_nuclei=False, extraction='total_intensity', **kwargs):
     """Create a matrix of cells by channels with the total counts of each marker in each cell.
 
     Args:
@@ -180,7 +185,7 @@ def create_marker_count_matrices(segmentation_labels, image_data, nuclear_counts
             boolean flag to determine whether nuclei which are larger than their assigned cell
             will get split into two different nuclear objects
         extraction (str):
-            extraction function used to compute marker counts. default is 'default'
+            extraction function used to compute marker counts. default is 'total_intensity'
         **kwargs:
             arbitrary keyword args
 
@@ -201,6 +206,11 @@ def create_marker_count_matrices(segmentation_labels, image_data, nuclear_counts
             nuclear_label='nuclear',
             compartment_names=segmentation_labels.compartments.values
         )
+
+    misc_utils.verify_in_list(
+        extraction=[extraction],
+        extraction_options=list(extraction_function.keys())
+    )
 
     misc_utils.verify_same_elements(segmentation_labels_fovs=segmentation_labels.fovs.values,
                                     img_data_fovs=image_data.fovs.values)
@@ -263,7 +273,7 @@ def create_marker_count_matrices(segmentation_labels, image_data, nuclear_counts
 
 def generate_cell_table(segmentation_labels, tiff_dir, img_sub_folder,
                         is_mibitiff=False, fovs=None, batch_size=5, dtype="int16",
-                        extraction='default', **kwargs):
+                        extraction='total_intensity', **kwargs):
     """This function takes the segmented data and computes the expression matrices batch-wise
     while also validating inputs
 
@@ -284,9 +294,9 @@ def generate_cell_table(segmentation_labels, tiff_dir, img_sub_folder,
         dtype (str/type):
             data type of base images
         extraction (str):
-            extraction function used to compute marker counts. default is 'default'
+            extraction function used to compute marker counts. default is 'total_intensity'
         **kwargs:
-            arbitrary keyword arguments
+            arbitrary keyword arguments for signal extraction
 
     Returns:
         tuple (pandas.DataFrame, pandas.DataFrame):å
@@ -303,6 +313,11 @@ def generate_cell_table(segmentation_labels, tiff_dir, img_sub_folder,
 
     # drop file extensions
     fovs = io_utils.remove_file_extensions(fovs)
+
+    misc_utils.verify_in_list(
+        extraction=[extraction],
+        extraction_options=list(extraction_function.keys())
+    )
 
     # check segmentation_labels for given fovs (img loaders will fail otherwise)
     misc_utils.verify_in_list(fovs=fovs,
--- a/ark/segmentation/marker_quantification.py
+++ b/ark/segmentation/marker_quantification.py
@@ -17,6 +17,7 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
                           regionprops_features=None, split_large_nuclei=False,
                           extraction='default', **kwargs):
     """Extract single cell protein expression data from channel TIFs for a single fov
+
     Args:
         input_images (xarray.DataArray):
             rows x columns x channels matrix of imaging data
--- a/ark/segmentation/marker_quantification.py
+++ b/ark/segmentation/marker_quantification.py
@@ -30,7 +30,7 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
             controls whether nuclei which have portions outside of the cell will get relabeled
         extraction (str):
             extraction function used to compute marker counts. default is 'default'
-        kwargs:
+        **kwargs:
             arbitrary keyword arguments
     Returns:
         xarray.DataArray:
@@ -180,7 +180,7 @@ def create_marker_count_matrices(segmentation_labels, image_data, nuclear_counts
             will get split into two different nuclear objects
         extraction (str):
             extraction function used to compute marker counts. default is 'default'
-        kwargs:
+        **kwargs:
             arbitrary keyword args
 
     Returns:
@@ -284,7 +284,7 @@ def generate_cell_table(segmentation_labels, tiff_dir, img_sub_folder,
             data type of base images
         extraction (str):
             extraction function used to compute marker counts. default is 'default'
-        kwargs:
+        **kwargs:
             arbitrary keyword arguments
 
     Returns:
--- a/ark/segmentation/marker_quantification.py
+++ b/ark/segmentation/marker_quantification.py
@@ -55,7 +55,7 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
         regionprops_features.append('label')
 
     # centroid is required
-    if 'centroid-0' not in regionprops_features and 'centroid' not in regionprops_features:
+    if not any(['centroid' in rpf for rpf in regionprops_features]):
         regionprops_features.append('centroid')
 
     # enforce post channel column is present and first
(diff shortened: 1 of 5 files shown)


ackagel:
also, thoughts on docstrings for kwargs would be appreciated; doesn't seem like there's a standard beyond what I added.

ngreenwald (changes requested):
Here's how deepcell handles kwargs to make RTD happy: https://deepcell.readthedocs.io/en/master/API/deepcell.image_generators.html

ackagel:
@alex-l-kong is the expectation that args/kwargs won't appear in the doc string?

https://github.com/angelolab/ark-analysis/blob/3580bdc61481d0645d50f7e1c0cc90bb66102bc7/docs/conf.py#L268-L291

ackagel:
nvm, I just needed to prefix with **

ackagel:
btw, the positive_pixel extraction function can actually take marker specific threshold values if given a numpy array.  There's a test covering that, but I didn't include an example in the notebook (might be hard/finicky to assure correct ordering as implemented atm).  Could change it later, but it's probably fine for now.

ark/segmentation/marker_quantification.py
```diff
@@ -27,6 +29,10 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
             morphology features for regionprops to extract for each cell
         split_large_nuclei (bool):
             controls whether nuclei which have portions outside of the cell will get relabeled
+        extraction (str):
```
ngreenwald, line 32: Can you change this to have a more informative name instead of default? Maybe pixel_sum, linear_sum, unweighted_sum, etc

ark/segmentation/marker_quantification.py
```diff
@@ -255,6 +283,10 @@ def generate_cell_table(segmentation_labels, tiff_dir, img_sub_folder,
             necessary for speed and memory considerations
         dtype (str/type):
             data type of base images
+        extraction (str):
+            extraction function used to compute marker counts. default is 'default'
+        **kwargs:
+            arbitrary keyword arguments
```
ngreenwald, line 289: We should validate that extraction belongs to a list of implemented extraction functions

ark/segmentation/marker_quantification_test.py
```diff
@@ -56,6 +56,17 @@ def test_compute_marker_counts_base():
     assert np.array_equal(segmentation_output.loc['whole_cell', :, settings.CELL_SIZE],
                           segmentation_output.loc['whole_cell', :, 'area'])
 
+    # test different extraction selection
+    center_extraction = \
+        marker_quantification.compute_marker_counts(input_images=input_images,
```
ngreenwald, line 61: I think it would be more useful to test the top level function, `generate_cell_table`, to make sure all the appropriate kwargs are getting propagated through

ark/segmentation/marker_quantification.py
```diff
@@ -27,6 +29,10 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
             morphology features for regionprops to extract for each cell
         split_large_nuclei (bool):
             controls whether nuclei which have portions outside of the cell will get relabeled
+        extraction (str):
+            extraction function used to compute marker counts. default is 'default'
+        **kwargs:
+            arbitrary keyword arguments
```
ngreenwald, line 35:
```suggestion
           keyword arguments for signal extraction
```

ark/segmentation/signal_extraction.py
```diff
@@ -36,16 +39,20 @@ def center_weighting_extraction(cell_coords, image_data, centroid):
...
+        **kwargs:
+            arbitrary keyword arguments
 
     Returns:
         numpy.ndarray:
             Sums of counts for each channel
     """
 
+    # check for centroid in kwargs
+    if 'centroid' not in kwargs.keys():
```
ngreenwald, line 51: This won't ever happen though, right? There's no user input here, it gets called internally

templates/Segment_Image_Data.ipynb
```diff
@@ -358,7 +358,29 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Afterwards, we can generate expression matrices from the labeling + imaging data"
+    "### Afterwards, we can generate expression matrices from the labeling + imaging data\n",
```
ngreenwald, line 361: I think this is a fairly niche use case, at least for now. Did Tyler want it for some reason? I think it's just going to add confusion if we include it in the main notebook

ngreenwald (changes requested):

ark/segmentation/signal_extraction.py
ackagel: True.  Just auto-piloted that one haha. Thought it might help catch bugs if we end up dev-ing but accessing a non-existent argument in a dictionary already throws an error, so it's just redundant.

templates/Segment_Image_Data.ipynb
ackagel: Hadeesha wanted the positive pixel extraction.  Makes extra sense for exotic (eg. brain) panels where some markers might not stain well (making the intensity differences between channels less directly comparable).  But I do agree that it looks kinda out of place in the main notebook (at least the way I wrote it this pass).

ngreenwald: For sure. If he has questions you can point him to the documentation to see how to get it to work

ark/segmentation/marker_quantification.py
```diff
@@ -27,11 +29,20 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
...
+            extraction function used to compute marker counts. default is 'total_intensity'
+        **kwargs:
+            arbitrary keyword arguments
     Returns:
         xarray.DataArray:
             xarray containing segmented data of cells x markers
     """
 
+    misc_utils.verify_in_list(
+        extraction=[extraction],
```
alex-l-kong, line 42: You don't need to put `extraction` in a list for `verify_in_list` to work.

ark/segmentation/marker_quantification.py
```diff
@@ -27,11 +29,20 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
...
+        extraction_options=list(extraction_function.keys())
+    )
+
     if regionprops_features is None:
         regionprops_features = ['label', 'area', 'eccentricity', 'major_axis_length',
                                 'minor_axis_length', 'perimeter', 'centroid']
```
alex-l-kong, line 48: Candidate for `ark.settings`?

ark/segmentation/marker_quantification.py
```diff
@@ -176,6 +207,11 @@ def create_marker_count_matrices(segmentation_labels, image_data, nuclear_counts
             compartment_names=segmentation_labels.compartments.values
         )
 
+    misc_utils.verify_in_list(
+        extraction=[extraction],
```
alex-l-kong, line 211: See above

alex-l-kong (changes requested):
A few minor changes here and there. About **kwargs in docstrings: RTD parses them really weirdly (they don't appear in the param list but do in the arg list, or the other way around, which messes up one of the verifications), so I explicitly removed them from appearing on RTD for the time being. I know some library documentation tells you where to find the arguments that `**kwargs` can take without explicitly specifying them. We can definitely discuss what to do with these later.

ark/segmentation/marker_quantification.py
ackagel: I'll try it without that, but my intuition told me that python will regard that string as a list of characters.

ark/segmentation/marker_quantification.py
ackagel: Used to be in `ark.settings` before being merged in.  Removed them since the names won't change and it's doubtful that we'll want to include any other props by default.  Not a hassle to put it back in, but that was last I heard on that front.

ngreenwald: I think it's fine to leave it defined within this function, since it's only ever going to be used in here, not called anywhere else.

ark/segmentation/marker_quantification.py
```diff
@@ -157,6 +184,10 @@ def create_marker_count_matrices(segmentation_labels, image_data, nuclear_counts
         split_large_nuclei (bool):
             boolean flag to determine whether nuclei which are larger than their assigned cell
             will get split into two different nuclear objects
+        extraction (str):
+            extraction function used to compute marker counts. default is 'total_intensity'
```
ngreenwald, line 188: If default is defined in function def, don't need to be put default in docstring as well

ark/segmentation/signal_extraction.py
```diff
@@ -57,14 +60,16 @@ def center_weighting_extraction(cell_coords, image_data, centroid):
     return channel_counts
 
 
-def default_extraction(cell_coords, image_data):
+def default_extraction(cell_coords, image_data, **kwargs):
```
ngreenwald, line 63: rename to total intensity

ark/segmentation/signal_extraction.py
```diff
@@ -1,31 +1,34 @@
...
         image_data (xarray.DataArray): array containing channel counts
-        threshold (int): where we want to set the cutoff for a positive pixel, default 0
+        **kwargs: arbitrary keyword arguments
 
     Returns:
         numpy.ndarray:
             Sums of counts for each channel
     """
 
+    if 'threshold' not in kwargs.keys():
```
ngreenwald, line 19: So I have a slight preference for the pattern that we use in deepcell for handling functions with kwargs. 

```
compute_marker_counts(extraction='total_intensity', **kwargs)
     
    if extraction == 'total_intensity':
              cell_counts = total_intensity(image_data)

    elif extraction == 'positive_pixels`:
                threshold = kwargs.get('threshold', 0)
                cell_counts = positive_pixels(image_data, threshold=threshold)

    elif extraction == 'center_weighting'
                 centroid = kwargs.get(centroid)
                  cell_counts = center_weighting(image_data, centroid=centroid)
```
The benefit of this is that we don't need the helper functions to get kwargs passed to them, their API is totally defined. This makes it less likely something will go run, and makes it more understandable to read what's happening. It also means that all the error checking happens in the top level function. 

The downside is that you need a if/else condition set up for all of the potential extraction options. I think the tradeoff is worth it, but I'm also open to your approach if you think its better

ngreenwald (changes requested):
Nice, one refactor suggestion, open to discuss though

ark/segmentation/signal_extraction.py
ackagel: I'd lean towards my implementation, mostly because it consolidates all necessary changes for an additional extraction to the `signal_extraction.py` file (and also because I think mapping generalized function pointers in a dictionary is cool and efficient).

I think the deepcell approach makes sense for larger functions with less, immediately understandable function bodies, where clear API instructions/docstrings are requisites.  If we end up including a complicated extraction function, I'd hop on board for that approach.

That being said, I'll probably switch to using `.get` for accessing kwargs (I didn't know it was a thing haha).

ark/segmentation/marker_quantification.py
```diff
@@ -43,6 +54,10 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
     if 'label' not in regionprops_features:
         regionprops_features.append('label')
 
+    # centroid is required
+    if 'centroid-0' not in regionprops_features and 'centroid' not in regionprops_features:
```
ngreenwald, line 58: Why are we checking for centroid-0 and centroid? Centroid-0 is the output for the first coord, centroid is the input name

ark/segmentation/signal_extraction.py
```diff
@@ -1,15 +1,15 @@
 import numpy as np
 
 
-def positive_pixels_extraction(cell_coords, image_data, threshold=0):
+def positive_pixels_extraction(cell_coords, image_data, **kwargs):
     """Extract channel counts by summing over the number of non-zero pixels in the cell.
 
-    Improves on default_extraction by not distinguishing between pixel expression values
+    Improves on total_intensity_extraction by not distinguishing between pixel expression values
```
ngreenwald, line 7: I think we can delete this little bit of editorializing

ngreenwald (changes requested):
Just two small tweaks

ark/segmentation/marker_quantification.py
alex-l-kong: Right, I forgot it was a string.

alex-l-kong (commented):
Not much else to add on top of Noah's review.

ark/segmentation/marker_quantification.py
ackagel: I think past Adam assumed centroid-0 and centroid-1 could get passed as `regionprops_features`, noticed that just a check for 'centroid' would miss that case and consequently create two copies of 'centroid-0' and 'centroid-1' later on, so he lazily compensated with a check for 'centroid-0'.

Basically a hacky 'fix' to a rare problem.  Less hacky fix would be to check for 'centroid' in any element of regionprops_features

ark/segmentation/signal_extraction.py
ackagel: Thoughts on 'A New York Times best extractor, it's the highly anticipated sequel to **total_intensity_extraction**'?

ngreenwald (approved):

Issue #356, opened 2020-11-30
segmentation_utils.visualize_segmentation-file path does not exist


HoxB8:
I created deepcell input files from several multi-channel tiffs.  Then I tried to enter multiple deepcell output tiff files into the post-processing script.  I am able to get the .xr file but when I try to run the segmentation_utils.visualize_segmentation, I get an error "File Path Does Not Exist".  I double-checked my file directory structure and made sure I had the deepcell_visualization directory in place and no errors in my viz_dir (which is  os.path.join(base_dir, "deepcell_visualization").   

**What have you tried so far?**
I've tried to update with "git pull" and "bash start_docker.sh --update".  I also made sure my file names are pretty simple like fov1_adjnorm.tiff.  
![Screen Shot 2020-11-29 at 5 03 52 PM](https://user-images.githubusercontent.com/58482675/100559051-e6c9c280-3265-11eb-91fd-5043bb1a6adb.png)

I'd appreciate any help on this.

alex-l-kong:
@HoxB8 can you first test if the file path exists in the notebook by calling `os.path.exists(viz_dir)`?

HoxB8:
Yes, I tried that and it returns True.  Also tried typing in the entire path directly, but still happens.  
![Screen Shot 2020-11-30 at 11 44 47 PM](https://user-images.githubusercontent.com/58482675/100711494-35f71c80-3366-11eb-93a0-d99a16910e6b.png)

alex-l-kong:
@HoxB8 are you on a Windows machine by any chance? In the past, we've had Windows users run into issues with file paths in the past.

HoxB8:
I'm using MacOS 10.15.7 with Python version 3.8.5.  I was thinking it might be something with a Mac as I had weird bugs that didn't seem normal.  Oddly, my colleague was able to run the script without issues, so maybe it's a software version issue.

PR #193, opened 2020-08-21
Integrated plotting function

william14641:

PR #357, opened 2020-12-01
Fix bugs in visualize segmentation

### LINKED ISSUE #284: Error in Segment Image Data notebok when making overlay using > 3 total nuclear and membrane channels

### LINKED ISSUE #354: segmentation_utils.visualize_segmentation error

**Describe the bug**
When I run the segmentation_utils.visualize_segmentation cell, I get the error attached at the bottom of this write-up. As for the output files from this cell, most points have all 3 files, one point only produces the DNA_overlay file (Point12), and some points don't have any output files. The deepcell output folder before this step has all the points.

**To Reproduce**

Run the downstream cells using the output files of Deepcell.
[491defc4f7884fa5b876860476d2e78a.zip](https://github.com/angelolab/ark-analysis/files/5598258/491defc4f7884fa5b876860476d2e78a.zip)

**Expected behavior**

All 164 points should have a DNA_overlay, segmentation_borders, and segmentation_labels in the deepcell_visualization folder

**Screenshots**
image

alex-l-kong:

We've received a lot of bugs that have resulted from merging #193 in. This PR encompasses the following 3 issues:

- Closes #284 
- Closes #354 
- #356 is definitely an outlier, I don't think it's worth addressing in this PR

It seems that the bugs aren't completely related to `visualize_segmentation` but rather on what's being passed in. `data_xr` instead of `data_xr_summed`, for example, was incorrectly passed in, which likely caused @awedwards issue.

`chan_list` verification logic was a little buggy, so that was improved as well.


alex-l-kong:
So this method works to fix `visualize_segmentation` in the notebook for now (which should address at least @awedwards issue), but the conflicts between `channels` and `compartments` is definitely a minor headache. Would be helpful to discuss alternatives here.

templates/Segment_Image_Data.ipynb
```diff
@@ -344,13 +344,13 @@
    ],
```
ngreenwald, line 344: <p>We should just change xr_dim_name from compartments to channels. Not sure why it was set to compartments here. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/357/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

ngreenwald, line 344: <p>Instead of indexing into data_xr_summed to get the FOVs list, I think we should create a variable called display_fov and set it equal to "Point8", and then pass that into visualize_segmentation</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/357/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

ngreenwald (changes requested):
Looks good. Can you update the description with which of the issues this is the closing?

The thing I don't like about `visualize_segmentation` is that there are two separate things people will want to do. 

1. Cycle through different example images in the notebook, seeing how the overlays look
2. Save the segmentation output for all FOVs in their dataset. 

Right now, they're combined together into the same function. This is partly what lead to the show=true errors. What do you think the best way to split these up would be?

alex-l-kong:
> Looks good. Can you update the description with which of the issues this is the closing?
> 
> The thing I don't like about `visualize_segmentation` is that there are two separate things people will want to do.
> 
> 1. Cycle through different example images in the notebook, seeing how the overlays look
> 2. Save the segmentation output for all FOVs in their dataset.
> 
> Right now, they're combined together into the same function. This is partly what lead to the show=true errors. What do you think the best way to split these up would be?

@ngreenwald I think the best way to do this is to have `plot_overlay` only have `imshow` calls and `visualize_segmentation` only have `imsave` calls. Right now, it's also possible for `plot_overlay` to save an image and I agree, it's really confusing. We can accomplish this by simply returning the result of what would normally be saved by `plot_overlay`, then allow `visualize_segmentation` to save it for us.

This is great for making the functionality clearer, but in terms of the issues people have been bringing up, I don't think it's actually `show=True` that's causing the problem anymore, but rather other errors (ex. `data_xr` and not `data_xr_summed`, quite possibly user-specific data issues and filesystem errors, etc.) that just happen to be seen by this particular call to `plot_overlay` first. I'm doing more investigating into these to see what the root problems are, but I do believe something is getting propagated into the call to `plot_overlay` with `show=True` (especially for #354).

ngreenwald:
The `get_topmost_subplot` error happens when you create two different plots in the same plotting window. This is happening because of the for loop, where each FOV is getting imshowed. The error goes away if only a single FOV is generated. 

Okay great, I agree. I think as a further simplification, we don't even need `visualize_segmentation` to call plot_overlay for generating the RGB overlay. We can just have a separate cell that directly calls plot_overlay to generate the overlay people care about (the channel overlay). 

Once they're done with that, then we can have them call visualize_segmentation (which we'll call something else). That function will loop through, and save the output files for all of the fovs in the xarray

alex-l-kong:
@ngreenwald a few questions:

1. The call to `plot_overlay` will display only the result for a single `fov` at a time, as opposed to multiple at a time?
2. With this new method, looks like we're only going to be saving `_segmentation_borders.tiff` and `_segmentation_labels.tiff` files?
3. Have we already tried using `plt.imshow` instead of `io.imshow`? I think `plt.imshow` could allow us to display multiple channel overlays at once, although that would be a problem if the user passed in 100 `fovs`.

ngreenwald:
I think we should move all of the stuff that's in `plot_overlay` which doesn't involve plotting a multi-color overlay and move it to the `visualize_segmentation` function, which we can rename to `save_segmentation_output` or something like that.

Then there will be one cell where the user calls plot_overlay directly, with a single FOV specified, to look at the output. And a separate cell which loops over all of the data and produces the borders.

ngreenwald:
Even better, we should modifiy plot_overlay to be `create_overlay`. Then it can return the image to displayed in the notebook. `save_segmentation_output` can then also call create_overlay, but without displaying the image each time.

ark/utils/plot_utils.py
```diff
@@ -11,15 +11,55 @@
...
+
+    if len(plotting_tif.shape) == 2:
+        if plotting_tif.shape != predicted_contour.shape:
+            raise ValueError("plotting_tif and predicted_contour array dimensions not equal.")
+        else:
+            # convert RGB image with same data across all three channels
+            plotting_tif = np.stack((plotting_tif, plotting_tif, plotting_tif), axis=2)
+
+    elif len(plotting_tif.shape) == 3:
+        blank_channel = np.zeros(plotting_tif.shape[:2] + (1,), dtype=plotting_tif.dtype)
```
ngreenwald, line 36: This entire thing can be replaced with
```
formatted_tif = np.zeros((plotting_tif.shape[0], plotting_tif.shape[1], 3))
formatted_tif[..., :plotting_tif.shape[2]] = plotting_tif
```

```

File: ark/utils/plot_utils.py [hunk 3]
```diff
@@ -28,52 +68,21 @@ def plot_overlay(predicted_contour, plotting_tif, alternate_contour=None, path=N
             2D or 3D numpy array of imaging signal
         alternate_contour (numpy.ndarray):
             2D numpy array of labeled cell objects
-        path (str):
-            path to save the resulting image
-        show (bool):
-            whether or not to show plot
     """
 
+    # just plot the outlines
```
ngreenwald, line 73: We should move this functionality (creating just the outlines) to `save_segmentation_labels`, since we'll want to do this for every FOV. Instead lets require a plotting tif for this function

ark/utils/plot_utils.py
```diff
@@ -28,52 +68,21 @@ def plot_overlay(predicted_contour, plotting_tif, alternate_contour=None, path=N
...
 
     # define borders of cells in mask
     predicted_contour_mask = find_boundaries(predicted_contour,
                                              connectivity=1, mode='inner').astype(np.uint8)
     predicted_contour_mask[predicted_contour_mask > 0] = 255
 
     if plotting_tif is None:
-        # will just save the contour mask
-        io.imsave(path, predicted_contour_mask)
+        return predicted_contour_mask
```
ngreenwald, line 85: We should move this functionality (creating just the outlines) to `save_segmentation_labels`, since we'll want to do this for every FOV

ark/utils/segmentation_utils.py
```diff
@@ -183,8 +183,7 @@ def concatenate_csv(base_dir, csv_files, column_name="fov", column_values=None):
     combined_data.to_csv(os.path.join(base_dir, "combined_data.csv"), index=False)
 
 
-def visualize_segmentation(segmentation_labels_xr, channel_data_xr,
-                           output_dir, chan_list=None, fovs=None, show=False):
+def save_segmentation_labels(segmentation_labels_xr, output_dir, fovs=None):
```
ngreenwald, line 186: We should add in a call to create_overlay and save the resulting RGB image to disk, without every displaying it

ark/utils/segmentation_utils.py
```diff
@@ -193,41 +192,30 @@ def visualize_segmentation(segmentation_labels_xr, channel_data_xr,
...
-                path=os.path.join(output_dir, save_path), show=show
-            )
-        # Generates segmentation borders and labels
-        plot_utils.plot_overlay(
+
+        # save the labels respectively
+        io.imsave(os.path.join(output_dir, f'{fov}_segmentation_labels.tiff'), labels)
+
+        # generate segmentation borders and labels
+        contour_mask = plot_utils.create_overlay(
```
ngreenwald, line 216: We can pull this functionality out directly, rater than using create_overlay to do it, since it's just a call to skimage.find_boundaries

ngreenwald (changes requested):
Looks good. Just a few last remaining skeletons to remove from the plot function that can be moved into the segmentation_labels function, so that all the plot function is responsible for is taking an image and creating an RGB overlay

ark/utils/segmentation_utils.py
alex-l-kong: A minor problem here is that `create_overlay` needs the result of `find_boundaries` to figure out which values in `rescaled` to set to white. Two ways around this:

1. Live with duplicated code in `create_overlay`
2. Add an argument `predicted_contour_mask` to `create_overlay` so we can pass in the result of `find_boundaries` in `save_segmentation_labels` in (however, the `find_boundaries` logic now needs to be duplicated in the notebook call to `create_overlay`)

ngreenwald: Yea, it's fine to call `find_boundaries` twice, it's very fast

ngreenwald: I'd just have both functions calculate it

ark/utils/plot_utils.py
```diff
@@ -11,15 +11,49 @@
 from skimage.exposure import rescale_intensity
 
 
-# plotting functions
+def preprocess_tif(predicted_contour, plotting_tif):
```
ngreenwald, line 14: Maybe preprocess_tif_for_overlay? Or something slightly more specific

ark/utils/plot_utils.py
```diff
@@ -11,15 +11,49 @@
...
+
+
+def create_overlay(predicted_contour, plotting_tif, alternate_contour=None):
     """Take in labeled contour data, along with optional mibi tif and second contour,
     and overlay them for comparison"
 
-    Plots the outline(s) of the mask(s) as well as intensity from plotting tif.  Predicted
-    contours are plotted in red, while alternate contours are plotted in white.  Plot is
-    saved as a TIF file in provided file path, if specified.
+    Generates the outline(s) of the mask(s) as well as intensity from plotting tif. Predicted
```
ngreenwald, line 55: typo here. Also, top-level description should be updated. I think segmentation_label is more clear than labeled contour data

ark/utils/plot_utils.py
```diff
@@ -28,86 +62,44 @@ def plot_overlay(predicted_contour, plotting_tif, alternate_contour=None, path=N
...
+            rescaled_intensity = rescale_intensity(plotting_tif[:, :, idx],
+                                                   in_range=(percentiles[0], percentiles[1]),
+                                                   out_range='uint8')
+            rescaled[:, :, idx] = rescaled_intensity
+
+    # overlay first contour on all three RGB, to have it show up as white border
+    rescaled[predicted_contour_mask > 0, :] = 255
+
+    # overlay second contour as red outline if present
+    if alternate_contour is not None:
```
ngreenwald, line 93: Can you try plotting this? I'm pretty sure I coded this wrong. It looks like the second overlay is red, but the DNA channel is also getting put in as red (I think)

ark/utils/segmentation_utils.py
```diff
@@ -183,8 +184,8 @@ def concatenate_csv(base_dir, csv_files, column_name="fov", column_values=None):
     combined_data.to_csv(os.path.join(base_dir, "combined_data.csv"), index=False)
 
 
-def visualize_segmentation(segmentation_labels_xr, channel_data_xr,
-                           output_dir, chan_list=None, fovs=None, show=False):
+def save_segmentation_labels(segmentation_labels_xr, channel_data_xr, output_dir,
+                             fovs=None, channels=None):
```
ngreenwald, line 188: Docstring still references chan_list

ark/utils/segmentation_utils.py
```diff
@@ -194,40 +195,44 @@ def visualize_segmentation(segmentation_labels_xr, channel_data_xr,
...
 
+    # assign fovs and channels to everything if None
     if fovs is None:
         fovs = segmentation_labels_xr.fovs.values
-    misc_utils.verify_in_list(fovs=fovs, segmentation_label_fovs=segmentation_labels_xr.fovs)
 
-    if chan_list is not None:
-        misc_utils.verify_in_list(channels=chan_list,
-                                  channel_data_channels=channel_data_xr.channels)
+    if channels is None:
```
ngreenwald, line 211: If channels is None we shouldn't generate the overlay

ngreenwald (changes requested):
Nice, almost there

ark/utils/plot_utils.py
alex-l-kong: I can double check, but it looks correct: you're setting the first contour as white by setting all 3 respective channel values as 255. The second contour will be red because you're only setting the first channel as 255, but explicitly set the other 2 channels as 0.

ngreenwald: Right, I'm saying that the plotting_tif is going to assign DNA to be red. So if we have borders and DNA as red, it won't be that useful

alex-l-kong: Oh ok, I misunderstood. I suppose we can simply assign`alternate_contour_mask` to be a different color, perhaps blue?

ngreenwald: Because we index into rescaled, the first channel of plotting tif gets assigned to the first channel of rescaled, and the channel order goes RGB. I think it makes more sense for the contour overlay to be red and the DNA to be blue/green

alex-l-kong: @ngreenwald oh OK, so if I'm understanding correctly, this happens in `tif_overlay_preprocess` when we assign `plotting_tif` to take up just the first channel of `formatted_tif` if the third dimension of `plotting_tif` only contains 1 channel. If so, I can just assign it to a different channel. A clarification, though: if `plotting_tif` contains 2 channels, it will be assigned to the red and green channels of `formatted_tif`. Should we leave it like this or assign it differently?

ngreenwald: As long as we're already raising an informative ValueError, and it's not a numpy error, that's fine. 

Single channel input: assigned to blue
Two channel input: first channel assigned to blue, second channel assigned to green
three channel input: first channel assigned to blue, second channel assigned to green, third channel assigned to red

First overlay assigned to white
Second overlay assigned to red

alex-l-kong:
@ngreenwald OK, the new channel structure has been added, ready for review.

ark/utils/plot_utils.py
```diff
@@ -11,103 +11,100 @@
...
+        numpy.ndarray:
+            The preprocessed image
+    """
+
+    if len(plotting_tif.shape) == 2:
+        if plotting_tif.shape != segmentation_labels.shape:
+            raise ValueError("plotting_tif and segmentation_labels array dimensions not equal.")
+        else:
+            # convert RGB image with same data across all three channels
+            formatted_tif = np.stack((plotting_tif, plotting_tif, plotting_tif), axis=2)
```
ngreenwald, line 33: this looks like the same logic as before for assigning channels a color. See my previous comment for the desired mapping between number of channels and which colors they should get.

ark/utils/segmentation_utils_test.py
```diff
@@ -205,69 +205,80 @@ def test_transform_expression_matrix_multiple_compartments():
...
     with tempfile.TemporaryDirectory() as temp_dir:
-        segmentation_utils.visualize_segmentation(
-            segmentation_labels_xr=segmentation_labels_xr,
-            fovs=None, channel_data_xr=channel_xr,
-            output_dir=temp_dir)
+        segmentation_utils.save_segmentation_labels(segmentation_labels_xr=segmentation_labels_xr,
+                                                    channel_data_xr=channel_xr,
+                                                    output_dir=temp_dir)
+
+        # make sure all files exist
```
ngreenwald, line 220: Can also add a check that no overlay.tiff is created

ngreenwald (changes requested):
Close, just need to update the colors of the input data

ark/utils/plot_utils.py
```diff
@@ -11,103 +11,103 @@
...
+            formatted_tif = np.zeros((plotting_tif.shape[0], plotting_tif.shape[1], 3),
+                                     dtype=plotting_tif.dtype)
+            formatted_tif[..., 2] = plotting_tif
+    elif len(plotting_tif.shape) == 3:
+        # can only support up to 3 channels
+        if plotting_tif.shape[2] > 3:
+            raise ValueError("max 3 channels of overlay supported, got {}".
+                             format(plotting_tif.shape))
+
+        # set first n channels (in reverse order) of formatted_tif to plotting_tif
```
ngreenwald, line 42: clever!

ngreenwald (approved):
Nice stuff

PR #358, opened 2020-12-01
Default heatmap

### LINKED ISSUE #205: Default heatmap visualization
The current heatmap visualization function `visualize_z_scores` is not robust and extensible to different datatypes. 

A generalized heatmap function that is more flexible. It should take in a matrix of data to be visualized, a set of labels for the axes, and a colormap, and produce an annotated heatmap as output.

Additionally, the heatmap should support the following:

- An optional `dpi` parameter for resolution control if the user wants to save the figure
- An optional `center_val` parameter identifying which value to center the colorbar around

### LINKED ISSUE #294: Add heatmap visualization for num_cell_type_per_marker and mean_marker_exp_per_cluster for neighborhood analysis

After the count/mean tables per cell type/marker respectively have been created, we should provide a corresponding heatmap visualization to allow the user a preliminary overview of this information. This addresses the second TODO in `example_neighborhood_analysis_script.ipynb`.

**Additional context**

The TODO is a bit confusing: `all_data_cluster_labeled` should be `num_cell_type_per_marker`.

If taking on this issue, please address #205 as well.

william14641:

Closes #205 and closes #294. Generalize `visualize_z_scores` to `draw_heatmap` in `visualize.py` to visualize any heatmap. Function takes in data, labels for x and y axes, and optional centerval, dpi, colormap parameters.

We also add the optional `dpi` parameter to other visualize functions, as well as to `save_figure`. Finally, add the heatmap visualizations in `example_neighborhood_analysis` and `example_spatial_analysis` notebooks.


ark/analysis/visualize_test.py
```diff
@@ -65,15 +65,17 @@ def test_visualize_z_scores():
...
         # test that without save_dir, we do not save
-        visualize.visualize_z_scores(z, pheno_titles)
+        visualize.visualize_z_scores(z, pheno_titles, pheno_titles,
+                                     center_val=np.random.ranf())
         assert not os.path.exists(os.path.join(temp_dir, "z_score_viz.png"))
 
         # test that with save_dir, we do save
-        visualize.visualize_z_scores(z, pheno_titles, save_dir=temp_dir)
+        visualize.visualize_z_scores(z, pheno_titles, pheno_titles,
+                                     center_val=np.random.ranf(), save_dir=temp_dir)
```
alex-l-kong, line 78: You should also include tests for the following:

- `center_val=None`
- `dpi` is explicitly set (currently, we're only testing `dpi=None` cases)

In both cases, you only need to include tests with a save_dir.

templates/example_spatial_analysis_script.ipynb
```diff
@@ -149,13 +149,13 @@
...
     "pheno_titles = all_data[\"cell_type\"].drop_duplicates()\n",
-    "visualize.visualize_z_scores(stats_cluster.loc['fov6', \"z\", :, :].values, pheno_titles)\n",
+    "visualize.visualize_z_scores(stats_cluster.loc['fov6', \"z\", :, :].values, pheno_titles, pheno_titles)\n",
     "# For channel spatial analysis, first find all the marker titles to use as input for the clustergram\n",
     "channel_start = np.where(all_data.columns == settings.PRE_CHANNEL_COL)[0][0] + 1\n",
     "channel_end = np.where(all_data.columns == settings.POST_CHANNEL_COL)[0][0]\n",
     "\n",
     "marker_titles = all_data.iloc[:, channel_start:channel_end].drop(excluded_channels, axis=1).columns\n",
-    "visualize.visualize_z_scores(stats_channel.loc['fov6', \"z\", :, :].values, marker_titles)"
+    "visualize.visualize_z_scores(stats_channel.loc['fov6', \"z\", :, :].values, marker_titles, marker_titles)"
```
alex-l-kong, line 158: Two things:

- Split up this cell into cluster spatial analysis and channel spatial analysis visualizations
- Can you display the resulting visualization for both cells? Currently, it only saves them.

alex-l-kong (changes requested):
Good start. Let's work on:

- Simplifying the `dpi` saving logic (no need to check if `None`) and moving it into `misc_utils.save_figure`.
- Making your tests more complete
- Displaying the visualizations in the notebook (and general tidying up)

ark/analysis/visualize.py
```diff
@@ -63,14 +62,23 @@ def draw_boxplot(cell_data, col_name, col_split=None, split_vals=None, save_dir=
         misc_utils.save_figure(save_dir, "boxplot_viz.png")
 
 
-def visualize_z_scores(z, pheno_titles, save_dir=None):
+def visualize_z_scores(z, x_labels, y_labels, dpi=None, center_val=None,
+                       colormap="vlag", save_dir=None):
```
alex-l-kong, line 66: One more thing I forgot: let's call this something different, since this function will be used for general heatmap plotting. Something like: `draw_heatmap`.

alex-l-kong (changes requested):
One more comment below.

ngreenwald:
The original reason for this function was before we knew that clustermap existed. Is there any reason we need to wrap this in our own function, rather than just calling clustermap directly?

alex-l-kong:
> The original reason for this function was before we knew that clustermap existed. Is there any reason we need to wrap this in our own function, rather than just calling clustermap directly?

The main reason is that the function also needs to handle saving the figure as well, which we've encapsulated in `misc_utils.save_figure`. Another reason is that there is at least one line of code that handles the formatting of the DataFrame to visualize beforehand, though we could extract that out into the notebook if needed (however, we'd also have to extract the saving of the figure to the notebook as well in that case).

ngreenwald:
Okay yeah, that makes sense.

ark/analysis/visualize_test.py
alex-l-kong: Now that I think about it, this isn't necessary as these are just basic visualization control parameters that only `matplotlib` needs to worry about.

templates/example_neighborhood_analysis_script.ipynb
```diff
@@ -2,7 +2,7 @@
  "cells": [
```
ngreenwald, line 2: <p>We should add a parameter that controls whether the value is overlaid over the heatmap or not, and set the default to False</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/358/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

ngreenwald (changes requested):
For some reason, test_draw_boxplot and test_draw_heatmap are taking a long time to run. Can you see what's going on there?

alex-l-kong:
@ngreenwald the problem with these functions (`test_draw_heatmap` and `test_draw_boxplot`) is they take a very long time to actually start running on Travis, and the reason is because there are a ton of visualization tests to setup. Granted, there will always be a little lag before the function actually starts running, but for these functions it's around 4-5 seconds, which is several orders longer than other visualization functions.

I was able to get both functions down to a minimal running time removing several tests. We don't need a lot of these for `test_draw_heatmap` since these just test basic `clustermap` parameters passed into `seaborn` that we know work. However, the only test I've left in `test_draw_boxplot` is the one that tests all the parameters at once, so we lose a couple tests where we leave certain defaults in place.

ngreenwald (changes requested):
Looks great, only thing is to modify plotting in spatial_analysis to hide the values in the heatmap by default, similar to what you did for neighborhood analysis

ngreenwald (approved):

PR #366, opened 2020-12-03
Add section for xarrays in developer section

### LINKED ISSUE #209: Add xarray explainer to docs
Once we have a developer section of the docs (#202), something to include would be a description of how xarrays work, common indexing use-cases, as well as our overall data model

alex-l-kong:

Addresses and closes #209. `xarray` is a library that not many people will be as familiar with, so we add a section in `virtualenv.md` explaining this.

**Remaining issues**

I'm not really liking the name `virtualenv.md` nor the title that displays on RTD (Setting Up Your Virtual Environment) now that this section is going to be for general development. There are two ways we could do this:

- Change the file name/title/structure of `virtualenv.md`
- Put the section on `xarray` into a separate `.md` file


ngreenwald:
Agreed, lets change the name

ngreenwald (changes requested):
I think "Setting up ark-analysis for development" and "Using ark functions directly" shouldn't be nested under the Virtual_Env heading. They work no matter how the env was set up, so it makes sense for them to be their own section. A few typos in the xarray section, but the content looks great

ngreenwald (approved):

PR #367, opened 2020-12-04
Mibilib workaround

ackagel:

`mibilib` has strict package requirements, which can create problems when pip has to resolve multiple versions.   This PR prevents mibilib from installing it's strict dependencies.

separates `mibilib` requirement in 'nodeps' file, which is installed with the `--no-deps` flag.


Dockerfile
```diff
@@ -6,8 +6,9 @@ RUN apt-get update && apt-get install -y gcc
 WORKDIR /scripts
 
 # copy over the requirements.txt, install dependencies, and README
-COPY setup.py requirements.txt README.md /opt/ark-analysis/
+COPY setup.py requirements.txt requirements-nodeps.txt README.md /opt/ark-analysis/
 RUN pip install -r /opt/ark-analysis/requirements.txt
+RUN pip install --no-deps -r /opt/ark-analysis/requirements-nodeps.txt
```
ngreenwald, line 11: Just a brief comment for what this does

ngreenwald (changes requested):

alex-l-kong (commented):
Looks great! Can we do something similar for `kiosk-client` and other `deepcell` dependencies?

ackagel:
I think we only really need to do this if/when there are dependency conflicts that don't cause actual codebase problems but will refuse to coexist.  Idk if that's happening or not with those, but it seems like we're good for now.

alex-l-kong (commented):
Great!

ngreenwald (approved):

PR #368, opened 2020-12-10
Boost clustering process (aka integrate R SOM into clustering pipeline)

### LINKED ISSUE #363: Batch processing for trained SOM
The current process for running the trained SOM is fairly slow (#345). This is in part because we have to perform individual operations for every pixel. I think we could do some linear algebra to create a batch version of this, and run multiple pixels through the trained SOM at once. This should substantial improvements in speed.

alex-l-kong:

Closes #363. One of the bottlenecks that we're currently facing is in the cluster assignment step; we do this row-by-row, and it doesn't need to be this way. This PR will transition this to a batch-processing approach.

We have to play around with `numpy` broadcasting because many of these operations require reshaping arrays and expanding dimensions. `numpy` broadcasting allows us to process thousands, even millions of rows at once, although there are diminishing returns the larger the batch size becomes.

UPDATE: we will not be using our own implementation anymore, but rather R's implementation. It runs hundreds of miles faster due to pre-compiled C-code and much better memory management than MiniSOM. It's also easy to integrate as R actually plays fairly nice with Docker.

**Remaining issues**

With a batch size of 1000000, we see a 1/3 speed improvement as opposed to going row-by-row, which is good, but it might not be good enough considering Candace's dataset is actually on the smaller side. There are a few things we should consider:

- Reducing the number of operations: currently, I have a ton of reshapes and swapaxes due to the way broadcasting works in `numpy`. However, we can potentially reduce some of these. Feel free to tackle these low-hanging fruits in the comments section!
- Calculating an optimal batch size to use. I'm not sure if this is possible, but if you know of a way, please let me know!
- Multithreading/multiprocessing: if we want an even larger boost, we could consider batching the processes in parallel as well. Note that this method will be dependent on the number of cores the user has and could lead to several kernels dying due to its memory intensity. However, if we've hit a lower bound in terms of time with just `numpy` broadcasting, we can consider 2x-ing or 4x-ing (in an ideal world) this process with parallelization.

UPDATE: with R SOM integration, we no longer need to worry about these.


alex-l-kong:
Not done yet, just want a first pass at this. I think we should focus on just the current optimizations I've made in this code review, we can discuss multithreading/multiprocessing separately.

ngreenwald (changes requested):
I'm going to let @vacuousplanet take a look at this first. I have a feeling there might be some ways to simplify the matrix operations. If not, then I have ideas for how to optimize, but I'll hold off on that to give Adam a first pass.

ark/phenotyping/cluster.py
```diff
@@ -46,6 +46,58 @@ def winner(sample, weights):
...
+            A weight matrix of dimensions [num_x, num_y, num_chans]
+
+    Returns:
+        list:
+            The list with indices corresponding to the prediction of the respective sample
+    """
+
+    # only 1 sample provided: expand to 2-D
+    if len(samples.shape) == 1:
+        samples = np.expand_dims(samples, axis=0)
```
ackagel, line 65: when would we just have one sample?

ark/phenotyping/cluster.py
```diff
@@ -46,6 +46,58 @@ def winner(sample, weights):
...
+    # c: number of channels
+    weights_collapse = np.reshape(weights, (weights.shape[0] * weights.shape[1], weights.shape[2]))
+
+    # subtract weights from each sample individually, need to reshape back to original
+    # for proper Euclidean distance calculation
+    samples_subtract = samples - weights_collapse[:, None]
+    samples_subtract = np.reshape(samples_subtract, (weights.shape[0], weights.shape[1],
+                                                     samples_subtract.shape[1],
+                                                     samples_subtract.shape[2]))
+
```
ackagel, line 79: could be confused about shape, but assuming `samples` has shape `r x c` and weights has shape `m x n x c` I think we can do:
```
samples_subtract = samples.T[np.newaxis, np.newaxis, :, :] - weights[:, :, :, np.newaxis]
```

This should give the same result as above (and if samples is `c x r`, just don't transpose it)

ark/phenotyping/cluster.py
```diff
@@ -46,6 +46,58 @@ def winner(sample, weights):
...
+    # take the Euclidean distance along the last axis, then reshuffle
+    # to a final matrix of r x (m x n)
+    # r: number of samples
+    # (m x n): number of m and n SOM nodes
+    activation_map = np.linalg.norm(samples_subtract, axis=-1)
```
ackagel, line 84: looked under the hood of `np.linalg.norm`.  for our purposes, we can bypass the 'sqrt' at the end here using:
```
activation_map = np.add.reduce(samples_subtract * samples_subtract, axis=-1)
```
since this won't change the winning coords.

ark/phenotyping/cluster.py
```diff
@@ -46,6 +46,58 @@ def winner(sample, weights):
...
+    activation_map = np.swapaxes(np.swapaxes(activation_map, 1, 2), 0, 1)
+
+    # collapse the last two dimensions of activation map so we can unravel the indices properly
+    activation_map_collapse = np.reshape(activation_map,
+                                         (samples.shape[0], weights.shape[0] * weights.shape[1]))
+
```
ackagel, line 90: Seems like some of this rearranging could be done at the top, before subtraction.  For example:
```
samples_subtract = samples.T[:, :, np.newaxis, np.newaxis] - weights.T[:, np.newaxis, :, :]
activation_map = ...axis=1)
```

ark/phenotyping/cluster.py
```diff
@@ -24,7 +24,7 @@ def decay_function(param, t, num_iters):
 
 
 def winner(sample, weights):
-    """Find the coordinates to the winning neuron in the SOM
+    """Find the coordinates to the winning neuron in the SOM for one sample
```
ackagel, line 27: suggested linalg optimization applies here as well

ackagel (changes requested):
Looking good so far. Couple suggestions for less reshaping/swapping and a potential speedup in computing activation

ark/phenotyping/cluster.py
alex-l-kong: This would happen in `cluster_som` if the batches divided up unequally with 1 row remaining.

ark/phenotyping/cluster.py
alex-l-kong: Good point

ark/phenotyping/cluster.py
alex-l-kong: The dimensions are slightly off with this method, but I think that's fine. Let me see if I can change the `np.linalg.norm` `axis` parameter, otherwise a call to `swapaxes` shouldn't hurt.

ark/phenotyping/cluster.py
ackagel: gotcha.  Either way, couldn't we just call/return the regular `winner` function in this case?

alex-l-kong: Yeah, good point. I did it this way because I wasn't sure if we'd keep `winner` at all. Now that we're potentially moving forward with a different method, could be a moot point.

ark/phenotyping/cluster.py
ackagel: shedding some light on the main slow down here.  The main slow down is actually not the norm/adding.  The main slowdown is actually the element wise product.  On a (10, 10, 100000, 30), the total time for norm (and add.reduce) is 780ms.  Almost **450ms** of that time is spent on the element-wise product!

Luckily, we can get ~5x speedup (170ms), using:
```
activation_map = np.einsum('ijkl,ijkl->ijk', samples_subtract, samples_subtract)
```

Furthermore, if we use np.float32's instead of 64's, we basically have a full 10x speedup (110ms).  Extrapolating to Candace's dataset, this would cluster in ~5mins as opposed to 12mins (but that should be tested).  If/when we add multithreading/processing, 4 threads tackle this on the order of 1 min.

ackagel: btw, resizing is free, time wise (like 1us levels of free)

alex-l-kong: @ackagel yeah this should work, seems like `einsum` generally runs much faster than regular `numpy` multiplication due to better cache management and memory access in the underlying C implementation. I had originally complained about Sophie's C implementation of FlowSOM being very unintuitive, but now that I think about it, I think she did it that way to make the SOM training several orders more cache/memory efficient, which is in large part why Candace only needed 40 minutes to run it.

However, I'm not sure if we can extract this into our current MiniSOM-based training step. Fundamentally, the issue here is that the weights are updated per row, meaning we cannot calculate the winning coordinates for 100000 rows at a time without changing the way the algorithm trains. Hypothetically, we should get a similar speedup by structuring the array the way Sophie did it in Cython, which should also give us the C-optimization benefits as well. But at that point, it might be better to simply run Sophie's SOM directly.

@ngreenwald @ackagel what do you think?

ngreenwald: Yes, we can't do a weights update with 1 million rows at a time. What was the result of the R benchmarking? Is it fast on your laptop? If so, I think we should start with that, since it will likely require the least amount of work to just call the R function directly. Once we've got a working version, then I think it makes sense to think about whether we need to go back and optimize it ourself. 

My suspicion is that it would be one day's work at most to make the specific R functions we need callable from python. If that's not the case, then the above tradeoff might not be relevant.

alex-l-kong: @ngreenwald hopefully I'll have the results by today, otherwise over the weekend. Been a pain setting up all the dependencies and configuring my R environment to run it.

ngreenwald: sounds good! no worries

alex-l-kong:
This is just for a first pass. Please ignore any changes made in `cluster.py` since we'll no longer be using that code. I've configured the Docker and the notebook to run the clustering step on a subset of FOVs using R SOM. Note that due to the high amount of data being processed, it is not possible to cluster the entire dataset on a Docker instance. Subsetting will have to be finished before we can fully test on Docker.

ark/phenotyping/som_runner.R
```diff
@@ -0,0 +1,36 @@
...
+
+print("Reading command args")
+args <- commandArgs(trailingOnly=TRUE)
+
+# get path to pixel mat data
+print("Getting pixel mat data path")
+pixelMatPath <- args[1]
+
+# create a vector out of the list of channels provided
+print("Creating marker list")
```
ngreenwald, line 12: To me it makes more sense to do this step before saving the CSV file. 

But, if people want to try out different markers, it might be better to save everything in the CSV and then just subset to this function. Not sure.

ark/phenotyping/som_runner.R
```diff
@@ -0,0 +1,36 @@
...
+
+# read the pixel mat data
+print("Reading the pixel matrix data")
+pixelMatData <- fread(pixelMatPath, select=markers)
+
+# subset the data for SOM clustering only on the channels we're interested in
+
+# run the SOM clustering step
+print("Run the SOM clustering")
+som_results <- SOM(data=as.matrix(pixelMatData))
```
ngreenwald, line 28: This looks like the top-level SOM function. This trains a network and evaluates it on the training data. 

However, we want to separate this into a training the network step and an evaluating the network on data step. Otherwise, we won't ever be able to subset the data, train on the subset and evaluate on the whole dataset.

ngreenwald (changes requested):
Great start. The main thing is to dig into the SOM library a bit more to find out where the training and evaluation happens. Right now they're linked together which will prevent subsetting, or running a previously trained SOM on new data.

templates/example_flowsom_clustering.ipynb
```diff
@@ -15,6 +15,7 @@
    "source": [
```
ackagel, line 15: <p>Is this the best way to call R?  If it's possible, it might be better to hide away our calls to R within python functions (keep notebooks all python).  Disregard this comment if you're just using this for debugging/dev.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

ackagel:
not much to add to Noah's review, just a note on R presentation/interaction within the notebooks

templates/example_flowsom_clustering.ipynb
alex-l-kong: @ackagel for now it's just for debugging (you'll see that I also have a ton of `sys.path.append`'s as well, those gotta go too). The `!` makes it super easy for now to call `Rscript` inside the Docker, but we can absolutely move that to a `subprocess` call inside a function in `phenotyping`.

ark/phenotyping/som_runner.R
alex-l-kong: I'm specifying the marker list now to make it easy to read the CSV data in with only the markers we need before training the SOM. Would it be better to save to a CSV before training the SOM? I'm not entirely sure what the aim is here.

alex-l-kong:
This PR now contains the following:

- A fully-integrated R SOM implementation with a separate R script running the process in the background. It reads in the preprocessed pixel matrix and can use the `MapDataToCodes` function to map clusters to the entire dataset based on weights trained on just a small subset.
- Complete integration with Docker.
- Complete integration with Travis, which runs the entire `example_flowsom_clustering` notebook and does unit-based testing with some of the parameters.

This PR does not address:

- Subsetting, which needs to be handled as a separate command-line argument in `som_runner.R`. Another PR will be handling that one. Currently, we simply train and run `MapDataToCodes` on the same dataset passed in.
- Preprocessing memory optimization. Another PR will be handling that one.

Concerns:

- Due to the intense R setup that needs to be done in Travis, the tests take a very long time to run. In particular, `FlowSOM` is a massive installation that requires a ton of dependencies to be compiled. I have not yet found a way to cache the required R packages needed on Travis, and this is in large part because Travis has extremely poor multi-language support (for example, cannot specify separate caches for Python and R packages). if anyone has any experience in this regard would appreciate it.

ngreenwald:
Looks like you picked a name for a branch that existed previously, which is leading to a super messy history with 300+ commits. Can you copy your branch with a new rename and reopen the PR?

That's pretty annoying re: travis installation. Worst case scenario, we'll have to mock FlowSOM, and just test the preprocessing/postprocessing part of the codebase.

alex-l-kong:
@ngreenwald so the reason the commit history is really long is because I ran into an issue of committing file larger than 1GB to the repo and I messed up in trying to remove the file from the commit history. The same problem of large commit history exists when I tried to create a separate branch (r-som). I'll have to figure out how to remove all the unnecessary commits without causing a ton of merge conflicts.

ngreenwald:
Got it. You should be able to rebase all of the commits in your current which are before the 1gb file onto a new branch, then rebase all of the commits after the 1 gb commit onto that same new branch

alex-l-kong:
With a better PR without a ton of older commit history stored, this one's officially nuked.

PR #369, opened 2020-12-12
Auto upload patch

### LINKED ISSUE #302: Create multiple, smaller zip files to avoid timeout errors
As described in vanvalenlab/kiosk-client#59, there is currently an issue with zip uploads which are too large (somewhere around 150 Mb). It's a bit of mystery why this is happening.

 It would be good for us to figure out about how many 1024x1024 images we can safely upload without running into this error. We'll first need to address #301. Once we know what that number is, we can decide what our next steps are. We can either wait until the bug gets fixed in the kiosk, or add functionality to create multiple, smaller zip files, each of which gets uploaded separately, to work around the issue.

ackagel:
## **Purpose**

Replaces kiosk-client with direct calls to deepcell API.  This bypasses the kiosk client upload bug described in issue #302 so smaller zip files don't need to be created; we can just upload one.  Consequently closes #302 

## **Implementation**

Replaces kiosk-client usage in `deepcell_service_utils.py` with direct calls to deepcell's api.  This more closely mimics direct user interaction with deepcell.org.

## **Remaining issues**

Might want to check with Van Valen lab on how stable their deepcell API will be.

### DIFF
--- a/ark/utils/deepcell_service_utils.py
+++ b/ark/utils/deepcell_service_utils.py
@@ -175,6 +175,7 @@ def run_deepcell_direct(input_dir, output_dir, host='https://deepcell.org',
 
     upload_responce = requests.post(
         upload_url,
+        timeout=timeout,
         files=upload_fields
     ).json()
 
--- a/ark/utils/deepcell_service_utils.py
+++ b/ark/utils/deepcell_service_utils.py
@@ -14,7 +14,7 @@ from ark.utils import misc_utils
 
 def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
                            suffix='_feature_0', host='https://deepcell.org', job_type='multiplex',
-                           scale=1.0, timeout=3600):
+                           scale=1.0, timeout=3600, zip_size=100):
     """ Handles all of the necessary data manipulation for running deepcell tasks.
 
         Creates .zip files (to be used as input for DeepCell),
@@ -45,6 +45,9 @@ def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
             timeout (int):
                 Approximate seconds until timeout.
                 Default: 1 hour (3600)
+            zip_size (int):
+                Maximum number of files to include in zip.
+                Default: 100
         Raises:
             ValueError:
                 Raised if there is some fov X (from fovs list) s.t.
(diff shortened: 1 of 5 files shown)


ackagel:
tested .zip upload of ~183MB.  Looks like the uploading is fine, but it seems like deepcell can have an issue where it times itself out with too big of an upload (happens on website too).  That looks something like:

```
Traceback (most recent call last):
File "/usr/src/app/redis_consumer/consumers/base_consumer.py", line 202, in consume
status = self._consume(redis_hash)
File "/usr/src/app/redis_consumer/consumers/multiplex_consumer.py", line 124, in _consume
image = self.predict(image, model_name, model_version)
File "/usr/src/app/redis_consumer/consumers/base_consumer.py", line 614, in predict
model_dtype, untile=untile)
File "/usr/src/app/redis_consumer/consumers/base_consumer.py", line 464, in _predict_big_image
in_tensor_dtype=model_dtype)
File "/usr/src/app/redis_consumer/consumers/base_consumer.py", line 311, in grpc_image
prediction = client.predict(req_data, settings.GRPC_TIMEOUT)
File "/usr/src/app/redis_consumer/grpc_clients.py", line 185, in predict
response = self._retry_grpc(request, request_timeout)
File "/usr/src/app/redis_consumer/grpc_clients.py", line 134, in _retry_grpc
raise err
File "/usr/src/app/redis_consumer/grpc_clients.py", line 123, in _retry_grpc
response = api_call(request, timeout=request_timeout)
File "/usr/local/lib/python3.6/site-packages/grpc/_channel.py", line 826, in __call__
return _end_unary_response_blocking(state, call, False, None)
File "/usr/local/lib/python3.6/site-packages/grpc/_channel.py", line 729, in _end_unary_response_blocking
raise _InactiveRpcError(state)
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
status = StatusCode.DEADLINE_EXCEEDED
details = "Deadline Exceeded"
debug_error_string = "{"created":"@1609789741.491197858","description":"Error received from peer ipv4:10.43.253.0:8500","file":"src/core/lib/surface/call.cc","file_line":1056,"grpc_message":"Deadline Exceeded","grpc_status":4}"
```

When this type of error happens, some of the files are processed, and others are not, so one possible fix for this type of error would be to recursively upload remaining files until all the expected files exist, or to just batch from the start with smaller zip files.

The only thing with the batching, is that this error doesn't seem to be *directly* 'size' related, so much as it's 'difficulty-of-segmentation' related, which means there's no obvious batching size/number.

That being said, if deepcell fixes or extends the redis timeouts, then this isn't really an issue, since the uploading itself works fine

ngreenwald:
What do you mean it's difficulty of segmentation related?

ackagel:
I suppose that's more of a guess, but basically, on Tyler's DCIS data (51MB) less than half of the images were processed, while on the duplicated 'fov8.tif' template (183MB), more than 3/4ths of the images were processed.  Could be something else, but it's more of a problem of how long it takes the servers to process each image.  We could ask Will if there's any reason not to bump up the default redis timeout (`settings.GRPC_TIMEOUT`), since that would fix the problem.  Looks like it's currently set to 30s

ackagel:
https://github.com/vanvalenlab/kiosk-redis-consumer/blob/cefc4736d1658505966bd8bbe3638a67b83290f2/redis_consumer/settings.py#L67

ngreenwald:
Yes, will you copy the error message and open an issue in the kiosk-redis-consumer repo

alex-l-kong (commented):
Not much to add on top of Noah. Looks good so far!

ackagel:
@ngreenwald same pytest/testbook bug on this PR.  Other than that, it's good for review.

ark/utils/deepcell_service_utils.py
```diff
@@ -86,7 +94,8 @@ def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
 
     # pass the zip file to deepcell.org
     print('Uploading files to DeepCell server.')
-    run_deepcell_task(zip_path, deepcell_output_dir, host, job_type, scale)
+    # run_deepcell_task(zip_path, deepcell_output_dir, host, job_type, scale)
```
ngreenwald, line 97: Can remove the commented out line

ngreenwald (changes requested):
Just one commented line left. Did the timeout issue get resolved with Redis?

ackagel:
not sure if the timeout has been increased yet, but I'll test it today

ark/utils/deepcell_service_utils_test.py
```diff
@@ -7,7 +7,7 @@
 from ark.utils.deepcell_service_utils import create_deepcell_output
 
 
-def mocked_run_deepcell(input_dir, output_dir, host, job_type, scale):
+def mocked_run_deepcell(input_dir, output_dir, host, job_type, scale, timeout):
```
ngreenwald, line 10: Can add some simple checks for ```_zip_run_extract```
e.g.  for 210 images, make sure first zip has 100, second zip has 100, third zip has 10. 
Make sure first zip contains first 100 images. 
Make sure changing zip_size actually changes number of images included. 

Can change the default to be like 5 or 2 or something to make previous tests faster.

ngreenwald (changes requested):
Looks good, just some simple tests for new function. 

If you could also test it yourself actually uploading to deepcell.org with real data to make sure when you supply bad data, the right error message is raised.

alex-l-kong (commented):
Nothing else to add on top of Noah's comments.

ackagel:
looks like I'm getting the 'correct' error messages from deepcell for bad data

ngreenwald (changes requested):
Everything looks good. I'm fine to merge in the PR with reduced coverage since it's mostly error messages from Deepcell.org that we're catching. The one thing that isn't tested which I think is worth testing is the parallel execution option.

In addition to adding a test, can you also confirm locally that uploading 3 zips of 50 FOVs each simultaneously doesn't lead to any issues with Deepcell.org?

ark/utils/deepcell_service_utils_test.py
```diff
@@ -58,6 +64,19 @@ def test_create_deepcell_output(mocker):
...
+        # test parallel
+        create_deepcell_output(deepcell_input_dir=input_dir, deepcell_output_dir=output_dir,
+                               fovs=['fov1', 'fov2'], zip_size=1, parallel=True)
+
+        # make sure DeepCell (.zip's) output exists
+        assert os.path.exists(os.path.join(output_dir, 'example_output.zip'))
+        assert os.path.exists(os.path.join(output_dir, 'example_output_2.zip'))
+
+        os.remove(os.path.join(output_dir, 'fov1_feature_0.tif'))
+        os.remove(os.path.join(output_dir, 'fov2_feature_0.tif'))
```
ngreenwald, line 76: This would be cleaner if we made a new temp folder, rather than deleting the files we've already created. Can either make output1, output2, output3 for each part of the test, or a context manager with a temp dir

ngreenwald (changes requested):
Nice, almost there

ark/utils/deepcell_service_utils_test.py
```diff
@@ -22,87 +28,103 @@ def mocked_run_deepcell(input_dir, output_dir, host, job_type, scale):
...
+            with pytest.warns(UserWarning):
+                create_deepcell_output(deepcell_input_dir=input_dir,
+                                       deepcell_output_dir=output_dir, suffix='_other_suffix',
+                                       fovs=['fov1'])
 
-        # Warning should be displayed if fovs.zip file exists (will be overwritten)
-        with pytest.warns(UserWarning):
-            create_deepcell_output(deepcell_input_dir=input_dir, deepcell_output_dir=output_dir,
-                                   fovs=['fov1'])
+            pathlib.Path(os.path.join(input_dir, 'fov4.tif')).touch()
```
ngreenwald, line 112: Is there a specific warning that we're making sure is raised here, or is this just so that we now have 4 TIFs for the batch size checks that come next?

ackagel: the last `pytest.warns(UserWarning)` is redundantly checking warnings for 'overwriting file'; definitely could be removed if needed.  The extra fov is created for the batch size checking later.

ngreenwald (approved):

PR #371, opened 2021-01-05
Removes mibilib dependency

ackagel:
## **What is the purpose of this PR?**

Some users have had trouble building the docker due to mibilib's strict library dependencies.  We've attempted to address this before, but problems persist.  This PR removes the mibilib dependency, while preserving basic mibitiff support.

Calls to `mibilib.tiff` are no longer used; `tiff_utils.read_mibitiff` and `tiff_utils.write_mibitiff` replace its usages.

## **Remaining issues**

Only a limited subset of mibitiff is supported, not the entire specification.  All that's been implemented is the basic functionality required to read/write channel data from/to mibitiffs.

### DIFF
--- a/ark/utils/tiff_utils.py
+++ b/ark/utils/tiff_utils.py
@@ -1,0 +1,144 @@
+from fractions import Fraction
+import numpy as np
+from skimage.external.tifffile import TiffFile, TiffWriter
+import json
+import datetime
+
+
+def read_mibitiff(file, channels=None):
+    """ Reads MIBI data from an IonpathMIBI TIFF file.
+
+    Currently, only SIMS data is supported
+
+    Args:
+        file (str): The string path or an open file object to a MIBItiff file.
+        channels (list): Targets to load. If None, all targets/channels are loaded
+
+    Returns:
+        tuple (np.ndarray, list[tuple]):
+        - image data
+        - channel data
+    """
+    return_channels = []
+    img_data = []
+    with TiffFile(file) as tif:
+
+        # make sure it's a mibitiff
+        _check_version(tif)
+
+        for page in tif.pages:
+
+            # get tags as json
+            description = json.loads(
+                page.tags['image_description'].value.decode('utf-8')
+            )
+
+            # only load supplied channels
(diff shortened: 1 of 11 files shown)


ackagel:
This passes tests locally, so it should be ready for a quick review

ackagel:
@ngreenwald @alex-l-kong should be ready for review now.

btw, there was a testing bug, that slipped past us, involving something to the effect of travis/pytest not actually collecting the functions with the `@textbook` decorators, in previous tests.  Apparently, now it does, and caught a hard-coded variable that needed to be changed within those funcitons.  

Fixed that error, so this PR's good for review now, although that bug will also be present in the other open PR's.

ark/utils/tiff_utils.py
```diff
@@ -0,0 +1,144 @@
+from fractions import Fraction
+import numpy as np
+from skimage.external.tifffile import TiffFile, TiffWriter
+import json
+import datetime
+
+
+def read_mibitiff(file, channels=None):
```
ngreenwald, line 8: We don't need to exhaustively test this, but I think it would be good to have a simple test function that creates a mibi-tiff, reads it in, makes sure the channels are in the right order, etc.

ark/utils/tiff_utils.py
```diff
@@ -0,0 +1,144 @@
...
+_PREFIXED_METADATA_ATTRIBUTES = ('run', 'coordinates', 'size', 'slide',
+                                 'fov_id', 'fov_name', 'folder', 'dwell',
+                                 'scans', 'aperture', 'instrument', 'tissue',
+                                 'panel', 'mass_offset', 'mass_gain',
+                                 'time_resolution', 'miscalibrated',
+                                 'check_reg', 'filename', 'description',
+                                 'version')
+
+
+def write_mibitiff(filepath, img_data, channel_tuples, metadata):
```
ngreenwald, line 75: Same thing here

ngreenwald (changes requested):
Looks great, gonna be much cleaner. Just a couple simple tests

ngreenwald (approved):

PR #374, opened 2021-01-18
Add SOM implementation

### LINKED ISSUE #325: Create basic SOM implementation

alex-l-kong:

Had to reopen this PR because the previous one's commit history got massively messed up.

Addresses and closes #325. Candace's pixel clustering algorithm needs to be integrated with `ark-analysis` now. This PR begins to add this functionality in, starting with a basic SOM implementation. We need to verify that this PR produces the same or at least very similar results.

We piggyback off of MiniSOM (https://github.com/JustGlowing/minisom), a Python-based SOM implementation that should work for our purposes. It shouldn't be hard to translate most of the code into our project, but optimizing the process may be difficult.

Once we've verified our outputs, we then need to implement downsampling. A SOM would take a very long time to train on hundreds of thousands of pixel data from various images. We don't need all of these pixel entries.

We also need to implement 

**Remaining issues**

Preprocessing takes a very long time. Unfortunately, it seems like `apply_along_axis` doesn't offer much of a speedup. Alternatives would be greatly welcomed!


ark/utils/notebooks_test.py
```diff
@@ -123,6 +128,42 @@ def test_segment_image_data_folder(tb):
...
+    with tdir() as base_dir:
+        # create input files
+        notebooks_test_utils.flowsom_setup(tb, flowsom_dir=base_dir)
+
+        # run the FlowSOM preprocessing and clustering
+        notebooks_test_utils.flowsom_pixel_run(
+            tb, fovs=['fov0', 'fov1'], channels=['chan0', 'chan1']
+        )
+
+        # TODO: see what Brian discovers about R testing, then add cell clustering tests
```
ngreenwald, line 164: How long does the pixel clustering tests take? Can we add a simple cell clustering one to make sure the notebook runs?

ngreenwald (changes requested):
Looking nice! There's some random files that look like they got added in here too:
`data/example_dataset/flowsom_data/deepcell_output`, and 'flowsom_data/input_data`. Are those here for a particular reason?

alex-l-kong:
The notebook tests now run the visualization code, mocking only the necessary files and directories needed.

ark/utils/notebooks_test.py
```diff
@@ -22,24 +29,21 @@ def _exec_notebook(nb_filename, base_folder):
                 "--output", fout.name, path]
         subprocess.check_call(args)
 
-# test runs with default inputs
-
 
-def test_segment_image_data():
-    _exec_notebook('Segment_Image_Data.ipynb', 'templates_ark')
+# # test runs with default inputs
+# def test_segment_image_data():
```
ngreenwald, line 34: Comments

ngreenwald (changes requested):
Looks good, but some commented code still in here

ngreenwald:
Awesome, looks good. Will you make a new release and bump to v0.3.0

PR #375, opened 2021-01-18
R SOM integration

alex-l-kong:

Reopens the PR that handled R SOM implementation because the commit histories got massively messed up.

This PR now contains the following:

- A fully-integrated R SOM implementation with a separate R script running the process in the background. It reads in the preprocessed pixel matrix and can use the `MapDataToCodes` function to map clusters to the entire dataset based on weights trained on just a small subset.
- Complete integration with Docker.
- Complete integration with Travis, which runs the entire `example_flowsom_clustering` notebook and does unit-based testing with some of the parameters.

This PR does not address:

- Subsetting, which needs to be handled as a separate command-line argument in `som_runner.R`. Another PR will be handling that one. Currently, we simply train and run `MapDataToCodes` on the same dataset passed in.
- Preprocessing memory optimization to prevent Docker crashes. Another PR will be handling that one.

Concerns:

- Due to the intense R setup that needs to be done in Travis, the tests take a very long time to run. In particular, `FlowSOM` is a massive installation that requires a ton of dependencies to be compiled. I have not yet found a way to cache the required R packages needed on Travis, and this is in large part because Travis has extremely poor multi-language support (for example, cannot specify separate caches for Python and R packages). if anyone has any experience in this regard would appreciate it.


alex-l-kong:
Because of the imperfect restart, all of the `preprocessing.py` and initial setup stuff got condensed into one commit with the R SOM stuff. When reviewing this PR, please only look at:

- example_flowsom_clustering.ipynb
- som_runner.R
- Dockerfile
- .travis.yml

ark/phenotyping/som_runner.R
```diff
@@ -0,0 +1,34 @@
...
+
+# read the pixel mat data
+print("Reading the pixel matrix data")
+pixelMatData <- fread(pixelMatPath, select=markers)
+
+# TODO: subset the data for SOM clustering only on the channels we're interested in
+
+# run the SOM clustering step
+print("Run the SOM clustering")
+som_results <- SOM(data=as.matrix(pixelMatData))
```
ngreenwald, line 28: Now that we know this sanity check works, is there a way to access the hidden function SOM uses purely for training, rather than calling the entire FlowSOM process?

ark/utils/notebooks_test.py
```diff
@@ -106,3 +115,36 @@ def test_segment_image_data_folder(tb):
 
         # create the expression matrix
         notebooks_test_utils.create_exp_mat(tb)
+
+
+# test mibitiff clustering
+@testbook(FLOWSOM_CLUSTER_PATH, timeout=6000)
+def test_flowsom_cluster_mibitiff(tb):
```
ngreenwald, line 122: Each of these tests takes about 30 seconds. Once we figure out the Travis R installation issue, this will become the next biggest time sync. We can probably use images that are like 50x50 pixels, instead of 1024x1024, to dramatically speed this up.

ngreenwald (changes requested):
Nice start! Small comments below. Main thing is that we don't need to add all of these random TIFs in flowsomdata/Point14/TIFsnoagg to the repo. We should keep like 4 or 5 of them and use those as examples, not the entire mibi run. Talk to Candace and see which she thinks would work best for a simplified pixel-clustering example.

ark/phenotyping/som_runner.R
alex-l-kong: So `SOM` only does the training. The other stuff (ex. meta-clustering) are handled by separate calls.

ngreenwald: Okay got it. I assumed based on your comment below ("add original cluster labels to pixelMatData and save, used for verification"), that both the ```SOM``` call and the ```MapDataToCodes``` function were producing labels for for each row of the matrix

alex-l-kong: OK, I think I understand. `SOM` will automatically assign labels to the dataset passed in because it makes a call to `MapDataToCodes`. We need to make a separate call to `MapDataToCodes` because we don't want this default functionality of training and clustering on the same dataset, but rather only want to use the trained weights returned by `SOM` (which is done on a subset) and run `MapDataToCodes` separately on the entire dataset.

Ideally, we don't want this extra `MapDataToCodes` call during the call to `SOM`; unfortunately, as of the current version, this is default functionality. I can talk with Sophie and Sam at FlowSOM to see if it's possible to do this.

ngreenwald: No problem, it's not a big deal. If there's not already a way to do it we can put this on the backburner

ark/utils/notebooks_test.py
alex-l-kong: After removing all but 7 TIFs (the ones Candace said would pop up most frequently), I think mocking up the call to `cluster_pixels` (which implicitly calls `som_runner.R`) will provide the next biggest speed boost. Let's see how much of a speedup that provides, if it's still too much then I can reduce the size of the images.

alex-l-kong:
So here's what's changed:

- R installations are no longer necessary in `.travis.yml`: we've foregone testing the process in `som_runner.R` (which is called by `cluster_pixels`) and instead have mocked up `cluster_pixels` in the testing
- We no longer do a full notebook test of `example_flowsom_clustering.ipynb` with the above point in mind, defaulting this to `testbook` which also allows us to mock the call to `cluster_pixels`
- `testbook` tests for `example_flowsom_clustering.ipynb` are much more lightweight: we leave only a few significant markers in `example_dataset` and have reduced the test image sizes to 50x50.

.travis.yml
```diff
@@ -8,13 +8,11 @@ language: python
 python:
   - 3.6
 
-# We add python path to enable testing jupyter notebooks
+# We add python path and all R dependencies to enable testing jupyter notebooks
```
ngreenwald, line 11: No R anymore, right?

ark/phenotyping/som_utils.py
```diff
@@ -0,0 +1,115 @@
...
+        chan_list (list):
+            The list of markers to subset on
+    """
+
+    # get the path to som_runner.R
+    som_runner_path = os.path.dirname(os.path.realpath(__file__))
+
+    # use Rscript to run som_runner.R with the correct command line args
+    subprocess.call(['Rscript', os.path.join(som_runner_path, 'som_runner.R'),
+                     os.path.join(base_dir, 'example_pixel_matrix.csv'),
```
ngreenwald, line 114: We should change name from example_pixel_matrix

ark/phenotyping/som_utils_test.py
```diff
@@ -0,0 +1,101 @@
...
+import numpy as np
+import pandas as pd
+import xarray as xr
+
+import ark.phenotyping.som_utils as som_utils
+import ark.utils.misc_utils as misc_utils
+import ark.utils.test_utils as test_utils
+
+
+def mocked_cluster_pixels(base_dir, chan_list):
```
ngreenwald, line 14: The way you have this set up, the mocked function won't report errors if we created a totally garbage csv with none of the right columns and no values. 

The mocked function should as closely as possible replicate the original function, so that we catch as many errors as possible. 

We should instead actually read in the csv file, subset based on the provided column names, do a simple manipulation with all of the values in the remaining columns, use that manipulation to create cluster labels, and then save the result

ngreenwald (changes requested):
Looks good, nice progress. 

There are a couple more random files (cohortData_info, Granuluma_panel_final_comments) which can be removed. Also, don't need the nested subfolder structure for the TIFs.

ark/phenotyping/som_utils_test.py
alex-l-kong: What kind of manipulation are we talking about? The SOM process doesn't modify any of the channel values. I'm currently using the average of each row (because the test matrix values are uniform over [0, 1)) to generate dummy cluster IDs.

ngreenwald: Yeah, I think average makes sense.

ark/phenotyping/som_utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Args:
+        base_dir (str):
+            The path to the base directory
+        chan_list (list):
+            The list of markers to subset on
+    """
+
+    # get the path to som_runner.R
+    som_runner_path = os.path.dirname(os.path.realpath(__file__))
```
ngreenwald, line 110: What's the reasoning behind using os.path.realpath() here?

ngreenwald (changes requested):
Looks great. Just need to clean up the multiple nested folders currently holding the example data, and remove the .DS_Store that got added in this recent commit.

ark/phenotyping/som_utils.py
alex-l-kong: 1. This is because `som_runner.R` originally had to be run both by Travis and in Docker, and the directories are different. Now that we're not actually running `som_runner.R` in Travis, I think we can remove this and replace it with a hard coded path.
2. Can we can safely assume that the segmentation label tif file names will include the same prefixes before the underscore (`segmentationmask`) every time? It would make the `load_utils` calls a lot simpler in that case, and we can put both the channel and the segmentation TIFFs in the root of the Point directory.

ngreenwald: We can assume that the segmentation labels will be coming from our deepcell.org api call, and will have whatever format we want them to. We should mirror that setup for the example point here, or make any modifications to the directory structure we currently use for calling the deepcell server to make our life easier.

ark/phenotyping/som_utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Args:
+        img_xr (xarray.DataArray):
+            Array representing image data for each fov
+        seg_labels (xarray.DataArray):
+            Array representing segmentation labels for each image
+        fovs (list):
+            List of fovs to subset over, if None selects all
+        channels (list):
+            List of channels to subset over, if None selects all
```
ackagel, line 27: don't think this needs to be addressed now, but we should probably settle on a standard way of dealing with channels ('include_channels' vs 'exclude_channels' vs 'channels (None => all)' etc.).

ackagel (commented):
Looks good.  Not much to add.

ark/phenotyping/som_utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+def cluster_pixels(base_dir, chan_list):
+    """Run the FlowSOM training on the pixel data.
+
+    Saves results to pixel_mat_clustered.csv in base_dir.
+    Usage: Rscript som_runner.R {path_to_pixel_matrix} {chan_list_comma_separated} {save_path}
+
+    Args:
+        base_dir (str):
+            The path to the base directory
```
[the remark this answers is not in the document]
alex-l-kong: I like having `base_dir` to allow for flexibility in the save path. Any thoughts @ngreenwald?

ark/phenotyping/som_utils.py
alex-l-kong: Definitely, we can handle that in a later PR.

ark/phenotyping/som_utils.py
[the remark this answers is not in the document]
ackagel: oh I see.  I'd say just do `pixel_mat_path` as a new argument, and change `base_dir` to `out_dir` or even a full `out_path`.  Could do default path like:

```
def cluster_pixels(chan_list, out_dir, pixel_mat_path = None):
    ...
    if pixel_mat_path is None:
        # assume pixel matrix csv is in out_dir
        pixel_mat_path = os.path.join(out_dir, 'pixel_mat_preprocessed.csv')
    ...
```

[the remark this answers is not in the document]
ngreenwald: I think out_path is a good idea, otherwise the only way to change the filename is in the R code.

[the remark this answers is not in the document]
alex-l-kong: Ok sounds good, I'll put this in.

alex-l-kong:
OK, the PR is ready to be reviewed!

ark/phenotyping/som_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        base_dir (str):
+            The path to the directory to save the clustered pixel matrix in
+        pixel_pre_name (str):
+            The name of the preprocessed file name
+        pixel_cluster_name (str):
+            The name of the file to write the clustered csv to
+    """
+
+    # set the path to the preprocessed matrix, default to pixel_mat_preprocessed.csv
+    if pixel_pre_name is None:
```
ngreenwald, line 114: Rather than setting the default to None and replacing if it's not None, we should just set the default to the value that you assign here

ngreenwald (changes requested):

ngreenwald (approved):

PR #378, opened 2021-01-27
Add framework for customizable morphology metrics

### LINKED ISSUE #372: Add improved morphology metrics to default output
The current cell_data file contains only a subset of the features we've found to be useful. 

There are two things that need to be accomplished. 
1. The first, relatively easy one, is to change the default properties passed to regionprops so that all of them are included in the final csv. These should include the following:

```
props = ['label', 'area',   'major_axis_length', 'minor_axis_length',
             'perimeter', 'centroid', 'convex_area',
             'equivalent_diameter']
```

In addition, there are some custom values which can be computed from the regionprops values which we also want present in the final csv. This will require a couple lines of code once the regionprops values have been computed. These are the following:

```
cell_data['major_minor_axis_ratio'] = cell_data['major_axis_length'] / cell_data['minor_axis_length']
cell_data['perim_square_over_area'] = np.square(cell_data['perimeter']) / cell_data['area']
cell_data['major_axis_equiv_diam_ratio'] = cell_data['major_axis_length'] / cluster_df_selected['equivalent_diameter']
cell_data['convex_hull_resid'] = (cell_data['convex_area'] - cell_data['area']) / cell_data['convex_area']
cell_data['nc_ratio'] = cell_data['area_nuclear'] / cell_data['area']
```

2. The second thing that needs to be accomplished is adding in additional morphology metrics which are not simple combinations of skimage functions. The way I currently have this set up is that I run a separate skimage.regionprops call, and iterate over each object returned to compute these metrics. However, this for loop isn't super fast. It looks like the newest version of skimage includes an ["extra_properties"](https://scikit-image.org/docs/dev/api/skimage.measure.html?highlight=regionprops#skimage.measure.regionprops) argument to regionprops, which allows user-defined functions to be passed in and evaluated. This doesn't speed up the actual function I wrote, which is still slow, but it does help with the for loop aspect. However, this new version isn't compatible with some of our other packages I think. 

```
concavity_count = []
centroid_dif = []
updated_id = []

for fov in np.unique(cell_data['point']):
    current_label = segmentation_labels_selected.loc[fov, :, :, 'whole_cell'].values

    current_props = regionprops(current_label)

    for prop in current_props:
        # cell centroid shift
        cell_image = prop.image
        convex_image = prop.convex_image
        cell_M = skimage.measure.moments(cell_image)
        cell_centroid = cell_M[1, 0] / cell_M[0, 0], cell_M[0, 1] / cell_M[0, 0]

        convex_M = skimage.measure.moments(convex_image)
        convex_centroid = convex_M[1, 0] / convex_M[0, 0], convex_M[0, 1] / convex_M[0, 0]

        centroid_dist = np.sqrt((cell_centroid[1] - convex_centroid[1]) ** 2 +
                                (cell_centroid[0] - convex_centroid[0]) ** 2)

        centroid_dist /= np.sqrt(prop.area)

        # number of concavities
        diff_img = convex_image ^ cell_image
        if np.sum(diff_img) > 0:
            labeled_diff_img = label(diff_img)
            hull_prop_df = pd.DataFrame(regionprops_table(labeled_diff_img, properties=['area', 'perimeter']))
            hull_prop_df['compactness'] = np.square(hull_prop_df['perimeter']) / hull_prop_df['area']
            small_idx = np.logical_and(hull_prop_df['area'] > 10,
                                       hull_prop_df['compactness'] < 60)
            large_idx = hull_prop_df['area'] > 150
            combined_idx = np.logical_or(small_idx, large_idx)

            concavities = np.sum(combined_idx)
        else:
            concavities = 0

        concavity_count.append(concavities)
        centroid_dif.append(centroid_dist)
        updated_id.append(fov + '_' + str(prop.label))

new_df = pd.DataFrame({'unique_id': updated_id,
                       'concavity_count': concavity_count,
                       'centroid_dif': centroid_dif})


cell_data = cell_data.merge(new_df)

```

alex-l-kong:

Addresses and closes #372. Lab members have requested metrics not directly computed by `skimage` in ARK. Noah has written a couple of them and this PR will focus on getting those in. Additionally, we add a framework for flexible addition of new metrics into our cell table.

We add two new functions: `centroid_dif` and `num_concavities`, as well as a general runner function: `compute_extra_prop_info`.

**Remaining issues**

Correctness is one thing, but so is the general structure of the morphology calculations. A lot of things in the existing framework can be refactored.

### DIFF
--- a/ark/segmentation/marker_quantification.py
+++ b/ark/segmentation/marker_quantification.py
@@ -14,42 +14,6 @@ from ark.segmentation.regionprops_extraction import REGIONPROPS_FUNCTION
 import ark.settings as settings
 
 
-def compute_extra_props(props, regionprops_extras, **kwargs):
-    """Derives new features specified by regionprops_extras from a regionprops features
-
-    Args:
-        props (skimage.measure.regionprops):
-            A list of property information returned by regionprops
-        regionprops_extras (list):
-            A list of regionprops features to compute, each value should correspond to
-            a function in the global regionprops_functions variable
-        **kwargs:
-            Arbitrary keyword arguments
-
-    Returns:
-        pandas.DataFrame:
-            A dataframe with columns corresponding to each regionprops_extras feature
-    """
-
-    misc_utils.verify_in_list(
-        extras_props=regionprops_extras,
-        props_options=list(REGIONPROPS_FUNCTION.keys())
-    )
-
-    # define an empty list for each regionprop feature
-    prop_extra_data = {re: [] for re in regionprops_extras}
-
-    # generate the required data for each cell
-    for prop in props:
-        for re in regionprops_extras:
-            prop_extra_data[re].append(REGIONPROPS_FUNCTION[re](prop, **kwargs))
-
-    # convert the dictionary to a DataFrame
-    prop_extra_df = pd.DataFrame.from_dict(prop_extra_data)
-
-    return prop_extra_df
-
-
 def get_single_compartment_props(segmentation_labels, regionprops_base,
                                  regionprops_single_comp, **kwargs):
     """Gets regionprops features from the provided segmentation labels for a fov
--- a/ark/segmentation/marker_quantification.py
+++ b/ark/segmentation/marker_quantification.py
@@ -31,6 +31,11 @@ def compute_extra_prop_info(prop_info, regionprops_extras, **kwargs):
             A dataframe with columns corresponding to each regionprops_extras feature
     """
 
+    misc_utils.verify_in_list(
+        regionprops_extras=regionprops_extras,
+        regionprops_extras_options=list(REGIONPROPS_FUNCTION.keys())
+    )
+
     # define an empty list for each regionprop feature
     prop_extra_data = {re: [] for re in regionprops_extras}
 
@@ -162,6 +167,9 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
         regionprops_features = ['label', 'area', 'eccentricity', 'major_axis_length',
                                 'minor_axis_length', 'perimeter', 'centroid']
 
+    if regionprops_extras is None:
+        regionprops_extras = ['centroid_dif', 'num_concavities']
+
     if 'coords' not in regionprops_features:
         regionprops_features.append('coords')
 
@@ -188,6 +196,10 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
         regionprops_names.remove('centroid')
         regionprops_names += ['centroid-0', 'centroid-1']
 
+    # add the extras features to regionprops_names
+    regionprops_names.extend(regionprops_extras)
+
+    # get all the cell ids
     unique_cell_ids = np.unique(segmentation_labels[..., 0].values)
     unique_cell_ids = unique_cell_ids[np.nonzero(unique_cell_ids)]
 
@@ -221,10 +233,6 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
 
         nuc_props = get_cell_props(segmentation_labels.loc[:, :, 'nuclear'].values,
                                    regionprops_features, regionprops_extras, **kwargs)
-
-    # extend regionprops_names to include regionprops_extras for assigning cell features
-    if regionprops_extras is not None:
-        regionprops_names.extend(regionprops_extras)
 
     # loop through each cell in mask
     for cell_id in cell_props['label']:
--- a/ark/segmentation/marker_quantification.py
+++ b/ark/segmentation/marker_quantification.py
@@ -50,16 +50,19 @@ def compute_extra_props(props, regionprops_extras, **kwargs):
     return prop_extra_df
 
 
-def get_cell_props(segmentation_labels, regionprops_features, regionprops_extras, **kwargs):
+def get_single_compartment_props(segmentation_labels, regionprops_base,
+                                 regionprops_single_comp, **kwargs):
     """Gets regionprops features from the provided segmentation labels for a fov
+
+    Based on segmentation labels from a single compartment
 
     Args:
         segmentation_labels (numpy.ndarray):
             rows x columns matrix of masks
-        regionprops_features (list):
-            morphology features for regionprops to extract for each cell
-        regionprops_extras (list):
-            list of extra properties derived from regionprops to compute
+        regionprops_base (list):
+            base morphology features directly computed by regionprops to extract for each cell
+        regionprops_single_comp (list):
+            list of single compartment extra properties derived from regionprops to compute
         **kwargs:
             Arbitrary keyword arguments for compute_extra_props
 
@@ -68,20 +71,40 @@ def get_cell_props(segmentation_labels, regionprops_features, regionprops_extras
             Contains the regionprops info (base and derived) for each labeled cell
     """
 
-    cell_props = pd.DataFrame(regionprops_table(segmentation_labels,
-                                                properties=regionprops_features))
+    # verify that all the regionprops single compartment featues provided actually exist
+    misc_utils.verify_in_list(
+        extras_props=regionprops_single_comp,
+        props_options=list(REGIONPROPS_FUNCTION.keys())
+    )
 
-    # compute the extras properties for each cell, and append to cell_props
+    # get the base features
+    cell_props = pd.DataFrame(regionprops_table(segmentation_labels,
+                                                properties=regionprops_base))
+
+    # define an empty list for each regionprop feature
+    cell_props_single_comp = {re: [] for re in regionprops_single_comp}
+
+    # get regionprop info needed for single compartment computations
     props = regionprops(segmentation_labels)
-    extra_prop_data = compute_extra_props(props, regionprops_extras, **kwargs)
-    cell_props = pd.concat([cell_props, extra_prop_data], axis=1)
+
+    # generate the required data for each cell
+    for prop in props:
+        for re in regionprops_single_comp:
+            cell_props_single_comp[re].append(REGIONPROPS_FUNCTION[re](prop, **kwargs))
+
+    # convert the dictionary to a DataFrame
+    cell_props_single_comp = pd.DataFrame.from_dict(cell_props_single_comp)
+
+    # append the single compartment derived props info to the cell_props DataFrame
+    cell_props = pd.concat([cell_props, cell_props_single_comp], axis=1)
 
     return cell_props
 
 
-def assign_cell_features(marker_counts, compartment, cell_props, cell_coords, cell_id,
-                         label_id, input_images, regionprops_names, extraction, **kwargs):
-    """Assign the regionprops features and signal intensity to cell_id in marker_counts
+def assign_single_compartment_features(marker_counts, compartment, cell_props, cell_coords,
+                                       cell_id, label_id, input_images, regionprops_names,
+                                       extraction, **kwargs):
+    """Assign computed regionprops features and signal intensity to cell_id in marker_counts
 
     Args:
         marker_counts (xarray.DataArray):
@@ -104,6 +127,10 @@ def assign_cell_features(marker_counts, compartment, cell_props, cell_coords, ce
             the extraction method to use for signal intensity calculation
         **kwargs:
             arbitrary keyword arguments
+
+    Returns:
+        xarray.DataArray:
+            the updated marker_counts matrix with data for the specified cell_id and compartment
     """
 
     # get centroid corresponding to current cell
@@ -127,33 +154,41 @@ def assign_cell_features(marker_counts, compartment, cell_props, cell_coords, ce
     # add cell size to first column
     marker_counts.loc[compartment, cell_id, marker_counts.features[0]] = cell_coords.shape[0]
 
+    return marker_counts
 
-def assign_nuclear_features(marker_counts, regionprops_nuclear, **kwargs):
-    """Assigns the nuclear-specific properties for marker_counts
+
+def assign_multi_compartment_features(marker_counts, regionprops_multi_comp, **kwargs):
+    """Assigns features to marker_counts that depend on multiple compartments
 
     Args:
         marker_counts (xarray.DataArray):
             xarray containing segmentaed data of cells x markers
-        regionprops_nuclear (list):
-            list of nuclear properties derived from regionprops to compute, each value
-            should correspond to a value in REGIONPROPS_FUNCTION
+        regionprops_multi_comp (list):
+            list of multi-compartment properties derived from regionprops to compute,
+            each value should correspond to a value in REGIONPROPS_FUNCTION
         **kwargs:
             arbitrary keyword arguments
+
+    Returns:
+        xarray.DataArray:
+            the updated marker_counts matrix with data for the specified cell_id and compartment
     """
 
     misc_utils.verify_in_list(
-        nuclear_props=regionprops_nuclear,
+        nuclear_props=regionprops_multi_comp,
         props_options=list(REGIONPROPS_FUNCTION.keys())
     )
 
-    for rn in regionprops_nuclear:
-        REGIONPROPS_FUNCTION[rn](marker_counts, **kwargs)
+    for rn in regionprops_multi_comp:
+        marker_counts = REGIONPROPS_FUNCTION[rn](marker_counts, **kwargs)
+
+    return marker_counts
 
 
 def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=False,
-                          regionprops_features=copy.deepcopy(settings.REGIONPROPS_FEATURES),
-                          regionprops_extras=copy.deepcopy(settings.REGIONPROPS_EXTRAS),
-                          regionprops_nuclear=copy.deepcopy(settings.REGIONPROPS_NUCLEAR),
+                          regionprops_base=copy.deepcopy(settings.REGIONPROPS_BASE),
+                          regionprops_single_comp=copy.deepcopy(settings.REGIONPROPS_SINGLE_COMP),
+                          regionprops_multi_comp=copy.deepcopy(settings.REGIONPROPS_MULTI_COMP),
                           split_large_nuclei=False,
                           extraction='total_intensity', **kwargs):
     """Extract single cell protein expression data from channel TIFs for a single fov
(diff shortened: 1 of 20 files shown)


alex-l-kong:
Not there yet, just looking for some initial feedback on the current structure I've put in.

ackagel (commented):
good first pass.  structure-wise, I think the extra regionprop functions should be moved into a new utils file.  Other than that, a few condensing suggestions.

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,108 @@
...
+            The property information for a cell returned by regionprops
+
+    Returns:
+        float:
+            The centroid shift for the cell
+    """
+
+    cell_image = prop.image
+    cell_M = skimage.measure.moments(cell_image)
+    cell_centroid = cell_M[1, 0] / cell_M[0, 0], cell_M[0, 1] / cell_M[0, 0]
```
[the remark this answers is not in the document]
ngreenwald: props has the global cell centroid, but the convex hull centroid is defined wrt to the bounding box around the cell. Or maybe it's the other way around? I can't remember, but that mismatch is why I recomputed it. There might be a better way to handle that.

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,108 @@
...
+                            (cell_centroid[0] - convex_centroid[0]) ** 2)
+
+    centroid_dist /= np.sqrt(prop.area)
+
+    centroid_dif.append(centroid_dist)
+
+    return centroid_dif
+
+
+def num_concavities(prop):
```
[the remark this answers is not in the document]
ngreenwald: Yeah definitely

ngreenwald (changes requested):
Left my comments on Adam's review

ark/segmentation/marker_quantification.py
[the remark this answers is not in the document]
ackagel: cool cool.  then in terms of formatting the extra regionprops functions, they should probably take arbitrary kwargs, or use an if/else-if 'tree' instead of a global string->function map.  (I know we've had that philosophical discussion before with something else; I still stand on team function map)

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,108 @@
...
+        hull_prop_df['compactness'] = np.square(hull_prop_df['perimeter']) / hull_prop_df['area']
+        small_idx = np.logical_and(hull_prop_df['area'] > 10,
+                                   hull_prop_df['compactness'] < 60)
+        large_idx = hull_prop_df['area'] > 150
+        combined_idx = np.logical_or(small_idx, large_idx)
+
+        concavities = np.sum(combined_idx)
+    else:
+        concavities = 0
+
```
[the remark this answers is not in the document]
alex-l-kong: Ideally, this PR will likely handle a full refactor. But if these utils are needed soon, I'll break refactoring into a separate PR so people can start using these ASAP while that is happening. Thoughts @ngreenwald?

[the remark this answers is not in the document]
ngreenwald: Lets do it all at once. I'm sure people can find other ways to entertain themselves while they wait

ark/segmentation/marker_quantification.py
```diff
@@ -373,3 +481,9 @@ def generate_cell_table(segmentation_labels, tiff_dir, img_sub_folder,
         )
 
     return combined_cell_table_size_normalized, combined_cell_table_arcsinh_transformed
+
+
+regionprops_functions = {
+    'centroid_dif': centroid_dif,
+    'num_concavities': num_concavities
+}
```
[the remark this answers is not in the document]
alex-l-kong: I can capitalize `regionprops_functions` but the problem is that these contain function references, so I'll have to put this just below them at best (and not at the very top).

ark/segmentation/marker_quantification.py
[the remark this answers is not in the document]
alex-l-kong: @ackagel yes, I agree that `kwargs` is the better way to go.

@ngreenwald the way `Segment_Image_Data.ipynb` is set up right now, it's not obvious to the user that they can pass in a different `extraction` parameter or a `kwargs` dictionary to tweak how `signal_extraction`/`regionprops_extraction` works. Is it time to add a sample `kwargs` dict in the notebook, say before the cell which calls `generate_cell_table`?

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,108 @@
...
+            The centroid shift for the cell
+    """
+
+    cell_image = prop.image
+    cell_M = skimage.measure.moments(cell_image)
+    cell_centroid = cell_M[1, 0] / cell_M[0, 0], cell_M[0, 1] / cell_M[0, 0]
+
+    convex_image = prop.convex_image
+    convex_M = skimage.measure.moments(convex_image)
+    convex_centroid = convex_M[1, 0] / convex_M[0, 0], convex_M[0, 1] / convex_M[0, 0]
```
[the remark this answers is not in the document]
alex-l-kong: @ackagel I like the idea of condensing `centroid_dist` into one call to `np.linalg.norm`, but I think the `image in [prop.image, prop.convex_image]` is a bit more confusing. As a compromise, we can leave the `cell_centroid`/`convex_centroid` code in but do `centroid_dist = np.linalg.norm(cell_centroid - convex_centroid) / np.sqrt(prop.area)` at the end instead.

ark/segmentation/marker_quantification.py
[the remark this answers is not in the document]
ngreenwald: Lets keep it on the secret menu for now. There's currently not a reason to use anything besides default extraction, so I'd rather keep things simple until we have a clear use-case for this functionality.

ark/segmentation/marker_quantification.py
```diff
@@ -27,26 +144,32 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
...
         extraction=extraction,
-        extraction_options=list(extraction_function.keys())
+        extraction_options=list(EXTRACTION_FUNCTION.keys())
     )
 
     if regionprops_features is None:
         regionprops_features = ['label', 'area', 'eccentricity', 'major_axis_length',
                                 'minor_axis_length', 'perimeter', 'centroid']
 
+    if regionprops_extras is None:
```
ngreenwald, line 170: I think we've been overusing this pattern. It makes are our function definitions less clear. If we're going to assign this as the default, just make this the default in the function. ```None``` makes sense as the default when we need to compute something, and hence can't set an informative default. However, if we can set the default, we should just set it.

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,133 @@
...
+    prop_extra_df = pd.DataFrame.from_dict(prop_extra_data)
+
+    return prop_extra_df
+
+
+def get_cell_props(segmentation_labels, regionprops_features, regionprops_extras=None, **kwargs):
+    """Gets regionprops features from the provided segmentation labels
+
+    Args:
+        segmentation_labels (numpy.ndarray):
```
ngreenwald, line 57: Specify that this represents segmentation from a single compartment from a single FOV

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,133 @@
...
 
 from ark.utils import io_utils, load_utils, misc_utils, segmentation_utils
-from ark.segmentation.signal_extraction import extraction_function
+from ark.segmentation.signal_extraction import EXTRACTION_FUNCTION
+from ark.segmentation.regionprops_extraction import REGIONPROPS_FUNCTION
 
 import ark.settings as settings
 
 
+def compute_extra_prop_info(prop_info, regionprops_extras, **kwargs):
```
ngreenwald, line 17: can shorten to compute_extra_props

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,133 @@
...
+    """Derives new features specified by regionprops_extras from a regionprops features
+
+    Args:
+        prop_info (skimage.measure.regionprops):
```
ngreenwald, line 21: lets use props instead of prop_info to be more concise and consistent with regionprops naming

ark/segmentation/marker_quantification.py
```diff
@@ -104,66 +231,34 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
...
-        # combine marker counts and morphology metrics together
-        cell_features = np.concatenate((cell_counts, current_cell_props), axis=None)
-
-        # add counts of each marker to appropriate column
-        marker_counts.loc['whole_cell', cell_id, marker_counts.features[1]:] = cell_features
-
-        # add cell size to first column
-        marker_counts.loc['whole_cell', cell_id, marker_counts.features[0]] = cell_coords.shape[0]
+        # assign whole_cell features
+        assign_cell_features(
```
ngreenwald, line 243: Nice, this is much cleaner

ark/segmentation/marker_quantification_test.py
```diff
@@ -1,17 +1,146 @@
...
+
+    # define the names of the extras
+    regionprops_extras = ['centroid_dif', 'num_concavities']
+
+    # define the names of everything
+    regionprops_names = ['label', 'area', 'eccentricity', 'major_axis_length',
+                         'minor_axis_length', 'perimeter', 'label',
+                         'centroid-0', 'centroid-1', 'centroid_dif', 'num_concavities']
+
+    # get all the cell ids, for testing we'll only use 1 cell id
```
ngreenwald, line 101: Can you test two cells instead of 1, and check that the output is right for each of them?

ark/segmentation/regionprops_extraction.py
```diff
@@ -0,0 +1,80 @@
+import copy
+
+import numpy as np
+import pandas as pd
+
+from skimage.measure import label, moments, regionprops_table
+
+
+def centroid_dif(prop, **kwargs):
+    """Return the cell centroid shift for a cell
```
ngreenwald, line 10:
```suggestion
    """Return the normalized euclidian distance between the centroid of the cell and the centroid of the corresponding convex hull
```

ark/segmentation/regionprops_extraction.py
```diff
@@ -0,0 +1,80 @@
...
+
+    diff_img = convex_image ^ cell_image
+
+    if np.sum(diff_img) > 0:
+        labeled_diff_img = label(diff_img)
+        hull_prop_df = pd.DataFrame(regionprops_table(labeled_diff_img,
+                                                      properties=['area', 'perimeter']))
+        hull_prop_df['compactness'] = np.square(hull_prop_df['perimeter']) / hull_prop_df['area']
+
+        small_idx_area_cutoff = kwargs.get('small_idx_area_cutoff', 10)
```
ngreenwald, line 61: Can we rename the "cutoffs" so that the directionality is clear? i.e. small_concavity_minimum, max_compactness, large_concavity_minimum

ark/segmentation/regionprops_extraction_test.py
```diff
@@ -0,0 +1,35 @@
...
+    sample_arr[ellipse_x, ellipse_y] = 1
+
+    prop_info = regionprops(sample_arr)[0]
+
+    centroid_dist = regionprops_extraction.centroid_dif(prop_info)
+
+    assert np.round(centroid_dist, 4) == 0.3562
+
+
+def test_num_concavities():
```
ngreenwald, line 25: Add a check that kwargs are passed correctly and change # of concavities

ngreenwald (changes requested):
Great start! Looking much cleaner already

ark/segmentation/marker_quantification.py
alex-l-kong: OK sounds good. This would make a perfect candidate for ark.settings.

ngreenwald: If they're going to be used in multiple functions, then I agree. If it's just going to be used for a single function, then probably overkill. I'm guessing some of these will be copied multiple places, others only once, so you'll be able to decide on a case-by-case basis

ackagel (commented):
Looks good; nothing to add but my shame

alex-l-kong:
This commit adds all the new properties to `marker_counts`.

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,156 @@
...
+    current_cell_props = cell_props.loc[cell_props['label'] == label_id, regionprops_names]
+
+    # combine marker counts and morphology metrics together
+    cell_features = np.concatenate((cell_counts, current_cell_props), axis=None)
+
+    # add counts of each marker to appropriate column
+    marker_counts.loc[compartment, cell_id, marker_counts.features[1]:] = cell_features
+
+    # add cell size to first column
+    marker_counts.loc[compartment, cell_id, marker_counts.features[0]] = cell_coords.shape[0]
```
ngreenwald, line 128: I think for clarity of what's happening, it's better to return the updated data_frame instead of just modifying in place, makes the top-level functions easier to follow.

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,156 @@
...
+
+
+def assign_nuclear_features(marker_counts, regionprops_nuclear, **kwargs):
```
ngreenwald, line 131: N/C ratio isn't nuclear-specific, it relates to both the nuclear compartment and the cell compartment. 

Some name that suggests that this function is handling the features which relate to multiple different compartments would be good.

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,156 @@
...
+
+    # compute the extras properties for each cell, and append to cell_props
+    props = regionprops(segmentation_labels)
+    extra_prop_data = compute_extra_props(props, regionprops_extras, **kwargs)
+    cell_props = pd.concat([cell_props, extra_prop_data], axis=1)
+
+    return cell_props
+
+
+def assign_cell_features(marker_counts, compartment, cell_props, cell_coords, cell_id,
```
ngreenwald, line 82: Same thing here, assign compartment features would be more clear. And potentially replacing all of the references to cell with compartment? Up to you

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,156 @@
...
+            arbitrary keyword arguments
+    """
+
+    misc_utils.verify_in_list(
+        nuclear_props=regionprops_nuclear,
+        props_options=list(REGIONPROPS_FUNCTION.keys())
+    )
+
+    for rn in regionprops_nuclear:
+        REGIONPROPS_FUNCTION[rn](marker_counts, **kwargs)
```
ngreenwald, line 150: And here, I think returning modified df is more clear

ngreenwald (changes requested):
Reorganization looks great, just some naming suggestions to make it more clear what's going on in the code

ark/segmentation/marker_quantification.py
alex-l-kong: I think it would be best to leave the cell references because those actually refer to a specific cell's id, coordinates, etc. However, the corresponding cell's data will be assigned to a specific compartment, which is specified as an argument to `assign_cell_features`.

ark/segmentation/marker_quantification.py
alex-l-kong: I'll call this `assign_multi_compartment_features` and change some of the parameters accordingly.

ark/segmentation/marker_quantification.py
```diff
@@ -5,16 +5,155 @@
...
+    current_cell_props = cell_props.loc[cell_props['label'] == label_id, regionprops_names]
+
+    # combine marker counts and morphology metrics together
+    cell_features = np.concatenate((cell_counts, current_cell_props), axis=None)
+
+    # add counts of each marker to appropriate column
+    marker_counts.loc[compartment, cell_id, marker_counts.features[1]:] = cell_features
+
+    # add cell size to first column
+    marker_counts.loc[compartment, cell_id, marker_counts.features[0]] = cell_coords.shape[0]
```
ngreenwald, line 119: I think this is a relic of before we had this nice regionprops set up. Could we include 'area' as a required regionprop? Then we wouldn't need this step

ark/segmentation/marker_quantification.py
```diff
@@ -104,66 +255,43 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
...
+                # assign properties for nuclear compartment
+                marker_counts = assign_single_compartment_features(
+                    marker_counts, 'nuclear', nuc_props, nuc_coords, cell_id, nuc_id,
+                    input_images, regionprops_names, extraction, **sig_kwargs
+                )
 
-                # add cell size to first column
-                marker_counts.loc['nuclear', cell_id, marker_counts.features[0]] = \
-                    nuc_coords.shape[0]
+                # assign nuclear-specific properties
```
ngreenwald, line 291:
```suggestion
                # generate properties which involve multiple compartments
```
```suggestion
                # assign nuclear-specific properties
```

ngreenwald (changes requested):
Looks nice

ark/segmentation/marker_quantification.py
alex-l-kong: `area` is already a required regionprop. If we don't want this `marker_counts.features[0]` assignment, what we would need to do is to remove appending `cell_size` (associated with `ark.settings.PRE_CHANNEL_COL`) as a feature. Is that what we intend?

ngreenwald: So the current setup always produces a cell table with both area and cell size as separate columns, but we're currently using cell_size as 0th column to determine where the channels start?

alex-l-kong: @ngreenwald I see the problem now. `cell_size` will always be equal to `area`. No need to leave that in then.

ngreenwald: Yeah, we can change it to label

ngreenwald: Also I don't get a notification when you edit your comment

alex-l-kong: @ngreenwald gotta keep that in mind for the future.

alex-l-kong: @ngreenwald looking over `transform_expression_matrix` more, it can be a bit of work understanding that `cell_size` and `label` are the 'bread slices' of the channel columns. Two alternatives:

1. Pass in `input_images.channels` as a parameter and use the first and last channel names in this array for indexing.
2. If we can assume that `cell_size` will always the 0th index under the current implementation, then when we remove it, the first channel will become the new 0th index. We then just need the index of `label` to identify the last channel name.

Personally, I like the first one more because it's more flexible to future changes we may make to the cell table structure. However, would definitely like your opinion on this too.

ngreenwald: You're tlaking about this line? https://github.com/angelolab/ark-analysis/blob/1776cc7647958481c6e0c82c5afeeec473a06608/ark/utils/segmentation_utils.py#L119

alex-l-kong: Yes. So instead, we would do something like:

```
channel_start = np.where(cell_table.features == channels[0])[0][0] + 1
channel_end = np.where(cell_table.features == channels[-1])[0][0] + 1
```

ngreenwald: I think adding a comment should serve the same purpose of clarification, right? Or is it something about the global variables that makes it particularly confusing?

alex-l-kong: Per today's meeting, we'll open a separate issue/PR for this.

ngreenwald (approved):

PR #380, opened 2021-02-08
adds requests dependency

ackagel:
## **Purpose**

I forgot to add the `requests` dependency to `requirements.txt` in PR #369.  Travis passed then only because `requests` is a sub-dependency of one of our `requirements-test.txt` dependencies, meaning it gets installed as if it were a regular requirement.

Adds `requests` as dependency in requirements and setup


alex-l-kong (approved):
👍

ngreenwald (approved):

PR #392, opened 2021-02-24
Throw error on nonexistent channel reads in mibitiffs

ackagel:
## **Purpose**

Adds error message to `read_mibitiff` if a non-existent channel is provided as input. This makes debugging typos/capitalizations in channel names easier to catch.


ark/utils/tiff_utils.py
```diff
@@ -46,6 +47,16 @@ def read_mibitiff(file, channels=None):
             # read image data
             img_data.append(page.asarray())
 
+    # make sure all passed channels were found
+    if channels is not None:
+        channel_names = [return_channel[1] for return_channel in return_channels]
+        found_channels = [channel not in channel_names for channel in channels]
```
ngreenwald, line 53: We can do verify_in_list for this right?

ngreenwald (changes requested):
Looks good, I think we can use verify_in_list for the actual check

ark/utils/tiff_utils.py
ackagel: totally.  I do think it should be an IndexError instead of a ValueError in this case, so I'll wrap it in a try/except block and re-raise the error as an IndexError.

ngreenwald: Sounds good. Other option would be to add option to specify error type as an argument to verify_in_list. I'm cool with either

ngreenwald (changes requested):
Does the 
```
raise IndexError('Passed unknown channels...') from exc
```
syntax also pass the message from the original exception? If so, then this is fine. If not, then it's going to hide the info telling people which values they have that are bad

ackagel:
yep, it passes the original message as well, and also notes that the error was caused by that error; basically builds a little traceback.

ngreenwald (approved):

PR #402, opened 2021-03-04
Split nuclei bugfix

ngreenwald:

The current split_nuclei logic first modifies the nuclear segmentation mask, and then uses that modified mask to extract signal. However, it doesn't provide that modified mask to the regionprops function, which results in a mismatch between the outputs from the two

I passed in the modified nuclear segmentations to the regionprops function so that the outputs match up. I also switched the min_size value to match up with the current min_size from Mesmer

**Remaining issues**
We don't pass these modified nuc masks back to the user, but given that this setting is turned off by default I think that's okay for now.

### DIFF
--- a/ark/segmentation/marker_quantification.py
+++ b/ark/segmentation/marker_quantification.py
@@ -255,7 +255,7 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
                                                       nuc_segmentation_labels=nuc_labels,
                                                       cell_ids=unique_cell_ids)
 
-        nuc_props = get_single_compartment_props(segmentation_labels.loc[:, :, 'nuclear'].values,
+        nuc_props = get_single_compartment_props(nuc_labels,
                                                  regionprops_base, regionprops_single_comp,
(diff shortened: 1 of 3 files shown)


alex-l-kong (approved):
Looks good!

ackagel (commented):
any reason we don't overwrite the unmodified nuclear mask with the modified one?

ngreenwald:
Yeah I wasn't sure what the best approach would be. We could save over the deepcell output directly?

ackagel:
is splitting the big nuclei a toggle-able option for the user?  I think we can overwrite as long as we're informing and/or putting the onus on the user

ngreenwald:
Yes, in theory it's toggle-able, but right now the kwarg isn't passed to the top-level function and it defaults to False. 

I think your suggestion makes sense once we give users the option of enabling this. For now, there hasn't been demand for it, so I'm going to leave it hidden and off to keep things simple until we start hearing issues from people re: large nuclei

ackagel (approved):

PR #403, opened 2021-03-04
Modify validate_paths to work outside docker

ngreenwald:

The current validate_paths function assumes that the user is within the docker and all paths are prefixed with /data. However, as we start to use these functions elsewhere and pip install the library, we don't want this assumption to necessarily be true. 

I've modified this function to give the option of dropping the /data prefix requirement, and updated all calls to validate_paths which are outside of a jupyter notebook to use this new logic.


ackagel (commented):
looks good.  are these all of the non-notebook calls to `validate_paths`?

ngreenwald:
All the ones that I found

alex-l-kong (approved):
Looks good.

PR #405, opened 2021-03-05
Prevent divide by zero

ngreenwald:

The custom regionprops functions don't do any validation of the inputs. In general this is fine, since they come from the base regionprops function. However, minor_axis_length seems to fairly regularly return a value of zero, which causes errors. 

Add error checking and return NaN for division by zero, rather than an error.


alex-l-kong (commented):
Looks great. There's no chance the other values will be zero, right?

ngreenwald:
Yeah, right now it seems like this is the only one

alex-l-kong (approved):
👍

ackagel (approved):

PR #407, opened 2021-03-09
Fix consensus clustering so it assigns hclust values to each pixel

alex-l-kong:

Currently, we're only saving the consensus cluster to SOM cluster mapping table. We actually want to use that mapping to assign the consensus clusters to each pixel in our dataset.

We pass in more arguments to `consensus_cluster.R` and write the results to a new directory: `pixel_mat_consensus`.

### DIFF
--- a/ark/phenotyping/som_utils.py
+++ b/ark/phenotyping/som_utils.py
@@ -16,7 +16,7 @@ from ark.utils import misc_utils
 
 def compute_cluster_avg(fovs, channels, base_dir,
                         cluster_dir='pixel_mat_clustered',
-                        cluster_avg_name='pixel_cluster_avg.feather'):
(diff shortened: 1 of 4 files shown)


ark/phenotyping/som_utils.py
```diff
@@ -16,7 +16,7 @@
 
 def compute_cluster_avg(fovs, channels, base_dir,
                         cluster_dir='pixel_mat_clustered',
-                        cluster_avg_name='pixel_cluster_avg.feather'):
+                        cluster_avg_name='pixel_cluster_avg.feather',):
```
ngreenwald, line 19: typo

ngreenwald (changes requested):
Looks good

ngreenwald (approved):

PR #409, opened 2021-03-10
Fix single compartment indexing for morphology computations

### LINKED ISSUE #401: Warning message raised when computing multiple compartments
The current implementation of the subcellular localization extraction has a warning message that gets raised for every cell in every image. 
image

The reason this occurs is because on this line
https://github.com/angelolab/ark-analysis/blob/29d2c4a2f6073567c6bfe1710d62079e3c121a36/ark/segmentation/marker_quantification.py#L110
, we're indexing into the array using only the cell-specific names. However, the cell_props contains columns for N/C ratio, which isn't included. 

It sounds like in the future, this warning may get converted into an error. Maybe we should avoid adding the multi-compartment names as columns to the cell_data until we compute them? Open to other suggestions as well

alex-l-kong:

Closes #401. Currently, we index the cell table when computing single compartment features with all the features (including multi compartment features), which throws a warning. This warning likely will become an error in the future, so best to patch this up now.

We don't append `regionprops_multi_comp` to `regionprops_names` until we actually need to compute multi compartment features. This means that we also need to explicitly append multi comp features to `marker_counts` prior to computing them.

### DIFF
--- a/ark/segmentation/marker_quantification.py
+++ b/ark/segmentation/marker_quantification.py
@@ -144,6 +144,19 @@ def assign_multi_compartment_features(marker_counts, regionprops_multi_comp, **k
     )
 
     for rn in regionprops_multi_comp:
+        # if rn is not a feature, then we add a new dimension to hold this regionprop info
+        if rn not in marker_counts.features.values:
+            rn_fill = np.zeros((marker_counts.shape[0], marker_counts.shape[1], 1))
+            rn_arr = xr.DataArray(
+                rn_fill,
+                coords=[marker_counts.compartments.values, marker_counts.cell_id.values, [rn]],
+                dims=['compartments', 'cell_id', 'features']
+            )
+
+            # append new dimension to marker_counts
+            marker_counts = xr.concat([marker_counts, rn_arr], dim='features')
+
+        # compute the multi-compatment regionprop info
         marker_counts = REGIONPROPS_FUNCTION[rn](marker_counts, **kwargs)
 
     return marker_counts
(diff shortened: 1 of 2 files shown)


ark/segmentation/marker_quantification.py
```diff
@@ -293,6 +303,10 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
                     marker_counts, regionprops_multi_comp
                 )
 
+                # if regionprops names does not contain multi_comp props then add them
+                if not set(regionprops_multi_comp).issubset(regionprops_names):
```
ngreenwald, line 307: This is so that after the first iteration through the for loop, subsequent calls to `assign_single_compartment_features` will have the right number of columns?

ark/segmentation/marker_quantification_test.py
```diff
@@ -161,6 +161,9 @@ def test_assign_multi_compartment_features():
         sample_marker_counts, regionprops_multi_comp
     )
 
+    # assert we added nc_ratio as a features key
+    assert 'nc_ratio' in sample_marker_counts.features.values
+
```
ngreenwald, line 166: Did you check that before this change you could see the warning, and after the warning no longer shows up?

ngreenwald (changes requested):
Small tweaks only

ark/segmentation/marker_quantification.py
alex-l-kong: Yes. We only want to add  `regionprops_multi_comp` to `regionprops_names` after all of them have been created in `marker_counts`, which happens in the first iteration.

ark/segmentation/marker_quantification_test.py
alex-l-kong: Yes, not seeing this warning pop up anymore 👍.

ngreenwald (approved):

PR #410, opened 2021-03-12
Add heatmap visualizations in for SOM results

alex-l-kong:

To verify results, we want to provide users a visualization of the average marker expression values per cluster. This is best accomplished using a heatmap.

We have a framework for implementing heatmap visualizations, so there's not much to be done other than adding a function in `som_utils` to call it and adding the visualization in `example_flowsom_clustering`.


alex-l-kong:
Some things to check for this review:

- The new directory setting format
- The heatmap aesthetics (I'll be verifying correctness with Candace later). Should we have a standardized center value?

templates/example_flowsom_clustering.ipynb
```diff
@@ -47,9 +47,13 @@
    "outputs": [],
```
ngreenwald, line 47: <p>I think we should separate these directories into those that only have to be run once for preprocessing, and those that will be different for different combinations of parameters. </p><p> </p><p>For example, base_dir, tiff_dir, seg_dir, preprocessed_dir all belong to the first, subset, som, and consensus clustered belong to the second. </p><p> </p><p>For the first set of directories, we can auto-generate them based off the directory structure we assume people will have from running earlier parts of the pipeline. </p><p> </p><p>For the second set of directories, we should set a single name for that "experiment", like default_som_results. Then within that single folder, we'll auto-generate the subset_dir, som_clutered_dir, and consensus_clutered dir (and potentially others) in a known spot. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/410/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

ngreenwald, line 47: <p>I think we can remove channels as an argument to create_pixel_matrix, and preprocess all of the channels. We can then move the cell directly above this cell defining the subset_channels to below this cell, since it will only be needed by the train_som function. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/410/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

ngreenwald (changes requested):
Nice start. I think we can clean up the directory structure a bit more since the user won't care much about the specific sub-folders

ark/phenotyping/create_som_matrix.R
```diff
@@ -13,6 +13,8 @@ library(arrow)
 library(data.table)
 library(FlowSOM)
 
+set.seed(59)
```
ngreenwald, line 16: Same

ark/phenotyping/som_utils.py
```diff
@@ -9,14 +9,14 @@
...
 from ark.utils import load_utils
 from ark.utils import misc_utils
 
 
-def compute_cluster_avg(fovs, channels, base_dir,
-                        cluster_dir='pixel_mat_clustered',
-                        cluster_avg_name='pixel_cluster_avg.feather'):
+def compute_cluster_avg(fovs, channels, base_dir, cluster_col,
+                        cluster_dir='pixel_mat_clustered'):
     """Averages channel values across all fovs in pixel_mat_clustered
```
ngreenwald, line 20: A bit more description in docstring: The averages being computed are within each cluster

ark/phenotyping/som_utils.py
```diff
@@ -129,21 +126,19 @@ def create_fov_pixel_data(fov, channels, img_data, seg_labels,
     return pixel_mat, pixel_mat_subset
 
 
-def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
+def create_pixel_matrix(fovs, base_dir, tiff_dir, seg_dir,
                         pre_dir='pixel_mat_preprocessed',
                         sub_dir='pixel_mat_subsetted', is_mibitiff=False,
-                        blur_factor=2, subset_proportion=0.1, seed=42):
+                        blur_factor=2, subset_proportion=0.1, seed=None):
```
ngreenwald, line 132: A bit more description in docstring. What preprocessing are we doing?

ark/phenotyping/som_utils.py
```diff
@@ -267,6 +264,11 @@ def train_som(fovs, channels, base_dir,
     # run the SOM training process
     process_args = ['Rscript', '/create_som_matrix.R', ','.join(fovs), ','.join(channels),
                     str(num_passes), subsetted_path, norm_vals_path, weights_path]
+
+    # add a random seed to function call if set
+    if seed is not None:
```
ngreenwald, line 269: I think it would be better to just set a default seed in the function definition

ark/phenotyping/som_utils.py
```diff
@@ -406,6 +416,10 @@ def consensus_cluster(fovs, channels, base_dir, max_k=20, cap=3,
     process_args = ['Rscript', '/consensus_cluster.R', ','.join(fovs), ','.join(channels),
                     str(max_k), str(cap), clustered_path, cluster_avg_path, consensus_path]
 
+    # add a random seed to function call if set
+    if seed is not None:
```
ngreenwald, line 420: Same thing here

ark/phenotyping/som_utils_test.py
```diff
@@ -148,13 +148,13 @@ def test_compute_cluster_avg():
                                                                      fov + '.feather'))
 
         # compute cluster average matrix
-        som_utils.compute_cluster_avg(fovs, chans, temp_dir)
+        cluster_avg = som_utils.compute_cluster_avg(fovs, chans, temp_dir, 'cluster')
 
         # assert that pixel_cluster_avg.feather was actually created
-        assert os.path.exists(os.path.join(temp_dir, 'pixel_cluster_avg.feather'))
+        # assert os.path.exists(os.path.join(temp_dir, 'pixel_cluster_avg.feather'))
```
ngreenwald, line 154: Decide if we're keeping these

ark/utils/notebooks_test.py
```diff
@@ -25,17 +25,17 @@ def _exec_notebook(nb_filename):
         subprocess.check_call(args)
 
 
-# test runs with default inputs
-def test_segment_image_data():
-    _exec_notebook('Segment_Image_Data.ipynb')
+# # test runs with default inputs
+# def test_segment_image_data():
```
ngreenwald, line 29: Same

ngreenwald (changes requested):
Looks good! Some minor comments

ark/phenotyping/som_utils.py
alex-l-kong: My only concern is that having a default seed makes it less intuitive for the user to ask for a fully stochastic run if they wish (they'd have to know to pass in `None`). That being said, if we don't imagine many lab members will need this feature, better to remove it.

ngreenwald: Yeah, I think only Candace would want something like that for testing purposes

templates/example_flowsom_clustering.ipynb
```diff
@@ -18,6 +18,7 @@
...
```
ngreenwald, line 18: <p>This didn't really address my previous comment about separating data that needs to be generated each time, and data that only needs to be generated once. </p><p> </p><p>In addition, I think it makes more sense to put the SOM-specific directories after the cell where we specify which channels are going to be included in this run. We should give people the option to name the folder an informative name, and otherwise default to the timestamp-based scheme you have here. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/410/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

ngreenwald, line 18: <p>We can remove this now that we set a seed by default</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/410/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

ark/phenotyping/som_utils.py
```diff
@@ -52,22 +52,19 @@ def compute_cluster_avg(fovs, channels, base_dir,
...
-    # save the DataFrame
-    feather.write_dataframe(sum_count_totals,
-                            os.path.join(base_dir, cluster_avg_name),
-                            compression='uncompressed')
+    return sum_count_totals
 
 
 def create_fov_pixel_data(fov, channels, img_data, seg_labels,
-                          blur_factor=2, subset_proportion=0.1, seed=42):
+                          blur_factor=2, subset_proportion=0.1, seed=None):
```
ngreenwald, line 67: If I leave a comment about specific changes to make to multiple distinct functions in a given PR, you can assume that comment applies to the rest of the functions. For example, updating the docstrings. Having incorrect docstrings is worse than having no docstrings at all, like the mis-specified outputs for this one.

I think it would be good coding practice to start doing a self review of your PRs before tagging me. You can go through the files listed as changed and pick off the obvious stuff before I get to it.

ark/phenotyping/som_utils.py
```diff
@@ -129,21 +126,19 @@ def create_fov_pixel_data(fov, channels, img_data, seg_labels,
     return pixel_mat, pixel_mat_subset
 
 
-def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
+def create_pixel_matrix(fovs, base_dir, tiff_dir, seg_dir,
                         pre_dir='pixel_mat_preprocessed',
                         sub_dir='pixel_mat_subsetted', is_mibitiff=False,
-                        blur_factor=2, subset_proportion=0.1, seed=42):
-    """Preprocess the images for FlowSOM clustering and creates a pixel-level matrix
+                        blur_factor=2, subset_proportion=0.1, seed=None):
```
ngreenwald, line 132: Is there a reason some of the functions are initialized with None and some with 42?

ngreenwald (changes requested):
Some additional tweaks, overall pretty close

ark/phenotyping/som_utils.py
alex-l-kong: Can you clarify which part of the output for `create_fov_pixel_data` is mis-specified in the docstring for `create_fov_pixel_data`? I've just updated the return values to be more descriptive of the specific preprocessing going on and fixed `pd.DataFrame` to `pandas.DataFrame`. Not sure if we were also following a specific format for a `tuple` return type or not.

There was a return value missing for `compute_cluster_avg` (also fixed), did you by chance mean that function?

ngreenwald: Yup, it was the return value that I noticed, couldn't comment in the exact right spot bc it hadn't been changed in this PR.

ngreenwald (approved):
Looks good!

PR #418, opened 2021-04-21
Add metadata columns to SOM cluster and consensus cluster data

### LINKED ISSUE #416: Add pixel metadata to the SOM cluster and consensus cluster data

Although the preprocessed data for SOM clustering contains the metadata and can easily be mapped, for ease of use it makes sense to include them in the SOM and consensus cluster data too.

We read in the entire datasets for running the trained SOM and consensus clustering so we can access these metadata columns. The `markers` variables can be used for subsetting on channels for the actual computations.

alex-l-kong:

Addresses and closes #416. We add the metadata columns to both the SOM and consensus clustered data so it will be easy to access for users.

We change the way we read in the `feather` dataframes from subsetting on just channel columns to reading in everything. We then use the `markers` vectors to subset the data for computational purposes, but we add additional columns to the full data.

### DIFF
--- a/ark/phenotyping/run_trained_som.R
+++ b/ark/phenotyping/run_trained_som.R
@@ -60,7 +60,7 @@ for (i in 1:length(fovs)) {
     clusters <- FlowSOM:::MapDataToCodes(somWeights, as.matrix(fovPixelData[, markers]))
 
     # assign cluster labels column to pixel data
-    fovPixelData <- cbind(fovPixelData, cluster=clusters[,1])
+    fovPixelData$cluster <- clusters[,1]
 
     # write to feather
     clusterPath <- file.path(pixelClusterDir, fileName)
--- a/ark/phenotyping/run_trained_som.R
+++ b/ark/phenotyping/run_trained_som.R
@@ -44,31 +44,32 @@ markers <- colnames(somWeights)
 print("Mapping data to cluster labels")
 for (i in 1:length(fovs)) {
     # read in pixel data
-    fileName <- paste(fovs[i], ".feather", sep="")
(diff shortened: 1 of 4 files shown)


ark/phenotyping/run_trained_som.R
```diff
@@ -44,31 +44,32 @@ markers <- colnames(somWeights)
...
         }
     }
 
     # map FlowSOM data
-    clusters <- FlowSOM:::MapDataToCodes(somWeights, fovPixelData)
+    clusters <- FlowSOM:::MapDataToCodes(somWeights, as.matrix(fovPixelData[, markers]))
 
     # assign cluster labels column to pixel data
-    fovPixelData <- as.matrix(cbind(as.matrix(fovPixelData), cluster=clusters[,1]))
+    fovPixelData <- cbind(fovPixelData, cluster=clusters[,1])
```
ngreenwald, line 63: I think you said cbind was the slow part here, right? It is faster to initialize this matrix from the beginning with a blank consensus clustered column, and assign it here? Or is assignment equally slow

ngreenwald (changes requested):
Looks good, minor comment

ark/phenotyping/run_trained_som.R
alex-l-kong: Let me try using `$` assignment first; that's what Candace does in her scripts. Initializing this matrix with a blank column and assigning later would essentially be the same, just with an extra step.

ngreenwald: Sounds good. There's probably some best practices about fastest way to update dataframes

ngreenwald (approved):

PR #422, opened 2021-05-26
Nuclear fix

### LINKED ISSUE #419: exporting multi-compartment morphological measurements
*Please refer to our [FAQ](https://ark-analysis.readthedocs.io/en/latest/_rtd/faq.html) and look at our [known issues](https://github.com/angelolab/ark-analysis/issues) before opening a new issue*

I am trying to export the nuclear and cell areas, as well as the nc_ratio.

**What is the problem?**
I see these metrics are computed, but I haven't been able to figure out how to export them. When I run the Segment_Image_data script, I can see a lot of the new morphological metrics (convex hull, etc), but I am not able to see the nuclear/cell area or nc_ratio.

**What have you tried so far?**
I've tried running the script on our data, as well as the example dataset. I've also tried adding in "nuclear_counts" as an argument into the ```marker_quantification.generate_cell_table``` function, since this flag seems to be needed for generating the multi-compartment morphological measurements. However, when I do this and set nuclear_counts=True, the code hangs. Maybe I didn't wait long enough (I waited >30 mins)?

alex-l-kong:

Closes #419. `nc_ratio` is currently taking a very long time to compute. We address this bottleneck in this PR.

The original method worked well for small datasets but didn't scale well due to the inefficiencies of looping through hundreds to thousands of cells. To avoid the bottleneck of looping through each cell, we instead broadcast this operation.

**Remaining issues**

The actual accuracy of the values still needs to be checked. In addition to the test data provided in `ark`, it would help to test on maybe a couple other datasets cause it looked like @awedwards was having issues of his own on his dataset.


alex-l-kong:
Sorry for the delay, but this version should run much faster. I'm still verifying accuracy, @awedwards if you wouldn't mind testing on your own dataset that would be very helpful.

ngreenwald:
Okay, I see the problem. For generating the figures for the Mesmer paper, I computed NC ratio the same way you now have it, outside of the loop and all at once. However, the first pass at the refactor to include these morphology metrics in ark computed it the same way the other stuff, cell-by-cell, leading to the slowdown. 

The next thing to check is why the wrong results are being returned, the indexing errors seems like the likely culprit.

templates/Segment_Image_Data.ipynb
```diff
@@ -281,12 +281,14 @@
...
     "                                                    xr_dim_name='compartments',\n",
     "                                                    xr_channel_names=['whole_cell'],\n",
+    "                                                    files=[fov + '_feature_0.tif' for fov in fovs],\n",
     "                                                    delimiter='_feature_0',\n",
     "                                                    force_ints=True)\n",
     "\n",
     "segmentation_labels_nuc = load_utils.load_imgs_from_dir(data_dir=deepcell_output_dir,\n",
     "                                                    xr_dim_name='compartments',\n",
     "                                                    xr_channel_names=['nuclear'],\n",
+    "                                                    files=[fov + '_feature_1.tif' for fov in fovs],\n",
```
ngreenwald, line 291: We need to address the underlying issue with the load_imgs function, otherwise this is just going to happen again. 

No error message was thrown when the delimiter didn't match the file name. Was it you or @ackagel who added the `delimiter_optional` flag to the `extract_delimited_names` function? What was the reasoning behind that?

ngreenwald (changes requested):

templates/Segment_Image_Data.ipynb
ackagel: Sounds my (apparently not so handi)-work.  I'll look into that, quickly rework the functionality, and rewrite/reimplement the test functions which should've caught this failure.

ackagel: @alex-l-kong I'll work on a PR for that and merge it into this branch once it's approved.  Sounds good?

alex-l-kong:
@awedwards for the time being, we've implemented a temporary fix on this branch addressing the `nc_ratios` always equalling 1 bug, so you can test the new process on this branch. @ackagel is working on a more permanent fix to this issue.

PR #424, opened 2021-06-03
File delimiter fix

ackagel:
## **Purpose**

The functionality of the 'delimiter' argument in `load_utils.load_imgs_from_dir()` is fairly ambiguous.  This PR removes the argument and replaces it with two clearer substitutes.

## **Implementation**

Swapped the 'delimiter' argument for `match_substring` and `trim_suffix` arguments.

The former is a mandatory match, in that all loaded filenames (excluding their extensions) must contain the value passed to `match_substring`.  This is ignored if files are directly provided via the `files` argument.

The latter mimics the original intended functionality of the 'delimiter' argument.  It will remove the value passed to `trim_suffix` from the end of each filename if it's present, and will otherwise leave the filename unchanged.

Tests for the `match_substring` argument were included; testing for `trim_suffix` was already implemented as `delimiter` argument tests.

## **Remaining issues**

The `match_substring` argument could probably be swapped with a regex expression for more flexibility, but that requires users to know regex.  I could add it in addition to `match_substring`, just as a more advanced tool for users whose filenames are needlessly overcomplicated, but that might be beyond the scope of this PR.


ackagel:
oops, forgot to check if the build passed before requesting review.  Looks like I forgot to change the injected notebook cells in the notebook test functions so I'll add that real quick.  Still welcome to a review on the implementation here.

ngreenwald:
I think it would be more general if this was match_substring rather than match_suffix. What do you think?

ngreenwald (changes requested):
Overall looks good. Someone did `git add *` however....

ackagel:
nah that's a really important `.DS_Store`.  Definitely not one to feel shameful about.....

PR #426, opened 2021-06-29
Optimize loading procedures in image segmentation notebook

### LINKED ISSUE #414: Docker memory size requirement calculation
What is the memory size needed to be allocated to Docker for segmentation on a set of images? For example, if I have a set of 50 images that I want the segmentation script to process altogether at once, and each image is around 17MB, how much memory would have to be allocated to Docker?

I am always unsure of how much memory is needed by Docker to segment my image sets. It is difficult to decide on which computer the segmentation script should be run when the required memory size is unknown.

The segmentation script reads in the image set and outputs an estimation of the memory size needed.

Given a formula to calculate the memory size needed, so we can make an estimation ourselves.

Thank you!

### LINKED ISSUE #415: Add additional batch-processing to segmentation notebook
The current segmentation notebook has some steps which can be batch-processed, but some which cannot. This means that the total dataset size is still a limiting factor for processing a collection of images, which leads to memory errors (see #414)

Rather than computing the amount of memory required, we should update the segmentation script to allow a batch size for all of the functions. 

Specifically, we'll need to do the following:
1. Modify ```generate_deepcell_input``` to take a path to the data_dir, rather than the full xarray. It will then read the data in one batch size at a time and process it
2. Remove the cell ("We can then load the segmented mask from deepcell via label-map TIFFs and save as an xarray") that loads all of the data and converts to an xarray. Instead, we can just load the first N images and use that for the subsequent script which plots example visualizations
3. Modify generate_cell_table to take a path to the segmentation directory, rather than an xarray with all of the labels.

alex-l-kong:

Addresses and closes #415. By extension, addresses and closes #414. In several parts of `Segment_Image_Data.ipynb`, we load all images/segmentations into memory. This is highly inefficient and will cause crashes for large image datasets. This PR will convert this to batch-loading.

We remove the loading functions to the helper functions where the images/labels are needed. These functions will call the necessary `load_imgs_from` function to retrieve the data for only some of the `fovs` at once. This will not only prevent crashes but also allow for a much more efficient use of Docker space.

**Remaining issues**

The batching process may still have a few inefficiencies and there may be places where some values are hard-coded.


ngreenwald:
This is another good one for @bcollica and @ackagel to review first since it's fairly self-contained, please take a look

bcollica (commented):
Given that the batch processing feature is meant to prevent crashes and use less memory, it might be helpful to add a check of some kind regarding the batch size argument in generate_cell_table(), unless I missed it somewhere.  Because if the user just sets the batch size to the total number of images then it defeats the purpose of the feature.  Was there was an upper limit to the number of files that could be processed without crashing?

ark/utils/plot_utils.py
```diff
@@ -85,26 +84,84 @@ def tif_overlay_preprocess(segmentation_labels, plotting_tif):
...
 
+    # load the specified fov data in
+    plotting_tif = load_utils.load_imgs_from_dir(
+        data_dir=data_dir,
+        files=[fov + '.tif'],
+        xr_dim_name='channels',
+        xr_channel_names=['nuclear_channel', 'membrane_channel']
+    )
+
+    # verify that the provided image channels exist in plotting_tif
```
ngreenwald, line 120: I don't understand why this doesn't fail. We name the channels 'nuclear_channel' and 'membrane_channel', so if the user passes a channel name this isn't going to match. But in the notebook_utils_test this passes when you provide ['chan0', 'chan1'] as an input

ngreenwald (changes requested):
Looks great overall

ark/segmentation/marker_quantification_test.py
```diff
@@ -549,47 +550,64 @@ def test_generate_cell_data_tree_loading():
...
+        cell_masks = np.zeros((3, 40, 40, 2), dtype="int16")
         cell_masks[0, :, :, 0] = cell_mask[0, :, :, 0]
         cell_masks[1, 5:, 5:, 0] = cell_mask[0, :-5, :-5, 0]
         cell_masks[2, 10:, 10:, 0] = cell_mask[0, :-10, :-10, 0]
+        cell_masks[..., 1] = cell_masks[..., 0]
 
         segmentation_masks = test_utils.make_labels_xarray(
             label_data=cell_masks,
-            compartment_names=['whole_cell']
+            compartment_names=['whole_cell', 'nuclear']
```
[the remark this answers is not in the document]
alex-l-kong: I create `segmentation_masks` to generate data that I can write to a directory. This directory with the written labels then gets passed to `generate_cell_table`.

If preferable, I can simply generate a random 1024x1024 `numpy` array for each mask needed instead.

ark/utils/plot_utils.py
alex-l-kong: The reason it didn't fail was because I never actually injected the call into the notebook. When I did, it failed spectacularly. Fixed.

ark/segmentation/marker_quantification_test.py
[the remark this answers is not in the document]
ngreenwald: Isn't that what `cell_masks` is? The data? I think `segmentation_masks` is just an xarray version of `cell_masks`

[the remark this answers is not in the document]
alex-l-kong: @ngreenwald good point, I'll just use `cell_masks`.

ark/segmentation/marker_quantification_test.py
```diff
@@ -549,47 +550,58 @@ def test_generate_cell_data_tree_loading():
...
         assert arcsinh_data.shape[0] > 0 and arcsinh_data.shape[1] > 0
 
+        # test nuclear_counts True
+        norm_data, arcsinh_data = marker_quantification.generate_cell_table(
+            segmentation_dir=temp_dir, tiff_dir=tiff_dir,
+            img_sub_folder=img_sub_folder, is_mibitiff=False, fovs=fovs_subset,
+            batch_size=2, nuclear_counts=True)
+
+        assert norm_data.shape[0] > 0 and norm_data.shape[1] > 0
+        assert arcsinh_data.shape[0] > 0 and arcsinh_data.shape[1] > 0
```
ngreenwald, line 603: This is just a copy paste of the same tests from without nuclear counts. This doesn't check that nuclear counts are returned. We should check that the shape is correct given that nuclear counts were generated.

ark/segmentation/marker_quantification_test.py
```diff
@@ -619,39 +631,51 @@ def test_generate_cell_data_mibitiff_loading():
...
         assert norm_data.shape[0] > 0 and norm_data.shape[1] > 0
         assert arcsinh_data.shape[0] > 0 and arcsinh_data.shape[1] > 0
 
+        # test nuclear_counts True
+        norm_data, arcsinh_data = marker_quantification.generate_cell_table(
+            segmentation_dir=temp_dir, tiff_dir=tiff_dir,
+            img_sub_folder=tiff_dir, is_mibitiff=True, fovs=fovs_subset,
+            batch_size=2, nuclear_counts=True)
+
+        assert norm_data.shape[0] > 0 and norm_data.shape[1] > 0
```
ngreenwald, line 676: Same here

ark/segmentation/marker_quantification_test.py
```diff
@@ -706,10 +732,20 @@ def test_generate_cell_data_extractions():
...
         )
 
         assert np.all(positive_pixel_data.iloc[:4][['chan0', 'chan1']].values == 0)
         assert np.all(positive_pixel_data.iloc[4:][chans].values == 1)
+
+        # verify thresh kwarg passes through and nuclear counts True
+        positive_pixel_data, _ = marker_quantification.generate_cell_table(
+            segmentation_dir=temp_dir, tiff_dir=tiff_dir,
+            img_sub_folder=img_sub_folder, is_mibitiff=False, batch_size=2,
+            extraction='positive_pixel', nuclear_counts=True, **thresh_kwargs
```
ngreenwald, line 747: And here

ark/utils/notebooks_test_utils.py
```diff
@@ -1,6 +1,7 @@
 import os
 import numpy as np
 import skimage.io as io
+import timeit
```
ngreenwald, line 4: Do we need this?

ngreenwald (changes requested):
Looks good, a few minor things for testings

ngreenwald (approved):
Looks good!

Issue #413, opened 2021-03-18
MIBI QC metrics


ngreenwald:
We may end up putting this code somewhere else, but I wanted to write down the main discussion points from our meeting with Mike. 

- Potential sources of systematic noise in the image: detector non-uniformity, staining differences, sample-preservation
- Is there a way we can identify each of these different noise sources, and set reasonable cutoffs to determine when certain images have exceeded them
- Tiling script for commercial instrument and Betty in Python to set XML automatically
- Script to combine different directories and XMLs into a single run
- MIBITiff -> folders of TIFs converter
- Assessing batch issues across FOVs: mean pixel intensity per channel per image, summarized by covariates (e.g. run, date, position)

PR #427, opened 2021-07-08
Add MIBI tiling script

alex-l-kong:

Begins addressing #413. We start by adding a tiling script to `ark` allowing users to generate tiled regions across fovs for a given run.

We add `example_fov_grid_generate.ipynb`, which runs the tiling process. A new repo, `ark/mibi`, will contain helper functions. So far, only `tiling_utils.py` has been added to encapsulate the tiling process.

Additionally, sample JSON data has been added to `data/example_dataset`.

**Remaining issues**

Testing still needs to be addressed.


alex-l-kong:
The tests have yet to be added in, but I would appreciate an initial review of the overall structure. I'll begin tagging Erin on this PR after more progress.

ngreenwald (changes requested):
Good start, see below for some initial comments

alex-l-kong:
I'm researching ways to test `set_tiling_params` because those require input arguments from the user. In the meantime, a test for `create_tiled_regions` has been added.

ark/mibi/tiling_utils.py
```diff
@@ -0,0 +1,218 @@
...
+    with open(fov_list_path, 'r') as flf:
+        fov_tile_info = json.load(flf)
+
+    # define the parameter dict to return
+    tiling_params = {}
+
+    # retrieve the format version
+    tiling_params['fovFormatVersion'] = fov_tile_info['fovFormatVersion']
+
+    # define lists to hold the starting x and y coordinates for each fov
```
ngreenwald, line 37:
```suggestion
    # define lists to hold the starting x and y coordinates for each region
```

ark/mibi/tiling_utils.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+def set_tiling_params(fov_list_path):
+    """Given a file specifying fov regions, set the MIBI tiling parameters
+
+    User inputs will be required for many values
+
+    Args:
+        fov_list_path (str):
+            Path to the JSON file containing the fov region data
```
ngreenwald, line 16:
```suggestion
            Path to the JSON file containing the fovs used to define each tiled region
```

ark/mibi/tiling_utils.py
```diff
@@ -0,0 +1,218 @@
...
+
+        while size_y < 1:
+            print("Error: step size must be positive")
+            size_y = int(input("Enter the y step size for region %s: " % fov['name']))
+
+        x_fov_size.append(size_x)
+        y_fov_size.append(size_y)
+
+    # allow the user to specify if the FOVs should be randomized
+    randomize = int(input("Randomize? Enter 0 for no and 1 for yes: "))
```
ngreenwald, line 86: This can be in the for loop as well

ark/mibi/tiling_utils.py
```diff
@@ -0,0 +1,218 @@
...
+        'exportDateTime': dt,
+        'fovFormatVersion': tiling_params['fovFormatVersion'],
+        'fovs': []
+    }
+
+    # define a counter to determine where to insert a moly point
+    # only used if tiling_params['moly_interval'] is set
+    moly_counter = 0
+
+    # append tile information to fov_list, iterate through each tiles iteratively
```
ngreenwald, line 172: This comment could be more clear. We're iterating through each tiled region here, not each tile

ark/mibi/tiling_utils.py
```diff
@@ -0,0 +1,218 @@
...
+        for t in combinations(lists, 2):
+            for pair in product(*t):
+                yield pair
+
+    # file path validation
+    if not os.path.exists(moly_path):
+        raise FileNotFoundError("Moly point file %s does not exist" % moly_path)
+
+    # read in the data
+    with open(moly_path, 'r') as mf:
```
ngreenwald, line 155: Is there a reason it makes more sense to do this here, rather than in the previous function?

ark/mibi/tiling_utils.py
```diff
@@ -0,0 +1,218 @@
...
+            # increment moly_counter as we've added another fov
+            moly_counter += 1
+
+            # append a moly point if moly_interval is set and we've reached the interval threshold
+            if 'moly_interval' in tiling_params and \
+               moly_counter % tiling_params['moly_interval'] == 0:
+                tiled_regions['fovs'].append(moly_point)
+
+        # append moly point to seperate runs if not last
+        if region_index != len(tiling_params['fov_num_x']) - 1:
```
ngreenwald, line 215: This can also be an argument set in previous function that controls whether to insert moly between distinct regions

ngreenwald (changes requested):
Looks good, some more tweaks

ark/mibi/tiling_utils.py
alex-l-kong: We can do it in the previous function and have it return both `tiling_params` and `moly_point`.

ngreenwald: It seems more natural to supply all of the data in the first function, since we're already providing the FOV list

ark/mibi/tiling_utils.py
alex-l-kong: I believe regardless of the other parameters set, moly points always need to be inserted between distinct regions. @efrancis28 can you clarify if this is the case?

ark/mibi/tiling_utils.py
```diff
@@ -0,0 +1,240 @@
...
+    while moly_run_insert not in [0, 1]:
+        print("Error: moly point run parameter must be either 0 or 1")
+        moly_run_insert = int(
+            input("Insert moly points between runs? Enter 0 for no and 1 for yes: ")
+        )
+
+    tiling_params['moly_run'] = moly_run_insert
+
+    # whether to insert moly points between tiles
+    # NOTE: moly points will be inserted between different runs regardless of what's set here
```
ngreenwald, line 135: Not anymore

ark/mibi/tiling_utils_test.py
```diff
@@ -0,0 +1,550 @@
...
+                'slideId': 5931,
+                'name': 'TheFirstFOV',
+                'timingDescription': '1 ms'
+            },
+            {
+                'scanCount': 1,
+                'centerPointMicrons': {
+                    'x': 100, 'y': 100
+                },
+                'timingChoice': 7,
```
ngreenwald, line 190: I think we can move this stuff to a helper function that creates a JSON file. Can probably combine this with the JSON file in previous function

ark/mibi/tiling_utils_test.py
```diff
@@ -0,0 +1,550 @@
...
+        sorted_center_points_run = [
+            (fov['centerPointMicrons']['x'], fov['centerPointMicrons']['y'])
+            for fov in tiled_regions_base['fovs']
+        ]
+
+        # define the actual center points in the same expected order
+        actual_center_points_run = [
+            (0, 100), (0, 110), (0, 120), (0, 130), (5, 100),
+            (5, 110), (5, 120), (5, 130), (14540, -10830),
+            (50, 150), (50, 155), (60, 150), (60, 155), (70, 150),
```
ngreenwald, line 261: This is pretty opaque. Is there a way to programatically generate this instead?

ark/mibi/tiling_utils_test.py
```diff
@@ -0,0 +1,550 @@
...
+        misc_utils.verify_same_elements(
+            random_center_points=random_center_points_run,
+            sorted_center_points=sorted_center_points_run
+        )
+
+        # the random center points should be ordered differently than the sorted center points
+        # due to the nature of randomization, this test will fail once in a blue moon
+        assert random_center_points_run != sorted_center_points_run
+
+        # test 12: randomization for both fovs, additional moly point interval,
```
ngreenwald, line 525: These tests are going to be super hard to maintain. We have 12 copies of almost the exact same code, each testing one of the possible permutations to the inputs. Given that these inputs don't actually interact with each other, I don't think it's necessary to test every possible combination. 

Either way, this is begging for a for loop. We should loop over the possible values of randomization (yes/no), molly point (yes/no), etc. Then, based on the condition, check whether it matches expected. This would collapse all 12 of these tests into a single test.

ngreenwald (changes requested):
Looks good. The testing needs to be refactored.

ark/mibi/tiling_utils.py
```diff
@@ -0,0 +1,240 @@
...
+    # read in the data for each fov (region_start from fov_list_path, fov_num from user)
+    for fov in fov_tile_info['fovs']:
+        region_start_x.append(fov['centerPointMicrons']['x'])
+        region_start_y.append(fov['centerPointMicrons']['y'])
+
+        num_x = int(input("Enter number of x fovs for region %s: " % fov['name']))
+
+        while num_x < 1:
+            print("Error: number of fovs must be positive")
+            num_x = int(input("Enter number of x fovs for region %s: " % fov['name']))
```
ackagel, line 73: this can be condensed w/ do-while pattern
```
while True:
    num_x = int(input("..."))
    if num_x < 1:
        print("Error: ...")
```

ark/mibi/tiling_utils.py
```diff
@@ -0,0 +1,240 @@
...
+    Returns:
+        dict:
+            Data containing information about each tile, will be saved to JSON
+    """
+
+    # helper function for creating all pairs between two lists
+    def pairs(*lists):
+        for t in combinations(lists, 2):
+            for pair in product(*t):
+                yield pair
```
ackagel, line 178: since this helper function doesn't depend on args/vars in the function scope, it probs should be moved outside the function

ark/mibi/tiling_utils.py
```diff
@@ -0,0 +1,240 @@
...
+    # iterate through each region and append created tiles to tiled_regions['fovs']
+    for region_index in range(len(tiling_params['fov_num_x'])):
+        # extract start coordinates
+        start_x = tiling_params['region_start_x'][region_index]
+        start_y = tiling_params['region_start_y'][region_index]
+
+        # generate range of x and y coordinates
+        x_range = list(range(tiling_params['fov_num_x'][region_index]))
+        y_range = list(range(tiling_params['fov_num_y'][region_index]))
+
```
ackagel, line 203: Noticing a pattern where most tiling params need to be accessed as `tiling_params['some_parameter'][region_index]`.  If the structure of `tiling_params` is fixed that's fine, but would there be any downside changing the `tiling_params` output of `set_tiling_params` to be structured like:

```
{
    ...,
    fovs: ...,
    region_params: [{
        region_start_x: 0,
        region_start_y: 0,
        ...,
    }, ...],
    ...,
}
```
instead of:
```
{
    ...,
    fovs: ...,
    region_start_x: [...],
    region_start_y: [...],
    ...,
}
```
for the syntax simplification:
```
for region_params in tiling_params['region_params']:
    start_x = region_params['region_start_x']
    ...
```

ackagel (commented):
Just some thoughts on the `tiling_params` structure.  Agree with Noah's point on condensing/generalizing the test functions.

ark/mibi/tiling_utils_test.py
alex-l-kong: I've definitely been meaning to get to this, the tests are only this way right now to make it super clear what each is testing.

ark/mibi/tiling_utils_test.py
alex-l-kong: As in something like: `[(x, y) for x in np.arange(0, 10, 5) for y in np.arange(100, 140, 10)] + [(14540, -10830)] + ...`?

ark/mibi/tiling_utils.py
alex-l-kong: This do-while pattern will never exit regardless of what `num_x` is set to. You would need to add an `else: break` statement after the `if` statement for this to work, but I'm trying to avoid `break` statements as much as possible here.

ackagel: oops sorry I think I meant:
```
while True:
    num_x = int(input("..."))
    if num_x >= 1:
        break
    print("Error: ...")
```
regardless, since this 'force integer input' pattern is used a lot, it might be worth putting some version of it into its own separate function:
```
def int_input(msg, error_msg, condition):
    out = int(input(msg))
    if condition(out):
        return out
    print(error_msg)
```
which you could use like:
```
num_x = int_input('...', 'Error: ...', lambda x: x >= 1)
```

ark/mibi/tiling_utils_test.py
ngreenwald: Yes exactly. Ideally with variables instead of constants which you define at the top. We're going to have the exact same type of test in the TMA tiling script, where want to check the coordinates were set appropriately

ark/mibi/tiling_utils_test.py
ngreenwald: Sounds good

bcollica (commented):
Just minor stuff

bcollica (commented):
Looks amazing!  My only thought would be that `set_tiling_params()` is a bit long at 125 lines, mostly due to the repetitive nature of the calls to `read_tiling_param()`.  Breaking those calls out as their own separate functions might help condense it, but not sure if it's really worth the trouble?

ngreenwald (approved):
Looks good!

PR #431, opened 2021-07-22
Create pixel and cell cluster overlays

### LINKED ISSUE #351: Visualization for cell phenotypes
To help people explore the accuracy of the pixel clusters, the following visualizations would be helpful.

- Show metrics (correlation, spatial enrichment, segmentation stats) from two specific clusters
- Color segmentation labels based on specified set of pixel clusters
- Create an overlay with specific clusters overlaid over image data of specific channels

### LINKED ISSUE #429: Create pixel mask for visualizing overlay of pixel clusters

Related to issue #351. For visualizing the overlay of pixel clusters, we need to create a mask that assigns each pixel in an image to its respective SOM or meta cluster label.

Ideally, this happens after running consensus labeling. We can use the data in `pixel_mat_consensus` to create this mask as it contains both the pixel SOM and meta cluster labels per cell.

alex-l-kong:

Closes #351 . Closes #429. While Zak from Toptal works on the other visualizations, the one we can handle on our end are the cluster label overlays for both pixels and cells. We add code to generate masks for both pixels and cells and add calls in `example_pixel_cell_clustering` to generate the visualizations.

For pixel clusters, we need to create a pixel mask which is similar to a cell mask except on a pixel-level granularity. For cell clusters, we can mostly make use of Omer's code in `data_utils` to generate a mask, but we have to do a little more work since we have to load in the necessary data beforehand.

Omer wrote a nice overlay function in `plot_utils` which we use to help us with both overlays.

**Remaining issues**

Still flushing out the tests.


alex-l-kong:
Still need to work on the tests, but let me know how the overlays and the structure looks!

ark/utils/data_utils.py
```diff
@@ -46,7 +47,131 @@ def label_cells_by_cluster(fovs, all_data, label_maps, fov_col=settings.FOV_ID,
...
+        raise FileNotFoundError(
+            "consensus_dir %s does not exist in base_dir %s" % (cell_consensus_name, base_dir)
+        )
+
+    # load the consensus data in
+    cell_consensus_data = feather.read_dataframe(os.path.join(base_dir, cell_consensus_name))
+
+    # define the files for whole cell and nuclear
+    whole_cell_files = [fov + '_feature_0.tif' for fov in fovs]
+    # nuclear_files = [fov + '_feature_1.tif' for fov in fovs]
```
ngreenwald, line 91: I don't think we'll need to label the nuclei for anything

ark/utils/data_utils.py
```diff
@@ -46,7 +47,131 @@ def label_cells_by_cluster(fovs, all_data, label_maps, fov_col=settings.FOV_ID,
...
+
+    # load the segmentation labels in
+    label_maps = load_utils.load_imgs_from_dir(data_dir=seg_dir,
+                                               files=whole_cell_files,
+                                               xr_dim_name='compartments',
+                                               xr_channel_names=['whole_cell'],
+                                               trim_suffix='_feature_0',
+                                               force_ints=True)
+
+    # use label_cells_by_cluster to create cell masks
```
ngreenwald, line 101: We can `verify` here that the supplied FOVs all appear in the `cell_consensus_data`

ark/utils/data_utils.py
```diff
@@ -46,7 +47,131 @@ def label_cells_by_cluster(fovs, all_data, label_maps, fov_col=settings.FOV_ID,
...
+    img_data = []
+
+    for fov in fovs:
+        # read the pixel data for the fov
+        fov_data = feather.read_dataframe(
+            os.path.join(base_dir, pixel_consensus_dir, fov + '.feather')
+        )
+
+        # read the segmentation mask to determine size of pixel cluster mask
+        seg_mask = io.imread(os.path.join(seg_dir, fov + '_feature_0.tif'))
```
ngreenwald, line 156: Is the reason we can't use the dimensions of the feather file due to zero pixels being excluded?

ark/utils/data_utils.py
```diff
@@ -46,7 +47,131 @@ def label_cells_by_cluster(fovs, all_data, label_maps, fov_col=settings.FOV_ID,
...
+    Args:
+        fovs (list):
+            List of fovs to relabel
+        base_dir (str):
+            The path to the data directory
+        seg_dir (str):
+            The path to the segmentation data
+        pixel_consensus_dir (str):
+            The path to the data with both pixel SOM and meta cluster assignments
+        cluster_col (str):
```
ngreenwald, line 123: Should check that this argument is correct

ngreenwald (changes requested):
Looks great. Lets also add an optional cell for both pixel and cell clustering that saves the masks as .tiff files

ark/utils/data_utils.py
alex-l-kong: Yes, because of this we cannot assume that the pixel data for a fov will have the same dimensions as the segmentation mask.

ngreenwald (approved):
Looks good!

PR #432, opened 2021-07-23
Metacluster gui

y2kbugger:
## Slack
Full development notes in slack: https://toptal.slack.com/archives/C02762FAZB6/p1626918772008500

## Launching
I have added jupyerlab to the requirements file because it gives more horizontal space to use the GUI, to try this for yourself, after starting jupyter you can change the url to `lab` instead of `tree`.

e.g. `http://127.0.0.1:8888/tree/` -> `http://127.0.0.1:8888/lab/`

## Usage
- To select one or more clusters, click anywhere on either heatmap. Clicking a selected cluster will deselect it.
- To remap the selected clusters, right click anywhere on either heatmap.
- The bars at the top show the relative size of each cluster.
- If you remap all cluster in a metacluster, the metacluster is now *gone*
- You can click new metacluster to add current selection to a new metacluster
- You final mapping will be in an output file in the same directory as the input files

image
image

## Backlog

- Todo Next
  - barchart for aggregate pixelsize of metaclusters
  - order rows by hcluster ordering | 5
  - metacluster name labels | 2
  - change names | 5

- bug
  - ensure stable sorts
- feature
  - hide/show some markers
  - undo | 8
  - alternatively add a prompt for (are you sure? before remapping) | 2
  - add statistics to highlight potential remaps | 5 -


ngreenwald (changes requested):
Looks great so far! Couple suggestions for functionality
1. Add a y axis label for barchart counts on top of each column
2. Add a rowname for the metacluster row and selection row

Once you're happy with the functionality, some code/documentation suggestions:
1. Please pull out all of the functions into a separate .py file that gets imported at the top
2. Add a markdown cell with a brief description of how to work with the GUI
3. Change the column name for clustering into variable (instead of hardcoding `hCluster_cap`)
4. Input validation for supplied data to make sure it's the appropriate type, has the required column names, etc

y2kbugger:
Thanks for the feedback I've added the functionality requests to the todo, they should be straightforward.

@ngreenwald I've added my "maintainability" backlog above since it seems to be appreciated. I laughed when I compared our lists, our first three were exactly the same:

## design/maintainability
-   **move object code to separate  file**
-   **remap hCluster_cap upon loading, allow to pass alternate name as to not hardcode**
-   **Add usage docs**
-   Update/finalize maintenance docs/comments
-   could reduce recalculation with caching of properties
    -   e.g. metaclusters, expire on mapping changes
-   put all references that peer into mcd behind methods, e.g. columns, etc
-   plot only dummy data in the create gui, move as much as possible to "update_gui" (DRY)

cliu72 (commented):
Looks awesome! Agree with everything that's been said, only other suggestion is to add the dendrogram/hierarchical clustering of the markers.

y2kbugger:
Are you all fine with leaving jupyterlab as a required dependency? If so, I can think of two ways forward:

1. Change default launcher to `jupyter lab` (instead of `jupyter notebook`)
    - I'd need some help testing that all other notebooks continue to work as expected.
2. I'll just leave my instructions in a markdown cell about how to open Jupyter Lab using the URL address bar.

Again the reason for this is because it gives more horizontal room for the Notebook based GUI.

y2kbugger:
Regarding the ability to give metaclusters human readable names:

Are the names needed in an output, or should they only live as long as you are running the interactive session?

y2kbugger:
## 2021-07-25
- pixel count data labels | 3
- move code to imported python package
- move usage docs to example notebook
- row labels for metacluster and selection | 1
- Ensure color stability of metaclusters
- Make metacluster plot via cluster number rather than index

image

ngreenwald:
I tried jupyter lab for like 2 minutes and couldn't get the notebook to work, so went back to regular. I don't think having the scroll bar was a big deal, so my preference would be to keep things simple. Up to you Candace. 

Re: cluster names, it would definitely be useful if the user-generated names replaced the current default names in the output.

cliu72:
I was able to get jupyter lab working after rebuilding the docker without issue. I do think being able to view both heatmaps together without scrolling is a big advantage, since we want to see how the metacluster heatmap changes after re-assigning something, and it may get annoying to scroll back and forth. If it's too much trouble to switch the default to jupyter labs (Noah, not sure what the error is for you), we can just leave the instructions at the top of this notebook. If a user can't get jupyter labs to work, they can always just deal with the scrolling in jupyter notebook.

For the cluster names, it would be great if you could output the user-generated names, so the final output is a file with 100 rows and 3 columns: cluster number (1-100), metacluster number, and metacluster name.

Also not sure if I missed this somewhere in your notes, but it would be great if the default colors were blue-white-red (easier to visualize z-scores like that).

y2kbugger:
## Jupyter lab 
either way is fine with me. If we're trying to keep it simple for the user,  I think chances are good it would be safe to launch it as Jupyter Lab from the start docker script. I think in general Jupyter Lab is the preferred way of using notebooks when working with multiple open notebooks.

Official docs:

> image

## Saving names
Got it, I'll design the naming so with saving in mind.

## Color scheme
Blue and white is already the default, I'll make sure to document the parameters better in the usage cell. I noticed that I get a better sense of magnitude with a non-divergent color scheme (especially while tweaking zscore in realtime with slider), so I overrode the default (with "viridis") for myself when developing and forgot to remove. Let me know what you think. I can make any Matplotlib scheme the default.

> image

PS. Thinking about it, it could be that the divergent heatmap is not mapping blue-white-red to min-zero-max, but instead giving a linear mapping of blue-white-red across min-max. Let me know if you want me to spend some time looking into this.

## Marker Dendrogram
PPS. Still cooking about ideas for how to best show dendrogram for markers. I think i could draw the tree. I could also just order the rows and give indication of hierarchy in the label name (uglier, might be less effort if drawing tree to the independent mpl ax  turns out to be hard). Main decision to be made is: should this part of pipeline rerun clusterering or take the heirarchy as an input. I think you are using some built in clustering from seaborn, correct me if wrong. I can look into how to save out the dendrogram structure if you want.

ngreenwald:
Sure thing, sounds good. Can you update the docker script and I'll check out the other notebooks to make sure nothing weird happened?

y2kbugger:
> Sure thing, sounds good. Can you update the docker script and I'll check out the other notebooks to make sure nothing weird happened?

Done, should be working by default after pull and rebuild.

ngreenwald:
This is the [seaborn clustermap](https://seaborn.pydata.org/generated/seaborn.clustermap.html) function candace was referring to. No issues with our other notebooks, I think switching to Jupyter Lab is totally fine.

y2kbugger:
Great I'll leave that in.

y2kbugger:
## 2021-07-28
- Setup local envs to enable linting/refactoring in vscode
- Clean up accidentally committed ipynb in wrong folder
- Working through lint

## 2021-07-29
- Fixup and put in pull request for enhanced Jupyter launching
- fix linting errors | 1
- Started adding unit and functional pytests
  - Add pure constructors for testing with simple dummy data

y2kbugger:
## 2021-07-31

- add test cases | 2
- remap hCluster_cap upon loading, allow to pass alternate name as to not hardcode
- Add backend for metacluster displaynames

y2kbugger:
## 2021-08-05
- Add unit tests for GUI w/ Virtual clicks

y2kbugger:
## 2021-08-09

- Finish unit tests for GUI
- add tests for distinct colormap
- add tests for throttle
- Clean up asyncio throttle implementation
- Achieve acceptable test coverage
- Start bar charts for metacluster totals

y2kbugger:
**2021-08-12**
- Test for weighted average/metaclustering
- Implement colormap normalizer to keep the dynamic range good, but keep the diverging heatmap centered about 0.
- Add mc names to output file
- Optimize frequency out outputfile saves
- Update usage documentation in Notebook

y2kbugger:
## 2021-08-17
- Order rows by hcluster ordering / draw marker dendrogram| 7

image

y2kbugger:
## 2021-09-07

- Fix file not saving where expected (better example in notebook)
- Fix "current metacluster" dropdrop delays reflecting newly added metacluster
- Start debugging "rename current metacluster" glitchiness

y2kbugger:
## 2021-09-08

- Finish debugging "rename current metacluster" glitchiness
- Start adding heatmap legend.
- white_check_mark
- eyes
- raised_hands

## 2021-09-09

- Finish adding heatmap legend
- Explore adding hover tooltips

## 2021-09-13

- 50% of input validation done

y2kbugger:
Only thing left is to finish validation unless anyone comes back with more feedback, let me know.

ark/utils/metacluster_remap_gui/.gitignore
```diff
@@ -0,0 +1,3 @@
+Pipfile
```
ngreenwald, line 1: Do we need a separate .gitignore here?

.gitignore
```diff
@@ -21,3 +22,9 @@ env
 */_build
 */_markdown
 */visualizations
+
+data/example_dataset/metaclustering/*output*.csv
+data/example_dataset/metaclustering/example_manually_adjust_metaclusters.ipynb
```
ngreenwald, line 27: This notebook should get moved to /templates and added to git

ark/utils/metacluster_remap_gui/colormap_helper.py
```diff
@@ -0,0 +1,29 @@
...
+
+import matplotlib
+
+
+def distinct_cmap(n=33):
+    rgbs = distinct_rgbs(n)
+    return matplotlib.colors.ListedColormap(rgbs)
+
+
+def distinct_rgbs(n=33):
```
ngreenwald, line 12: The other helper functions don't need a docstring since they're so short, but I think it would be useful to have one for `distinct_rgbs`

ark/utils/metacluster_remap_gui/test_colormap_helper.py
```diff
@@ -0,0 +1,9 @@
+from .colormap_helper import distinct_cmap, distinct_rgbs
+
+
+def test_colormap_is_distinct():
```
ngreenwald, line 4: for test functions, please name them original_function_name_test

ark/utils/metacluster_remap_gui/test_colormap_helper.py
```diff
@@ -0,0 +1,9 @@
+from .colormap_helper import distinct_cmap, distinct_rgbs
```
ngreenwald, line 1: for test files, please name them original_file_name_test.py

ark/utils/metacluster_remap_gui/metaclustergui.py
```diff
@@ -0,0 +1,463 @@
+import warnings
+
+import numpy as np
+import ipywidgets as widgets
+import matplotlib.pyplot as plt
+import matplotlib.patheffects as path_effects
```
ngreenwald, line 6: Please alphabetize these

ark/utils/metacluster_remap_gui/metaclustergui.py
```diff
@@ -0,0 +1,463 @@
...
+from .throttle import throttle
+
+# Third party ipympl causing this in it's backend_agg startup code
+warnings.filterwarnings("ignore", message="nbagg.transparent is deprecated")
+
+DEBUG_VIEW = widgets.Output(layout={'border': '1px solid black'})
+
+
+class MetaClusterGui():
+    def __init__(self, metaclusterdata, heatmapcolors=sns.diverging_palette(240, 10, n=3, as_cmap=True), width=17.0, debug=False, enable_throttle=True):  # noqa
```
ngreenwald, line 25: This can be indented instead to avoid the linter errors right?

ark/utils/metacluster_remap_gui/metaclustergui.py
```diff
@@ -0,0 +1,463 @@
...
+
+        # Tighten layout based on display
+        self.fig.tight_layout()
+        plt.subplots_adjust(hspace=.0)  # make color labels touch heatmap
+        plt.subplots_adjust(wspace=.02)
+
+        # initilize data, etc
+        self.update_gui()
+
+    def make_widgets(self):
```
ngreenwald, line 180: Please add a brief docstring

ark/utils/metacluster_remap_gui/metaclustergui.py
```diff
@@ -0,0 +1,463 @@
...
+        display(DEBUG_VIEW)
+
+    def remap_current_selection(self, metacluster):
+        for cluster in self.selected_clusters:
+            print('remapping', cluster, metacluster)
+            self.mcd.remap(cluster, metacluster)
+        self._heatmaps_stale = True
+        self.mcd.save_output_mapping()
+
+    @DEBUG_VIEW.capture(clear_output=False)
```
ngreenwald, line 366: Do you think the `DEBUG_VIEWs` here are useful to keep in case we need to make changes and debug? Or should they be removed

ark/utils/metacluster_remap_gui/metaclusterdata.py
```diff
@@ -0,0 +1,128 @@
...
+from sklearn.metrics.pairwise import cosine_similarity
+
+
+def metaclusterdata_from_files(cluster_filepath, pixelcount_filepath, metacluster_header='metacluster'):  # noqa
+    clusters = pd.read_csv(cluster_filepath)
+    assert 'cluster' in clusters.columns, "cluster csv must include column named \"cluster\""
+    assert metacluster_header in clusters.columns, "cluster csv must include column named \"metacluster\", alternately specify the metacluster indexs using keyword `metacluster_index`"  # noqa
+    clusters = clusters.rename(columns={metacluster_header: 'metacluster'})
+    pixelcounts = pd.read_csv(pixelcount_filepath)
+    assert 'cluster' in pixelcounts.columns, "pixelcounts csv must include column named \"cluster\""  # noqa
```
ngreenwald, line 14: `assert` statements should be reserved for tests only, please `Raise` an error instead for input validation. 

We have a couple helper functions for input validation here: https://github.com/angelolab/ark-analysis/blob/31cc477932b28eba05c3329c807ac2237cf0b143/ark/utils/misc_utils.py#L120

For an example of how they're used see here, or just click on the def in github and go to `references` https://github.com/angelolab/ark-analysis/blob/31cc477932b28eba05c3329c807ac2237cf0b143/ark/mibi/tiling_utils.py#L33

ngreenwald (changes requested):
Looking good! I left representative comments in a couple places re: test function names, testing patterns, etc, that apply to the rest of the PR as well.

ark/utils/metacluster_remap_gui/.gitignore
y2kbugger: I added this rule to ignore files that i personally used to create native venvs (in addition to docker venvs). This running allowed IDE autocompletion, linter rules, pytest, codecoverage, etc.

It's likely this gitignore rule is fine to add to the root `.gitignore`, but I was being conservative.

Additionally I know vscode IDE has support for connecting to docker container envs but it requires at least some configuration. Using a local venv just let me get to work quickly. I could,  work out and document method for running pytest locally via VSCode using the the docker venvs.

.gitignore
y2kbugger: Notebook is already in templates, this was just a quick solution for the opposite of `update_notebooks.sh` from a example notebook developer perspective.

I would love to discuss a better way to to this, honesty my method feels silly, but it worked and was simple.

Use case:
As a developer, I want changes to ipynb files to be immediately reflected in `git status` so that I choose to commit or discard those changes as I commit changes to supporting library files.

My simple solution is shown below, however, it requiree the additional gitignores to prevent tracking the intermediate file.

image

/speculation
Is the reason for copying the templates during docker build something like this:
As data analyst, I don't want any of my changes to ipynb files to be tracked by git so that pulling updates to the notebooks doesn't cause merge conflicts and I can manually pull latest upstream ipynb files using `update_notebooks.sh`

y2kbugger: Just thought of a more robust solution...and it's even simpler.

Add an optional arg to "start_docker.sh" e.g. `--develop-template-notebooks`

Which will bind the templates folder directly. I think this would allow for having the "favor upstream" workflow be the default behavior while also allowing for normal edit-add-commit workflow when developing template notebooks.

ark/utils/metacluster_remap_gui/colormap_helper.py
y2kbugger: Agreed  :+1:

ark/utils/metacluster_remap_gui/test_colormap_helper.py
y2kbugger: Fixed, I like this, makes it sort the test file right next to implementation. Cool to see pytest added this to suffix to file collection in 2018.

ark/utils/metacluster_remap_gui/test_colormap_helper.py
y2kbugger: After testing this change, pytest does not collect function unless they have a `test_` prefix.

Additionally I checked and it seems the other tests start with `test_` 
https://github.com/angelolab/ark-analysis/search?p=2&q=test
image

ark/utils/metacluster_remap_gui/metaclustergui.py
y2kbugger: Of course, I always try both ways and make a judgement call on readability. One technique to mitigate masking lint error is disabling only one specific lint rather than noqa catchall, pylint supports this for example, but pycodestyle doesn't seem to.

ark/utils/metacluster_remap_gui/metaclustergui.py
y2kbugger: They are disabled by default unless turned on. I did build the debug view as a first class feature meant to left in. Toggling is documented in the developer notes here:
image

ark/utils/metacluster_remap_gui/.gitignore
ngreenwald: That's okay, I think Adam already has this mostly worked out for his workflow

.gitignore
ngreenwald: Yes, the idea is that the vast majority of people using the repo are not developers, and we don't want them to run into merge conflicts when they pull new changes. Mounting the templates directly would solve the problem, or working on a copy that's somewhere else until the PR is ready to merge in. @vacuousplanet/ @alex-l-kong  any preference?

ark/utils/metacluster_remap_gui/test_colormap_helper.py
ngreenwald: Yup, totally right. It should be test_function_name, not function_name_test

ark/utils/metacluster_remap_gui/metaclustergui.py
ngreenwald: You didn't like the result of formatting it something like this?
``` 
def __init__ (xxxx, 
            yyy, zzz)
```

ark/utils/metacluster_remap_gui/.gitignore
y2kbugger: Ok. I've removed folder specific .gitignore

ark/utils/metacluster_remap_gui/metaclustergui.py
y2kbugger: ### tldr
I can and will do it this way, I see it is used throughout the rest of the codebase, it's not horrible.

### long

1. Bias against for wrapping any line which starts a block because it (slightly) degrades the block structure and negatively affects "skimmability"
2. Favor wrapping with constant indent of 4 to minimize noise (nearby lines touched) when commit edits. Otherwise renaming the in the first line forces you to adjust multiple lines. It also give you more space when the first line is already long.
    ```
    myobject.my_list_real_long_variable_naming = [
        a,
        b,
        c,
    ]
    ```
    vs
    ```
    myobject.my_list_real_long_variable_naming = [
                                                   a,
                                                   b,
                                                   c,
                                                     ]
    ```

source: https://www.youtube.com/watch?v=x-kB2o8sd5c ~35 minutes in.

ngreenwald: Totally, sounds good

ark/utils/metacluster_remap_gui/test_metaclusterdata.py
```diff
@@ -0,0 +1,152 @@
...
+
+from .metaclusterdata import MetaClusterData, metaclusterdata_from_files
+
+THIS_DIR = Path(__file__).parent
+TESTDATA_DIR = THIS_DIR / 'test_data'
+DATA_DIR = THIS_DIR.parent.parent.parent / 'data'
+MC_DATA_DIR = DATA_DIR / "example_dataset" / "metaclustering"
+
+
+def test_can_read_example_input_1():
```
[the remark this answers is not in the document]
y2kbugger: This turned out really well. Used my already in place dfs. Mutated the df in place to create the validation issue. Then used it to form an in memory generated csv. This makes it very readable what condition the test is checking for.

    def test_requires_metacluster_column(simple_clusters_df, simple_pixelcount_df):
        simple_clusters_df.rename(columns={'metacluster': 'wrongname'}, inplace=True)
        with pytest.raises(AssertionError):
            metaclusterdata_from_files(as_csv(simple_clusters_df), as_csv(simple_pixelcount_df))

Still have to switch to ValueError so ignore that part.

ark/utils/metacluster_remap_gui/test_data/ex1_clusters_pixelcount_mismatchedids.csv
```diff
@@ -0,0 +1,100 @@
+cluster,count
```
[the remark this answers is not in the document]
y2kbugger: None now :)

ark/utils/metacluster_remap_gui/metaclusterdata.py
```diff
@@ -0,0 +1,128 @@
...
+
+        sorted_clusters_df = raw_clusters_df.sort_values('cluster')
+        self._clusters = sorted_clusters_df.set_index('cluster').drop(columns='metacluster')
+        self.mapping = sorted_clusters_df[['cluster', 'metacluster']].set_index('cluster')
+        self._metacluster_displaynames_map = {}
+        self._marker_order = list(range(len(self._clusters.columns)))
+
+        assert len(set(self.clusters.index)) == len(list(self.clusters.index)), "Cluster ids must be unique."  # noqa
+        assert set(self.clusters.index) == set(self.cluster_pixelcounts.index), "Cluster ids in both files must match"  # noqa
+        assert 1 in self.clusters.index and 0 not in self.clusters.index, "Cluster ids must be integer, starting with 1."  # noqa
```
alex-l-kong, line 31: Change to `raise ValueError(...)` instead of `assert` statements.

ark/utils/metacluster_remap_gui/metaclusterdata.py
```diff
@@ -0,0 +1,128 @@
...
+        self._metacluster_displaynames_map[metacluster] = displayname
+        self.save_output_mapping()
+
+    def save_output_mapping(self):
+        out_df = self.mapping.copy()
+        out_df['mc_name'] = [self.get_metacluster_displayname(mc) for mc in out_df['metacluster']]
+        out_df.to_csv(self.output_mapping_filename)
+
+    def set_marker_order(self, new_indexes):
+        assert set(new_indexes) == set(self._marker_order), \
```
alex-l-kong, line 105: Same as above. We actually have a function that does this already: please take a look at `utils.misc_utils.verify_same_elements`.

ark/utils/metacluster_remap_gui/metaclustergui.py
```diff
@@ -0,0 +1,463 @@
...
+        self.ax_cp.xaxis.set_tick_params(which='both', bottom=False, labelbottom=False)
+        self.ax_cp.yaxis.set_tick_params(which='both', left=False, labelleft=False)
+        self.ax_cp.set_ylabel("Pixels (k)", rotation=90)
+        self.ax_cp.set_xlim(0, self.mcd.cluster_count)
+        self.rects_cp = self.ax_cp.bar(
+            np.arange(self.mcd.cluster_count)+0.5,
+            np.zeros(self.mcd.cluster_count))
+        self.labels_cp = []
+        label_alignment_fudge = 0.08
+        for x in np.arange(self.mcd.cluster_count)+0.5+label_alignment_fudge:
```
alex-l-kong, line 127:
```suggestion
        for x in np.arange(self.mcd.cluster_count) + 0.5 + label_alignment_fudge:
```

alex-l-kong: I believe our linter will complain that there should be a space in between each value connected by a mathematical operation.

ark/utils/metacluster_remap_gui/metaclustergui.py
```diff
@@ -0,0 +1,463 @@
...
+
+    @property
+    def cmap(self):
+        # will never have more metaclusters than clusters
+        return distinct_cmap(self.mcd.cluster_count)
+
+    def preplot(self, df):
+        return df.apply(zscore).clip(upper=self.max_zscore).T
+
+    def make_gui(self):
```
alex-l-kong, line 52: This function is 126 lines: is there a reason we can't break this up into helper functions?

ark/utils/metacluster_remap_gui/metaclustergui.py
```diff
@@ -0,0 +1,463 @@
...
+        # heatmaps
+        self.normalizer = ZScoreNormalize(-1, 0, 1)
+        self.im_c = self.ax_c.imshow(np.zeros((self.mcd.marker_count, self.mcd.cluster_count)), norm=self.normalizer, cmap=self.heatmapcolors, aspect='auto', picker=True)  # noqa
+        self.im_m = self.ax_m.imshow(np.zeros((self.mcd.marker_count, self.mcd.metacluster_count)), norm=self.normalizer, cmap=self.heatmapcolors, aspect='auto', picker=True)  # noqa
+
+        self.ax_c.yaxis.set_tick_params(which='major', labelleft=False)
+        self.ax_c.set_yticks(np.arange(self.mcd.marker_count)+0.5)
+        self.ax_c.set_yticklabels(self.mcd.marker_names)
+        self.ax_c.set_xticks(np.arange(self.mcd.cluster_count)+0.5)
+        self.ax_m.set_xticks(np.arange(self.mcd.metacluster_count)+0.5)
```
alex-l-kong, line 91: See comment about mathematical operation formatting

ark/utils/metacluster_remap_gui/metaclustergui.py
```diff
@@ -0,0 +1,463 @@
...
+
+        self.fig.canvas.mpl_connect('pick_event', self.onpick)
+
+        # heatmaps
+        self.normalizer = ZScoreNormalize(-1, 0, 1)
+        self.im_c = self.ax_c.imshow(np.zeros((self.mcd.marker_count, self.mcd.cluster_count)), norm=self.normalizer, cmap=self.heatmapcolors, aspect='auto', picker=True)  # noqa
+        self.im_m = self.ax_m.imshow(np.zeros((self.mcd.marker_count, self.mcd.metacluster_count)), norm=self.normalizer, cmap=self.heatmapcolors, aspect='auto', picker=True)  # noqa
+
+        self.ax_c.yaxis.set_tick_params(which='major', labelleft=False)
+        self.ax_c.set_yticks(np.arange(self.mcd.marker_count)+0.5)
```
alex-l-kong, line 88: Same comment about mathematical formatting.

ark/utils/metacluster_remap_gui/metaclustergui.py
```diff
@@ -0,0 +1,463 @@
...
+                    path_effects.Stroke(linewidth=4, foreground='white'),
+                    path_effects.Normal(),
+                    ])
+                lb.set_family('monospace')
+                lb.set_zorder(4)
+
+        add_room_for_labels()
+        stretch_dendro_leaves()
+        move_ax_labels()
+        restyle_ax_labels()
```
alex-l-kong, line 294: Any reason we need to do it this way, and we can't do something instead like:

```
# add room for axes labels
ax.set_axisbelow(False)
xlim = ax.get_xlim()
ax.set_xlim((xlim[0], -(xlim[0]*dendrosplit_ratio)))

# stretch dendro leaves
# code in stretch_dentro_leaves

...
```

ark/utils/metacluster_remap_gui/metaclustergui.py
```diff
@@ -0,0 +1,463 @@
...
+        self.im_c.set_clim(self.normalizer.vmin, self.normalizer.vmax)
+
+        # metaclusters heatmap
+        self.im_m.set_data(self.preplot(self.mcd.metaclusters))
+        self.im_m.set_extent((0, self.mcd.metacluster_count, 0, self.mcd.marker_count))
+        self.im_m.set_clim(self.normalizer.vmin, self.normalizer.vmax)
+
+        # xaxis metacluster color labels
+        assert max(self.mcd.metaclusters.index) <= self.mcd.cluster_count, \
+            "Can't support metaclusters idx > cluster count"
```
alex-l-kong, line 328: `raise ValueError(...)` here instead.

alex-l-kong (commented):
Just a few comments about formatting and code aesthetics.

ark/utils/metacluster_remap_gui/metaclustergui.py
y2kbugger: Could be done either way, I look into it, might have started out longer than it ended up.

Two clean code heuristics explain why I did it this way.

1. Single level of abstraction
2. Prefer self-documenting function names over comments

## Single level of abstraction
http://principles-wiki.net/principles:single_level_of_abstraction
All statements of a method should belong to the same level of abstraction. If there is a statement which belongs to a lower level of abstraction, it should go to a private method which comprises statements on this level. Doing so will result in smaller methods.

## Prefer self-documenting function names over comments
https://levelup.gitconnected.com/instead-of-a-comment-use-a-well-named-function-448eabb45607

ark/utils/metacluster_remap_gui/metaclustergui.py
y2kbugger: No i'll break it up. It could also be better factored in terms of "single level of abstraction' principal.

Right now it is a mixture of high and low level. (make_widget, update_gui, vs everything else)
image

ark/utils/metacluster_remap_gui/metaclusterdata.py
y2kbugger: So (unlike the others, which i've got on the backlog to fix) this one really is an assert, not input validation. Meant to catch bugs at develop time when someone changes code. Could honestly be removed at this point if your style guide is to have no asserts. let me know.

ark/utils/metacluster_remap_gui/metaclustergui.py
y2kbugger: Not sure why, but pycodestyle doesn't find this case. I'll recheck manually for any others.

ark/utils/metacluster_remap_gui/metaclustergui.py
y2kbugger: Ok after sleeping on it I realize you probably asked because I used nested defs. It's to decrease the scope of the def itself (e.g. private methods). Basically its a way of getting benefits of a static helper object with private methods but without overcomplicating architecture.

In languages without nested defs I would have created a separate class with the nested def put in as private methods.

ark/utils/metacluster_remap_gui/metaclustergui.py
```diff
@@ -0,0 +1,463 @@
...
+    def cmap(self):
+        # will never have more metaclusters than clusters
+        return distinct_cmap(self.mcd.cluster_count)
+
+    def preplot(self, df):
+        return df.apply(zscore).clip(upper=self.max_zscore).T
+
+    def make_gui(self):
+        #  |  |    Cluster     | Meta |
+        #  ----------------------------
```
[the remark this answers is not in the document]
y2kbugger: Added better explanation of the Axes map comment.

ark/utils/metacluster_remap_gui/metaclustergui.py
y2kbugger: I move highlevel coordination to `__init__` and now `make_widgets` and `make_gui`  are now analogous and orthogonal.

y2kbugger:
Remaining code review items:

- assert -> ValueErrors
- refactor/break out/clarify
    - make_gui
    - move_dendro_labels

y2kbugger:
@ngreenwald I have address all previous code review comments from both reviewer. Ready for re-review.

ark/utils/metacluster_remap_gui/metaclusterdata_test.py
```diff
@@ -0,0 +1,75 @@
...
+        (0.1, 0.2, 0.1),
+        (0.1, 0.1, 0.3),
+        ((0.5 * 50 + 0.7 * 77) / (50 + 77),
+         (0.1 * 50 + 0.2 * 77) / (50 + 77),
+         (0.1 * 50 + 0.1 * 77) / (50 + 77)),
+        ])
+    np.testing.assert_equal(simple_metaclusterdata.metaclusters.values, clusters_data)
+
+
+def test_can_reorder_markers(simple_metaclusterdata: MetaClusterData):
```
ngreenwald, line 62: These test names as well

PR #434, opened 2021-07-26
Add option to change scale factor for cell segmentation

### LINKED ISSUE #423: Add rescale information to notebook
The current segmentation notebook hides the rescale factor. We should pull this arg out with a description of when to change it

alex-l-kong:

Closes #423. `Segment_Image_Data.ipynb` does not make it obvious that the image scaling factor can be adjusted. We add an option to allow the user to easily override the default value of 1.0.

Add a cell before the call to `create_deepcell_output` to allow the user to set the `scale` argument.

### DIFF
--- a/templates/Segment_Image_Data.ipynb
+++ b/templates/Segment_Image_Data.ipynb
@@ -213,7 +213,7 @@
    },
    "outputs": [],
    "source": [
-    "# override default value to increase size of image by a factor of the value of rescale_factor\n",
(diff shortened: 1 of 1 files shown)


templates/Segment_Image_Data.ipynb
```diff
@@ -203,6 +203,20 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": [
+     "seg_scale_set"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "# override default value to increase size of image by a factor of the value of rescale_factor\n",
```
ngreenwald, line 216:
```suggestion
    "# Mesmer was trained on data acquired at 20X resolution. If your image data was acquired at a different resolution, you will get the best performance by rescaling. The rescale factor will increase or decrease the image resolution by the value you provide. For example, if you data was acquired at 10X, use a `rescale_factor` of 2. If your data was acquired at 60X resolution, use a `rescale_factor` of 0.33.\n",
```

ngreenwald (changes requested):
Looks good, added some additional clarification

ngreenwald (approved):

PR #435, opened 2021-07-27
Automatically rebuild Docker if requirements.txt modified

### LINKED ISSUE #391: Automatically rebuild the docker after changes to requirements.txt
Currently when the requirements.txt file changes, users need to rebuild the docker

It would be useful if the update bash script checked to see if the requirements.txt file had changed, and if so, automatically rebuilt the docker with a new image

alex-l-kong:

Closes #391. If the `requirements.txt` file is changed, Docker will need to be rebuilt. However, many users won't know to do this. To prevent confusion, we should automatically rebuild Docker for them if `requirements.txt` has been changed.

We update `start_docker.sh` to run the `docker build` command automatically if the file has been changed within the last day. Tough luck if the user misses the deadline, but I think it's a generous one (the user is likely going to immediately start Docker after a new pull anyways).

### DIFF
--- a/start_docker.sh
+++ b/start_docker.sh
@@ -3,6 +3,7 @@
 # if requirements.txt has been changed in the last day, automatically rebuild Docker first
 if [[ $(find . -mmin -1440 -type f -print | grep requirements.txt | wc -l) -eq 1 ]]
   then
+    echo "New requirements.txt file detected, rebuilding Docker"
     docker build -t ark-analysis .
 fi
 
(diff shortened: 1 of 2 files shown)


start_docker.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+# if requirements.txt has been changed in the last day, automatically rebuild Docker first
+if [[ $(find . -mmin -1440 -type f -print | grep requirements.txt | wc -l) -eq 1 ]]
+  then
```
ngreenwald, line 5: Looks good. Just add a message saying something to the effect that a new requirements file was detected, rebuilding docker image so user knows what's going on.

ngreenwald (changes requested):

ngreenwald (approved):

PR #437, opened 2021-07-28
Spatial lda

bcollica:

Creates the directories and some files for the spatial-LDA pipeline and gives an initial example of code for preprocessing steps.

Added a function for formatting a cell table to be compatible with the existing spatial_lda library.  Also added a function for checking valid arguments along with a test.

**Remaining issues**

Fixing code format to be pep8 compliant so builds can pass.


ark/spLDA/processing.py
```diff
@@ -0,0 +1,86 @@
...
+
+def format_cell_table(cell_table, markers=None, clusters=None, fovs="all_fovs"):
+    """
+    Formats a cell table containing one for more fields of view to be compatible with the spatial_lda library.
+    Args:
+        cell_table (pd.DataFrame):
+            A pandas DataFrame containing the columns of cell marker frequencies and/or cluster ids.
+        markers (list, str):
+            A list of strings corresponding to marker names.
+            Default: None
```
ngreenwald, line 20: No need to list defaults in the docstring

ark/spLDA/processing.py
```diff
@@ -0,0 +1,86 @@
...
+        clusters (list, int):
```
ngreenwald, line 21: Should specify in clusters and markers docstring something to the effect of "either clusters or markers must be provided"

ark/spLDA/processing.py
```diff
@@ -0,0 +1,86 @@
...
+        "label",
+        "cell_size",
+        "centroid-0",
+        "centroid-1",
+        "pixelfreq_hclust_cap",
+        "name"
+    ]
+    if markers is not None:
+        keep_columns.append(markers)
+    drop_columns = [c for c in all_columns if c not in keep_columns]
```
ngreenwald, line 52: Do we use all_columns again? If not, can just use cell_table.columns here

ark/spLDA/processing.py
```diff
@@ -0,0 +1,86 @@
...
+    Args:
+        cell_table (pd.DataFrame):
+            A pandas DataFrame containing the columns of cell marker frequencies and/or cluster ids.
+        markers (list, str):
+            A list of strings corresponding to marker names.
+            Default: None
+        clusters (list, int):
+            A list of integers corresponding to cluster ids.
+            Default: None
+        fovs (str, int):
```
ngreenwald, line 24: Do you think the subsetting should happen here, or later in the pipeline? My gut would be later makes more sense, but I could easily be persuaded

ark/spLDA/processing.py
```diff
@@ -0,0 +1,86 @@
...
+    # Create dictionary of FOVs
+    if fovs == "all_fovs":
+        fovs = [x + 1 for x in range(max(cell_table["point"]))]
+
+    fov_dict = {}
+    for i in fovs:
+        df = cell_table[cell_table["point"] == i].drop(columns=["point", "label"])
+        if clusters is not None:
+            df = df[df["cluster_id"].isin(clusters)]
+        df['is_index'] = True
```
ngreenwald, line 74: Do we need this and the following line? This was for compatibility with the TNBC notebook from what I remember?

ark/utils/spatial_lda_utils.py
```diff
@@ -0,0 +1,63 @@
...
+        "label",
+        "cell_size",
+        "centroid-0",
+        "centroid-1",
+        "pixelfreq_hclust_cap",
+        "name"
+    ]
+
+    # Check cell table
+    if not isinstance(cell_table, pd.DataFrame):
```
ngreenwald, line 36: One important part of the testing philosophy is figuring out which functions can trust each other, and which cannot. Because we'll be loading the .csv ourselves in a previous function, I think we can assume that if the user gets this far it was succesfully converted to a pandas file

ark/utils/spatial_lda_utils.py
```diff
@@ -0,0 +1,63 @@
...
+        raise ValueError("cell_table must be a pd.DataFrame")
+
+    if not all([True for x in base_cols if x in cell_table.columns]):
```
ngreenwald, line 39: This one is definitely necessary, because we have no idea where they got the CSV from. Perhaps a previous version of the pipeline that didn't have the right columns, etc

ark/utils/spatial_lda_utils.py
```diff
@@ -0,0 +1,63 @@
...
+            A list of integers corresponding to cluster ids.
+        fovs (str, int):
+            One of either "all_fovs" indicating all field of views are to be kept, or a list of integers corresponding
+            to the index of each field of view which should be kept.
+
+    Returns:
+        None
+    """
+
+    base_cols = [
```
ngreenwald, line 25: Since this is reused in this function and the preceeding one, we can change it to a global variable and place it here https://github.com/angelolab/ark-analysis/blob/master/ark/settings.py

ark/spLDA/processing.py
```diff
@@ -0,0 +1,86 @@
...
+            "cell_size": "area",
+            "centroid-0": "x",
+            "centroid-1": "y",
+            "pixelfreq_hclust_cap": "cluster_id",
+            "name": "cluster"
+        })
+
+    # Create dictionary of FOVs
+    if fovs == "all_fovs":
+        fovs = [x + 1 for x in range(max(cell_table["point"]))]
```
ngreenwald, line 67: Can't assume that they're sequential

ark/utils/spatial_lda_utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # Check markers/clusters
+    if all([markers is None, clusters is None]):
+        raise ValueError("markers and clusters cannot both be None")
+    if markers is not None:
+        if isinstance(markers, list) and len(markers) == 0:
+            raise ValueError("list of marker names cannot be empty")
+        if not isinstance(markers, list) or not all([isinstance(x, str) for x in markers]):
+            raise ValueError("clusters must be a list of integers")
+        assert all([x in cell_table.columns for x in markers])
```
ngreenwald, line 50: Asserts should be reserved for test functions. I don't know why they decided on this for python, but that's the way it's done.

ark/utils/spatial_lda_utils_test.py
```diff
@@ -0,0 +1,79 @@
+import os
+import numpy as np
+import pandas as pd
+import tempfile
+import pytest
+from ark.utils.spatial_lda_utils import *
```
ngreenwald, line 6: No wildcard imports unless absolutely necessary

ark/utils/spatial_lda_utils_test.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # FOVs
+    VALID_FOVS = [1, 2, 3]
+    # Strings instead of integers
+    INVALID_FOVS1 = ["a", "b", "c"]
+    # Empty List
+    INVALID_FOVS2 = []
+
+    # Run Checks
+    with pytest.raises(ValueError):
```
ngreenwald, line 54: I believe these tests will pass as long as at least one of these raises a ValueError; it won't require that they each do. You should double check this to make sure I'm not wrong though

ark/utils/spatial_lda_utils_test.py
```diff
@@ -0,0 +1,79 @@
...
+        # DataFrame Checks
+        check_format_cell_table_args(INVALID_DF1, VALID_MARKERS, VALID_CLUSTERS, VALID_FOVS)
```
ngreenwald, line 56: For input validation, it's important to make sure the right ValueError is being raised. You can do that like this
https://github.com/angelolab/ark-analysis/blob/16466141fbf6ab8bffb6b4e613010bbb763f4f64/ark/utils/io_utils_test.py#L44

ngreenwald (changes requested):
Awesome start! Let me know if any of the comments are confusing. And obviously if you disagree with any of them please let me know, don't feel the need to appease me.

ark/spLDA/processing.py
bcollica: Either way works for me.  I figured if people knew they only wanted to run some FOVs then it just cuts down on the amount of runtime in the other steps.

ark/spLDA/processing.py
bcollica: I believe it's because the featurization step is in the TNBC notebook.  This ends up being the argument of is_anchor_col in featurize_samples() in spatial_lda.featurization.py.  So for most people they will want to use all cells which is why it's hard coded here.  But if they do want to use some other condition contained in a different column then they could pass that into the featurize_cell_table function, but I will have to add support for that in format_cell_table().

ark/utils/spatial_lda_utils.py
ackagel: I have a nitpick

`if not all([x in cell_table.columns for x in base_cols]):`

ark/utils/spatial_lda_utils.py
bcollica: Yes, and we can change it as needed if the column names adjust

ark/utils/spatial_lda_utils.py
bcollica: Sounds good

ark/spLDA/processing.py
ngreenwald: That's true. However, I think people will likely try training with different settings. So it'll be easiest if we preprocess everything first, since they'll eventually need this table anyway to run the model

ark/spLDA/processing.py
ngreenwald: Got it. So do they also hardcode this for the other example notebook that isn't the TNBC dataset? Or they use a different preprocessing function

ark/utils/spatial_lda_utils.py
ngreenwald: `np.all(np.isin(list1, list2))`

ark/spLDA/processing.py
bcollica: It shouldn't matter because the cell table just gets filtered in the following for loop anyway so the order shouldn't matter.  But we might run into trouble if numbers get skipped (like 1,2,6,7,8...), so I can replace this with np.unique().

ark/spLDA/processing.py
bcollica: Sounds good

ark/spLDA/processing.py
bcollica: The spatial_lda package has essentially three separate featurization functions: featurize_spleens(), featurize_tumors(), and featurize_samples().  featurize_samples() is a generalization of the other two in the sense that it takes the is_anchor_col argument which is a column name that specifies which cells to use when constructing a neighborhood.  So featurize_spleens() and featurize_tumors() are just wrappers for featurize_samples() which pass hard coded values for is_anchor_col.

ngreenwald: Got it, in that case sounds like what you have here is the best solution

ark/spLDA/processing.py
ngreenwald: Yeah people will almost certainly drop points for various reasons.

ark/spLDA/processing.py
```diff
@@ -0,0 +1,86 @@
...
+from ark.utils.spatial_lda_utils import check_format_cell_table_args
+# from spatial_lda.featurization import *
+# from spatial_lda.visualization import plot_adjacency_graph()
+
+
+def format_cell_table(cell_table, markers=None, clusters=None, fovs="all_fovs"):
+    """
+    Formats a cell table containing one for more fields of view to be compatible with the spatial_lda library.
+    Args:
+        cell_table (pd.DataFrame):
```
alex-l-kong, line 16: When listing datatypes in the docstring, you'll need to use the full class name so it can get mocked up properly. Here, it should be `pandas.DataFrame`.

ark/spLDA/processing.py
```diff
@@ -0,0 +1,86 @@
...
+        columns={
+            "cell_size": "area",
+            "centroid-0": "x",
+            "centroid-1": "y",
+            "pixelfreq_hclust_cap": "cluster_id",
+            "name": "cluster"
+        })
+
+    # Create dictionary of FOVs
+    if fovs == "all_fovs":
```
alex-l-kong, line 66: I think to follow `ark-analysis` protocol you should set a default of `fovs` being None rather than `all_fovs`. If `fovs` is None it will retrieve all the fovs. An added benefit is you don't have to specify the `fovs` type as being `list, str` in the docstring, just `list`.

ark/spLDA/processing.py
```diff
@@ -0,0 +1,86 @@
...
+    all_columns = cell_table.columns
+    keep_columns = [
+        "point",
+        "label",
+        "cell_size",
+        "centroid-0",
+        "centroid-1",
+        "pixelfreq_hclust_cap",
+        "name"
+    ]
```
alex-l-kong, line 49: Consider putting this in `ark.settings`.

ark/spLDA/processing.py
alex-l-kong: Can `clusters` be a single `int` or is it just a list of `int`. If the latter, you do not need to specify `list, int`, just `list`. The user can figure out from the documentation that the `list` should always contain `int`.

ark/utils/spatial_lda_utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # Check cell table
+    if not isinstance(cell_table, pd.DataFrame):
+        raise ValueError("cell_table must be a pd.DataFrame")
+
+    if not all([True for x in base_cols if x in cell_table.columns]):
+        raise ValueError("cell table must contain the following columns:{}".format(base_cols))
+
+    # Check markers/clusters
+    if all([markers is None, clusters is None]):
```
alex-l-kong, line 43: `if markers is None and clusters is None` is easier to understand.

ark/utils/spatial_lda_utils.py
```diff
@@ -0,0 +1,63 @@
...
+        raise ValueError("markers and clusters cannot both be None")
+    if markers is not None:
+        if isinstance(markers, list) and len(markers) == 0:
```
alex-l-kong, line 46: I think we can leave out the `isinstance(markers, list)` portion and safely assume that `markers` is a `list` type.

ark/utils/spatial_lda_utils.py
```diff
@@ -0,0 +1,63 @@
...
+            raise ValueError("list of marker names cannot be empty")
+        if not isinstance(markers, list) or not all([isinstance(x, str) for x in markers]):
```
alex-l-kong, line 48: Same here.

ark/utils/spatial_lda_utils.py
```diff
@@ -0,0 +1,63 @@
...
+            raise ValueError("clusters must be a list of integers")
+        assert all([x in cell_table.columns for x in markers])
+    if clusters is not None:
+        if isinstance(clusters, list) and len(clusters) == 0:
```
alex-l-kong, line 52: And here.

ark/utils/spatial_lda_utils.py
```diff
@@ -0,0 +1,63 @@
...
+            raise ValueError("list of cluster ids cannot be empty")
+        if not isinstance(clusters, list) or not all([isinstance(x, int) for x in clusters]):
```
alex-l-kong, line 54: And here.

ark/utils/spatial_lda_utils.py
```diff
@@ -0,0 +1,63 @@
...
+            raise ValueError("markers must be a list of strings")
+
+    # Check fovs
+    if fovs != "all_fovs":
+        if not all([isinstance(x, int) for x in fovs]) or not isinstance(fovs, list) or len(fovs) == 0:
```
alex-l-kong, line 59: And here.

ark/utils/spatial_lda_utils.py
```diff
@@ -0,0 +1,63 @@
...
+            raise ValueError("fovs must be 'all_fovs' or a list of integers")
+
+    return None
```
alex-l-kong, line 62: There's no need to `return None` in Python. Simply leave the function without a return value and adjust the docstring to not have a `Returns:` section.

ark/spLDA/processing.py
```diff
@@ -0,0 +1,86 @@
+import functools
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import os
+import pickle
+from ark.utils.spatial_lda_utils import check_format_cell_table_args
+# from spatial_lda.featurization import *
+# from spatial_lda.visualization import plot_adjacency_graph()
```
alex-l-kong, line 9: You'll probably get to this later, but just a reminder not to have any unnecessary commented out code.

ark/spLDA/processing.py
```diff
@@ -0,0 +1,86 @@
...
+
+    return(fov_dict)
+
+
+
+
+
+
+
+
```
alex-l-kong, line 86: There should only be one blank line at the end of each file.

ark/utils/spatial_lda_utils_test.py
```diff
@@ -0,0 +1,79 @@
...
+        check_format_cell_table_args(VALID_DF, VALID_MARKERS, VALID_CLUSTERS, INVALID_FOVS2)
+
+
+
+
+
+
+
+
+
```
alex-l-kong, line 79: Same comment about 1 blank line applies here.

ark/utils/spatial_lda_utils_test.py
```diff
@@ -0,0 +1,79 @@
...
+    INVALID_CLUSTERS1 = ["a", "b", "c"]
+    # Empty List
+    INVALID_CLUSTERS2 = []
+
+    # FOVs
+    VALID_FOVS = [1, 2, 3]
+    # Strings instead of integers
+    INVALID_FOVS1 = ["a", "b", "c"]
+    # Empty List
+    INVALID_FOVS2 = []
```
alex-l-kong, line 51: Since these are not global constants, you should not make these variables all caps.

alex-l-kong (changes requested):
Good stuff, just added a few comments of my own.

ark/utils/spatial_lda_utils.py
bcollica: This was taken care of in #439

ark/spLDA/processing.py
bcollica: This was taken care of in #439

ackagel:
@ngreenwald just for reference, the exploratory topic number estimation in the example notebook took >>20mins (possibly more, I had to stop it to run something else), with only 2 fovs / 3k cells.  Most of that computation was just KMeans.  I changed that to MiniBatchKMeans, which isn't as precise, but much faster (~3mins).  Topic estimating statistics (elbow, silhouette, gap, etc) looked pretty much the same, just a smidge noisier.  I could make the batch size (currently 1024 cells) a keyword arg if the stats don't look clean enough (they look fine imo).

Order of ops for getting LDA in:
 - merge this branch w/ my minor updates into the inference/training branch
 - make sure that's working ok*
 - merge that into here
 - merge+resolve changes from master in here
 - merge into master

* If it's orders of magnitude too slow, or in desperate need of some additional visualizations, I can guide (actively or passively) Sri or Cami through fixing that issue, or just tackle it myself if it's something small (clearly <1 day's worth of time). It'd be great to run some a subset of your data through that to stress test it a bit (maybe on mibian or something?).

ngreenwald:
Hmmm that's strange, the reason Erin liked Kmeans was because it was so fast. Maybe there's something not quite right about the python implementation, I'm pretty sure she used R. 

Can you also check and see how long it's taking you to run the actual spatialLDA?

ngreenwald:
Agreed regarding the rest of the todos

ackagel:
Yeah, just looking at the backend implementation, sklearn's KMeans calls an updating cython function (very fast, very cool), but iteratively calls it in a regular ol, super slow python for loop (maybe for a good reason but it seems silly to me).  I imagine the R implementation Erin used had the foresight to put the iteration on the C/C++ side.

Regardless, the MiniBatchKMeans is >10x faster and should scale well with actual datasets.  Again, it's not as accurate as KMeans, but we're only using it to approximate the topic number anyways and it seems to work fine for that.

ackagel:
@ngreenwald not an actual review request; just letting you know those minimal tests are in and this is ready for the merge-arino

ngreenwald:
Great, meaning there's no other low-hanging fruit for this to be ready for lab use?

One thing I noticed: some of the diagnostic plots are empty
image

Were these taking a long time and hence are excluded?

ackagel:
Exactly.  I'm pretty sure all of those metrics can still be toggled on/off (off by default), so I left all of the plots there, since they don't crash/halt execution.  I could put a note about it in the notebook or something, get rid of them, or leave them blank, up to you/

ngreenwald:
Okay cool, I think it's fine to leave as is. 

This heatmap should be column normalized though:
image
Right now the most populous cell types are dominating, but we want to see the contribution within each cluster from each cell type

ackagel:
@ngreenwald do you want column normalization or row norm? i.e normalize by number of cells in cluster (col) or number of cells in cell_type (row)?

ngreenwald:
Column normalized

ackagel:
@ngreenwald columns should be normalized now, feel free to check that

ngreenwald:
I think something's not quite right. If the columns were normalized we'd expect to see same total signal intensity across the different columns. However, some columns are almost completely black, and others have tons of signal in different cell types
image

ngreenwald:
Cool, and what about the `spatial-analysis-nb-update` branch? Is that related to this?

ackagel:
nope that's just enrichment.  That one's fully functional (minus context + new distance column hack), but there were requests for a cell-cluster/feature overlay that I was planning on adding when I initially made that PR.

ngreenwald:
Oh, that's spatial_analysis, not spatial_lda. I didn't even look at what was in it, just figured it was part of spatial_LDA. Got it.

ngreenwald (approved):

PR #438, opened 2021-07-29
Improve jupyter launching (Jupyter Lab + enhanced alternate PORT)

y2kbugger:

- Launch **Jupyter** as **Jupyter Lab** instead of **Jupyter Notebook**
- Enhance `start_docker.sh` to pass the first available port to jupyter so that "remapped port" warning is no longer required.
- Enhance available port detection to be more reliable
- Add execute to start_docker.sh to allow running directly, updated readme to match.

I've touched start_docker.sh and Dockerfile.
For details, see commit history as I have broken up commits to tell story clearly.

**Remaining issues**

None


y2kbugger:
@ngreenwald Figured we could merge this into master prior to finalizing the metacluster GUI.

ngreenwald (commented):
Overall looks good! Is there a code cleanliness reason that having the script be executable is preferable to running bash scriptname? Or just a personal preference?

y2kbugger:
> Overall looks good! Is there a code cleanliness reason that having the script be executable is preferable to running bash scriptname? Or just a personal preference?

I'd say it follows "least astonishment" since it allows both `bash ./script.sh` and `./script.sh`and different people might be used to one or another.
Marking as executable also made sense since the script already included a the "hash-bang" (`#!`), which does nothing if not ran directly.

*start_docker.sh*

    #!/bin/bash
    ...

y2kbugger:
I did just push a small change, to use the _slightly_ more portable `#!/usr/bin/env bash` ipo `#!/bin/bash`

https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang

ngreenwald (approved):
Got it, that all makes sense.

PR #439, opened 2021-08-02
Processing branch

bcollica:

This PR merges functions, helper scripts, and tests for the initial data processing steps in spatial-lda analysis.  The PR is intentionally small to allow for easier integration into the spatial-LDA branch. 

Three main functions were added [format_cell_table(), featurize_cell_table(), create_difference_matrices()] which create the necessary data structures to be passed to the existing spatial_lda library.  For each processing function, an argument-checking helper function was added in spatial_lda_utils.py, and tests for the helper functions are included in spatial_lda_utils_test.py.

**Remaining issues**

No tests are currently implemented for the three processing functions specifically.

### DIFF
--- a/ark/settings.py
+++ b/ark/settings.py
@@ -8,6 +8,8 @@ CELL_TYPE = 'cell_type'             # cell type name (flowsom)
 CLUSTER_ID = 'FlowSOM_ID'           # cell cluster id (flowsom)
 PATIENT_ID = 'PatientID'            # cell's patient id
 KMEANS_CLUSTER = 'cluster_labels'   # generated cluster column name
+CENTROID_0 = 'centroid-0'           # cell centroid x-coordinate
+CENTROID_1 = 'centroid-1'           # cell centroid y-coordinate
 
 # standardized columns surrounding channel data
 PRE_CHANNEL_COL = CELL_SIZE  # last column before channel data
@@ -23,5 +25,4 @@ REGIONPROPS_SINGLE_COMP = ['major_minor_axis_ratio', 'perim_square_over_area',
 REGIONPROPS_MULTI_COMP = ['nc_ratio']
 
 # spatial-LDA minimum required columns
-BASE_COLS = ["point", "label", "cell_size", "centroid-0", "centroid-1",
-             "pixelfreq_hclust_cap", "name"]
+BASE_COLS = [FOV_ID, CELL_LABEL, CELL_SIZE, CENTROID_0, CENTROID_1, CLUSTER_ID, KMEANS_CLUSTER]
--- a/ark/settings.py
+++ b/ark/settings.py
@@ -1,17 +1,17 @@
 # hope u liek capital letters
 
 # default segmented csv column names
-CELL_SIZE = 'cell_size'                     # cell size
-CELL_LABEL = 'label'                        # cell label number (regionprops)
-FOV_ID = 'SampleID'                         # cell's fov name
-CELL_TYPE = 'cell_type'                     # cell type name (flowsom)
-CLUSTER_ID = 'FlowSOM_ID'                   # cell cluster id (flowsom)
-PATIENT_ID = 'PatientID'                    # cell's patient id
-KMEANS_CLUSTER = 'cluster_labels'           # generated cluster column name
+CELL_SIZE = 'cell_size'  # cell size
+CELL_LABEL = 'label'  # cell label number (regionprops)
+FOV_ID = 'SampleID'  # cell's fov name
+CELL_TYPE = 'cell_type'  # cell type name (flowsom)
+CLUSTER_ID = 'FlowSOM_ID'  # cell cluster id (flowsom)
+PATIENT_ID = 'PatientID'  # cell's patient id
+KMEANS_CLUSTER = 'cluster_labels'  # generated cluster column name
 
 # standardized columns surrounding channel data
-PRE_CHANNEL_COL = CELL_SIZE                 # last column before channel data
-POST_CHANNEL_COL = CELL_LABEL               # first column after channel data
+PRE_CHANNEL_COL = CELL_SIZE  # last column before channel data
+POST_CHANNEL_COL = CELL_LABEL  # first column after channel data
 
 # regionprops extraction
 REGIONPROPS_BASE = ['label', 'area', 'eccentricity', 'major_axis_length',
@@ -23,4 +23,5 @@ REGIONPROPS_SINGLE_COMP = ['major_minor_axis_ratio', 'perim_square_over_area',
 REGIONPROPS_MULTI_COMP = ['nc_ratio']
 
 # spatial-LDA minimum required columns
-BASE_COLS = ["point", "label", "cell_size", "centroid-0", "centroid-1", "pixelfreq_hclust_cap", "name"]
+BASE_COLS = ["point", "label", "cell_size", "centroid-0", "centroid-1",
+             "pixelfreq_hclust_cap", "name"]
--- a/ark/settings.py
+++ b/ark/settings.py
@@ -1,13 +1,13 @@
 # hope u liek capital letters
 
 # default segmented csv column names
-CELL_SIZE = 'cell_size'  # cell size
-CELL_LABEL = 'label'  # cell label number (regionprops)
-FOV_ID = 'SampleID'  # cell's fov name
-CELL_TYPE = 'cell_type'  # cell type name (flowsom)
-CLUSTER_ID = 'FlowSOM_ID'  # cell cluster id (flowsom)
-PATIENT_ID = 'PatientID'  # cell's patient id
-KMEANS_CLUSTER = 'cluster_labels'  # generated cluster column name
+CELL_SIZE = 'cell_size'             # cell size
+CELL_LABEL = 'label'                # cell label number (regionprops)
+FOV_ID = 'SampleID'                 # cell's fov name
+CELL_TYPE = 'cell_type'             # cell type name (flowsom)
+CLUSTER_ID = 'FlowSOM_ID'           # cell cluster id (flowsom)
+PATIENT_ID = 'PatientID'            # cell's patient id
+KMEANS_CLUSTER = 'cluster_labels'   # generated cluster column name
 
 # standardized columns surrounding channel data
 PRE_CHANNEL_COL = CELL_SIZE  # last column before channel data
--- a/ark/settings.py
+++ b/ark/settings.py
@@ -21,3 +21,6 @@ REGIONPROPS_SINGLE_COMP = ['major_minor_axis_ratio', 'perim_square_over_area',
                            'major_axis_equiv_diam_ratio', 'convex_hull_resid',
                            'centroid_dif', 'num_concavities']
 REGIONPROPS_MULTI_COMP = ['nc_ratio']
+
+# spatial-LDA minimum required columns
+BASE_COLS = ["point", "label", "cell_size", "centroid-0", "centroid-1", "pixelfreq_hclust_cap", "name"]
--- a/ark/spLDA/processing.py
+++ b/ark/spLDA/processing.py
@@ -3,32 +3,38 @@ from sklearn.model_selection import train_test_split
 from spatial_lda.featurization import *
 
 from ark.settings import BASE_COLS
-from ark.utils.spatial_lda_utils import check_format_cell_table_args, check_featurize_cell_table_args
+from ark.utils.spatial_lda_utils import check_format_cell_table_args, \
+    check_featurize_cell_table_args
 
 train_frac = 0.75
 
 
 def format_cell_table(cell_table, markers=None, clusters=None):
     """
-    Formats a cell table containing one for more fields of view to be compatible with the spatial_lda library.
+    Formats a cell table containing one for more fields of view to be
+    compatible with the spatial_lda library.
     Args:
         cell_table (pd.DataFrame):
-            A pandas DataFrame containing the columns of cell marker frequencies and/or cluster ids.
+            A pandas DataFrame containing the columns of cell marker
+            frequencies and/or cluster ids.
         markers (list, str):
-            A list of strings corresponding to marker names.  Either markers or clusters must be provided.
+            A list of strings corresponding to marker names.  Either markers
+            or clusters must be provided.
         clusters (list, int):
             A list of integers corresponding to cluster ids.
 
     Returns:
         dict:
 
-        - A dictionary of formatted cell tables for use in spatial-LDA analysis.  Each element in the dictionary is a
+        - A dictionary of formatted cell tables for use in spatial-LDA
+        analysis.  Each element in the dictionary is a
         pd.DataFrame corresponding to a single field of view.
 
     """
 
     # Check function arguments
-    check_format_cell_table_args(cell_table=cell_table, markers=markers, clusters=clusters)
+    check_format_cell_table_args(cell_table=cell_table, markers=markers,
+                                 clusters=clusters)
 
     # Only keep columns relevant for spatial-LDA
     if markers is not None:
@@ -51,7 +57,8 @@ def format_cell_table(cell_table, markers=None, clusters=None):
 
     fov_dict = {}
     for i in fovs:
-        df = cell_table[cell_table["point"] == i].drop(columns=["point", "label"])
+        df = cell_table[cell_table["point"] == i].drop(
+            columns=["point", "label"])
         if clusters is not None:
             df = df[df["cluster_id"].isin(clusters)]
         df["is_index"] = True
@@ -66,26 +73,34 @@ def format_cell_table(cell_table, markers=None, clusters=None):
     return fov_dict
 
 
-def featurize_cell_table(cell_table, feature_by="cluster", radius=100, cell_index="is_index", n_processes=None,
+def featurize_cell_table(cell_table, feature_by="cluster", radius=100,
+                         cell_index="is_index", n_processes=None,
                          train_split=True):
     """
     Args:
         cell_table (dict, pd.DataFrame):
-            A formatted cell table for use in spatial-LDA analysis.  Specifically, this is the output from
+            A formatted cell table for use in spatial-LDA analysis.
+            Specifically, this is the output from
             format_cell_table().
         feature_by (str):
             One of four choices of featurization method:
-                * marker: for each marker, count the total number of cells within a ``radius`` *r* from cell *i*
+                * marker: for each marker, count the total number of cells
+                within a ``radius`` *r* from cell *i*
                 having marker expression greater than 0.5.
-                * avg_marker: for each marker, compute the average marker expression of all cells with a ``radius`` *r*
+                * avg_marker: for each marker, compute the average marker
+                expression of all cells with a ``radius`` *r*
                 from cell *i*.
-                * cluster: for each cluster, count the total number of cells with a ``radius`` *r* from cell *i*
+                * cluster: for each cluster, count the total number of cells
+                with a ``radius`` *r* from cell *i*
                 belonging to that cell cluster.
-                * count: counts the total number of cells within a ``radius`` *r* from cell *i*.
+                * count: counts the total number of cells within a ``radius``
+                *r* from cell *i*.
         radius (int):
-            Size of the radius, in pixels, used to featurize cellular neighborhoods.
+            Size of the radius, in pixels, used to featurize cellular
+            neighborhoods.
         cell_index (str):
-            Name of the column containing the reference cell indexes.  If not specified, all cells are used.
+            Name of the column containing the reference cell indexes.  If not
+            specified, all cells are used.
         n_processes (int):
             Number of parallel processes to use.
         train_split (bool):
@@ -94,9 +109,12 @@ def featurize_cell_table(cell_table, feature_by="cluster", radius=100, cell_inde
     Returns:
         dict:
 
-        - A dictionary containing featurized cellular neighborhoods and training splits if applicable.
+        - A dictionary containing featurized cellular neighborhoods and
+        training splits if applicable.
     """
-    check_featurize_cell_table_args(cell_table=cell_table, feature_by=feature_by, radius=radius, cell_index=cell_index)
+    check_featurize_cell_table_args(cell_table=cell_table,
+                                    feature_by=feature_by, radius=radius,
+                                    cell_index=cell_index)
     # Define Featurization Function
     func_type = {
         "marker": neighborhood_to_marker,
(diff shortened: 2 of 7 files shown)


bcollica:
Read the docs build failed because I forgot add function descriptions.  Doing that now

bcollica:
the read the docs gods accepted my sacrifice

ark/settings.py
```diff
@@ -1,17 +1,17 @@
...
 
 # default segmented csv column names
-CELL_SIZE = 'cell_size'                     # cell size
-CELL_LABEL = 'label'                        # cell label number (regionprops)
-FOV_ID = 'SampleID'                         # cell's fov name
-CELL_TYPE = 'cell_type'                     # cell type name (flowsom)
-CLUSTER_ID = 'FlowSOM_ID'                   # cell cluster id (flowsom)
-PATIENT_ID = 'PatientID'                    # cell's patient id
-KMEANS_CLUSTER = 'cluster_labels'           # generated cluster column name
+CELL_SIZE = 'cell_size'  # cell size
```
ackagel, line 4: my tabs! 😢

ark/settings.py
```diff
@@ -21,3 +21,7 @@
                            'major_axis_equiv_diam_ratio', 'convex_hull_resid',
                            'centroid_dif', 'num_concavities']
 REGIONPROPS_MULTI_COMP = ['nc_ratio']
+
+# spatial-LDA minimum required columns
+BASE_COLS = ["point", "label", "cell_size", "centroid-0", "centroid-1",
+             "pixelfreq_hclust_cap", "name"]
```
ackagel, line 27: Not suer if we want to use the globals defined in the other settings here, i.e
```
BASE_COLS = [FOV_ID,  CELL_LABEL, ...]
```

ark/spLDA/processing.py
```diff
@@ -1,56 +1,45 @@
...
-# from spatial_lda.featurization import *
-# from spatial_lda.visualization import plot_adjacency_graph()
+from sklearn.model_selection import train_test_split
+from spatial_lda.featurization import *
 
+from ark.settings import BASE_COLS
+from ark.utils.spatial_lda_utils import check_format_cell_table_args, \
+    check_featurize_cell_table_args
+
+train_frac = 0.75
```
ackagel, line 9: spooky global.  Looks like it can be made into a defaulted argument for `featurize_cell_table`.  Either way if there isn't a work around, it should be capitalized.

ark/spLDA/processing.py
```diff
@@ -63,24 +52,153 @@ def format_cell_table(cell_table, markers=None, clusters=None, fovs="all_fovs"):
...
-    return(fov_dict)
+    # Save Arguments
+    fov_dict["fovs"] = fovs
+    fov_dict["markers"] = markers
+    fov_dict["clusters"] = clusters
+
+    return fov_dict
 
 
+def featurize_cell_table(cell_table, feature_by="cluster", radius=100,
```
ackagel, line 75: `feature_by` might be clearer as `featurization`

ark/spLDA/processing.py
```diff
@@ -63,24 +52,153 @@ def format_cell_table(cell_table, markers=None, clusters=None, fovs="all_fovs"):
...
+        "cluster": neighborhood_to_cluster,
+        "avg_marker": neighborhood_to_avg_marker,
+        "count": neighborhood_to_count
+    }
+    if feature_by in ["marker", "avg_marker"]:
+        neighborhood_feature_fn = functools.partial(func_type[feature_by],
+                                                    markers=cell_table[
+                                                        "markers"])
+    else:
+        neighborhood_feature_fn = functools.partial(func_type[feature_by])
```
ackagel, line 131: I understand we're pulling from a dict of functions, but what is this snippet doing?

ark/spLDA/processing.py
```diff
@@ -63,24 +52,153 @@ def format_cell_table(cell_table, markers=None, clusters=None, fovs="all_fovs"):
...
+                                        x_col='x',
+                                        y_col='y',
+                                        n_processes=n_processes,
+                                        include_anchors=True)
+
+    if train_split:
+        all_sample_idxs = featurized_fovs.index.map(lambda x: x[0])
+        _sets = train_test_split(featurized_fovs, test_size=1. - train_frac,
+                                 stratify=all_sample_idxs)
+        train_features_fraction, _ = _sets
```
ackagel, line 149: Any reason we're storing the `_sets` variable?  Could we just have:
```
train_features_fraction, _ = train_test_split(...)
```

ark/spLDA/processing.py
```diff
@@ -63,24 +52,153 @@ def format_cell_table(cell_table, markers=None, clusters=None, fovs="all_fovs"):
...
+    if inference:
+        inference_diff_mat = make_merged_difference_matrices(
+            sample_features=features["featurized_fovs"],
+            sample_dfs=cell_table,
+            x_col="x", y_col="y")
+    else:
+        inference_diff_mat = None
+
+    return {"train_diff_mat": train_diff_mat,
+            "inference_diff_mat": inference_diff_mat}
```
ackagel, line 204: Could be condensed as:
```
arg_settings = {
    "train_diff_mat" : (training, "train_features"),
    "inference_diff_mat": (inference, "featureized_fovs"),
}

return {
    k: None if not v[0] else make_merged_difference_matrices(
        sample_features=features[v[1]],
        sample_dfs=cell_table,
        x_col="x", y_col="y")
    for (k, v) in arg_settings
}
```
but with less nasty tabbing

ackagel (commented):
Wicked nice start.  Only things that looks off is that `train_frac` global; the other comments are mostly nitpicks.

ark/settings.py
bcollica: Oh sorry, I think that was Pycharm's fault!

ark/spLDA/processing.py
bcollica: I pulled this from the Calico Labs example notebook.  I believe it's necessary for certain steps in the featurize_samples() function.

ark/spLDA/processing.py
bcollica: Also pulled from Calico notebook, and yes i can change it to be that way.

ark/settings.py
ngreenwald: Yes, lets try and mirror the column names in the segmentation portion of this pipeline. You can rename the columns from the example data table you got from Alea to be in line with what the rest of the repo is using

ark/spLDA/processing.py
ngreenwald: I think this is a bit too fancy for my tastes. It's about half as long, but takes twice as long to figure out what's going on.

ark/spLDA/processing.py
```diff
@@ -63,24 +52,153 @@ def format_cell_table(cell_table, markers=None, clusters=None, fovs="all_fovs"):
...
+                from cell *i*.
+                * cluster: for each cluster, count the total number of cells
+                with a ``radius`` *r* from cell *i*
+                belonging to that cell cluster.
+                * count: counts the total number of cells within a ``radius``
+                *r* from cell *i*.
+        radius (int):
+            Size of the radius, in pixels, used to featurize cellular
+            neighborhoods.
+        cell_index (str):
```
ngreenwald, line 102: If we default to this, do we need to add index in `format_cell_table`?

ark/spLDA/processing.py
ngreenwald: It looks like the "marker" and "avg_marker" functions require the `markers` argument, whereas "cluster" and "count" do not

ark/spLDA/processing.py
ngreenwald: Is there a scenario where we wouldn't do a train split? We could just set train_frac to 1, right?

ark/spLDA/processing.py
```diff
@@ -63,24 +52,153 @@ def format_cell_table(cell_table, markers=None, clusters=None, fovs="all_fovs"):
...
+        train_diff_mat = None
+
+    if inference:
+        inference_diff_mat = make_merged_difference_matrices(
+            sample_features=features["featurized_fovs"],
+            sample_dfs=cell_table,
+            x_col="x", y_col="y")
+    else:
+        inference_diff_mat = None
+
```
ngreenwald, line 202: I don't understand the point of this function. How could you be doing training and inference at the same time? Why do we need this wrapper around make_merged_difference_matrices?

ark/utils/spatial_lda_utils.py
```diff
@@ -1,63 +1,116 @@
...
-        "name"
-    ]
-
     # Check cell table
-    if not isinstance(cell_table, pd.DataFrame):
-        raise ValueError("cell_table must be a pd.DataFrame")
-
-    if not all([True for x in base_cols if x in cell_table.columns]):
-        raise ValueError("cell table must contain the following columns:{}".format(base_cols))
+    if not all([x in cell_table.columns for x in BASE_COLS]):
```
ngreenwald, line 21: We have a helper function that identifies the specific element from the list that is missing: https://github.com/angelolab/ark-analysis/blob/16466141fbf6ab8bffb6b4e613010bbb763f4f64/ark/utils/misc_utils.py#L120

ark/utils/spatial_lda_utils.py
```diff
@@ -1,63 +1,116 @@
...
+        raise ValueError(
+            "cell table must contain the following columns:{}".format(
+                BASE_COLS))
 
     # Check markers/clusters
     if all([markers is None, clusters is None]):
```
ngreenwald, line 27: See Alex's comment in original PR about more pythonic phrasing for some of these tests

ark/utils/spatial_lda_utils.py
```diff
@@ -1,63 +1,116 @@
...
+            Name of the column in each field of view pd.Dataframe indicating
+            reference cells.
+
+    """
+    # Check valid data types
+    if not isinstance(cell_table, dict):
+        raise TypeError("cell_table should be of type 'dict'")
+    if not isinstance(cell_table[1], pd.DataFrame):
+        raise TypeError("cell_table should contain formatted pd.DataFrames")
+    if not isinstance(feature_by, str):
```
ngreenwald, line 66: Since we already check whether `feature_by` has the exact four values that are required, I don't think this test is necessary. I would say a fair number of these are overkill

ngreenwald (changes requested):
Great start, left some comments. There were a few places where same comment can be applied to multiple lines that had same issue

ark/spLDA/processing.py
bcollica: that's also true

ark/spLDA/processing.py
```diff
@@ -1,56 +1,45 @@
...
-        "centroid-1",
-        "pixelfreq_hclust_cap",
-        "name"
-    ]
     if markers is not None:
-        keep_columns.append(markers)
-    drop_columns = [c for c in all_columns if c not in keep_columns]
-    cell_table = cell_table.drop(columns = drop_columns)
+        BASE_COLS.append(markers)
+    drop_columns = [c for c in cell_table.columns if c not in BASE_COLS]
```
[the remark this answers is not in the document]
bcollica: You mean pixelfreq_hclust_cap? And yes it would keep it and also rename it as cluster_id.

ark/spLDA/processing.py
bcollica: I'm not sure what you mean.

ark/spLDA/processing.py
bcollica: Both the train() and infer() functions from the spatial_lda library take a difference matrix as one of their arguments.  In train(), the difference matrix is computed using the training fraction from the featurized data and the full formatted data.  In infer(), the difference matrix is calculated using the full featurized data and the full formatted data.  The wrapper function computes one of (or both of) these in one step so they can be passed to the respective training and inference functions later rather than having the user explicitly call them separately -- less steps for them.

ark/spLDA/processing.py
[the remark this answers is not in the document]
ngreenwald: I was worried this would mean that the cluster_freqs were getting included in the spatial lda computation, but that's not the case because the featurization function explicitly takes the list of markers. Given that, what is the purpose of subsetting this cell_table to only include the relevant markers if the featurization function already does that? Just memory reason?

ark/spLDA/processing.py
ngreenwald: Is this line necessary if we have this option enabled? 
image

ngreenwald: Meaning if cell_index is none, it assumes all cells are part of the index. So we can have cell_index be none, and not need to add a cell_index column to the data_table?

ark/spLDA/processing.py
ngreenwald: Okay, maybe once I see how this gets called it will make more sense to me. Right now, it seems like you would call this function once during training, and once during inference. But calling this function isn't any less work than calling make_merged_difference_matrices, so I'm not sure I see the benefit.

ark/spLDA/processing.py
alex-l-kong: This works but it's probably best to assign this to a separate variable first before returning. I like to keep my return statements simple 😂.

ark/utils/spatial_lda_utils.py
```diff
@@ -1,63 +1,116 @@
...
     if all([markers is None, clusters is None]):
         raise ValueError("markers and clusters cannot both be None")
     if markers is not None:
         if isinstance(markers, list) and len(markers) == 0:
             raise ValueError("list of marker names cannot be empty")
-        if not isinstance(markers, list) or not all([isinstance(x, str) for x in markers]):
-            raise ValueError("clusters must be a list of integers")
-        assert all([x in cell_table.columns for x in markers])
+        if not isinstance(markers, list) or not all(
+                [isinstance(x, str) for x in markers]):
```
alex-l-kong, line 33: If you've already specified that the datatype of `markers` is `list` in the docstring, there's no need to check for its datatype again. The same applies to the other places where the `isinstance` check is.

alex-l-kong (changes requested):
Looks good! I left a lot of stuff in the original branch, some of that may also apply here so please check that out too.

ark/spLDA/processing.py
bcollica: The featurize_samples() function needs a reference column in each FOV which it uses to run the featurization step.  In this case, "is_index" tells that function that all cells are to be used.  Alternatively, the user could specify any other existing column name that has values equal True for reference cells and False for non-reference cells, but they would have had to create this column on their own beforehand.  So either way a reference column has to exist.

ngreenwald: kk, sounds good

ark/spLDA/processing.py
```diff
@@ -1,86 +1,186 @@
...
+                         n_processes=None, train_frac=0.75):
+    """Calculates statistics for local cellular neighborhoods based on the specified features
+    and radius.
 
+    Args:
+        cell_table (dict):
+            A formatted cell table for use in spatial-LDA analysis. Specifically, this is the
+            output from :func:`~ark.spLDA.processing.format_cell_table`.
+        featurization (str):
+            One of four choices of featurization method, defaults to "cluster" if None is provided:
```
ngreenwald, line 82: Here's an example where setting the default to None isn't as useful. Because we know exactly what we're going to substitute None for ("cluster"), having None as a placeholder isn't better than just having the default be "cluster". When the default is None for all FOVs, it's a bit different, because the user isn't actually passing in all of the FOVs, they're passing in something which is telling us to use all the FOVs.

ark/spLDA/processing.py
```diff
@@ -1,86 +1,186 @@
...
+        dict:
 
+        - A dictionary containing a DataFrame of featurized cellular neighborhoods and a
+        separate DataFrame for designated training data.
+    """
+
+    # Default to "cluster", "is_index if featurization, cell_index are not provided.
+    if featurization is None:
+        featurization = "cluster"
+    if cell_index is None:
```
ngreenwald, line 110: Same with is_index.

ark/utils/spatial_lda_utils.py
```diff
@@ -1,63 +1,91 @@
...
+    # Check cell table columns
+    verify_in_list(cell_table_columns=cell_table.columns.to_list(), required_columns=BASE_COLS)
 
     # Check markers/clusters
-    if all([markers is None, clusters is None]):
+    if markers is None and clusters is None:
         raise ValueError("markers and clusters cannot both be None")
     if markers is not None:
-        if isinstance(markers, list) and len(markers) == 0:
+        if len(markers) == 0:
```
ngreenwald, line 26: What happens if you pass an empty list to verify_in_list? It should still give the right error message, which is the markers that are missing right?

ark/utils/spatial_lda_utils.py
```diff
@@ -1,63 +1,91 @@
...
-        assert all([x in cell_table.columns for x in markers])
+        verify_in_list(markers=markers, cell_table_columns=cell_table.columns.to_list())
     if clusters is not None:
-        if isinstance(clusters, list) and len(clusters) == 0:
+        if len(clusters) == 0:
             raise ValueError("list of cluster ids cannot be empty")
-        if not isinstance(clusters, list) or not all([isinstance(x, int) for x in clusters]):
-            raise ValueError("markers must be a list of strings")
+        if not isinstance(clusters, list) or not all(
+                [isinstance(x, int) for x in clusters]):
```
ngreenwald, line 33: We should modify this test to mirror the one above, using `verify_in_list` to ensure that the clusters which were provided are present in the cell_table

ark/utils/spatial_lda_utils.py
```diff
@@ -1,63 +1,91 @@
...
+            view.
+        featurization (str):
+            One of "cluster", "marker", "avg_marker", or "count".
+        radius (int):
+            Pixel radius corresponding to cellular neighborhood size.
+        cell_index (str):
+            Name of the column in each field of view pd.Dataframe indicating reference cells.
+    """
+    # Check valid data types
+    if not isinstance(cell_table, dict):
```
ngreenwald, line 52: I think we can assume that they passed the output of `format_cell_table` to this function, so don't need to check that it's a dict of dfs

ark/utils/spatial_lda_utils.py
```diff
@@ -1,63 +1,91 @@
...
+
+    verify_in_list(featurization=[featurization],
+                   featurization_options=["cluster", "marker", "avg_marker", "count"])
+    verify_in_list(cell_index=[cell_index], cell_table_columns=cell_table[1].columns.to_list())
+
+    if radius < 25:
+        raise ValueError("radius must not be less than 25")
+
+
+def check_create_difference_matrices_args(cell_table, features, training, inference):
```
ngreenwald, line 67: I think we can delete this function, and just have the bottom two checks of this function at the top of  `create_difference_matrices`

ngreenwald (changes requested):
Looks good, some simplifications to the tests

ackagel (changes requested):
Looking good.  Just a note on keeping arguments immutable and a suggestion for an argument validation pattern

ark/utils/spatial_lda_utils.py
bcollica: Nope, no error is thrown.  
`
import ark.utils.misc_utils as ms
ms.verify_in_list(l1=[], l2=["a", "b", "c"])
`

bcollica: This does though. 
 `ms.verify_in_list(l1=["a", "b", "c"], l2=[])`

Sorry, the function wording is somewhat confusing, is list 1 is supposed to contain the required values?

ngreenwald: No, you're correct. We check that all the entries in list1 are contained within list2, but we don't check that list1 is non-empty. Can you add a line to that function that checks that both lists are non-empty?

ngreenwald (changes requested):
Looks great! Just one minor comment to fix the helper function, and then remove the redundant checks for list length

ark/utils/spatial_lda_utils.py
```diff
@@ -1,63 +1,52 @@
...
+    """
+    # Check valid data types and values
+    if not isinstance(radius, int):
+        raise TypeError("radius should be of type 'int'")
+    if radius < 25:
+        raise ValueError("radius must not be less than 25")
+
+    verify_in_list(featurization=[featurization],
+                   featurization_options=["cluster", "marker", "avg_marker", "count"])
+    verify_in_list(cell_index=[cell_index], cell_table_columns=cell_table[1].columns.to_list())
```
alex-l-kong, line 52: To save you a slight amount of trouble, `verify_in_list` will work if you just set `featurization=featurization`, it's not necessary to use a list for a single element. Not a problem if you prefer doing it this way though.

ark/spLDA/processing.py
```diff
@@ -1,86 +1,186 @@
...
-import pickle
-from ark.utils.spatial_lda_utils import check_format_cell_table_args
-# from spatial_lda.featurization import *
-# from spatial_lda.visualization import plot_adjacency_graph()
+from sklearn.model_selection import train_test_split
+from spatial_lda import featurization as ft
 
+from ark.settings import BASE_COLS
+from ark.utils.spatial_lda_utils import check_format_cell_table_args, \
+    check_featurize_cell_table_args
```
alex-l-kong, line 9: If you don't expect `spatial_lda_utils` to become extremely massive, I recommend just importing the entire `ark.utils.spatial_lda_utils` module. Generally, I try not to use line continuations for import statements.

Another option is to break this into two separate imports or to use shorter function names.

ark/utils/misc_utils.py
```diff
@@ -135,6 +135,8 @@ def verify_in_list(**kwargs):
         raise ValueError("You must provide 2 arguments to verify_in_list")
 
     test_list, good_values = kwargs.values()
+    if len(test_list) == 0 or len(good_values) == 0:
+        raise ValueError("List arguments cannot be empty")
```
alex-l-kong, line 139: Were we running into issues with empty lists, or is this simply an extra error check?

requirements.txt
```diff
@@ -12,3 +12,4 @@ statsmodels>=0.11.1,<1
 umap-learn>=0.4.6,<1
 xarray>=0.12.3,<1
 tqdm>=4.54.1,<5
+spatial_lda>=0.1.0
```
alex-l-kong, line 15: We sometimes get sloppy about this but let's keep the `requirements.txt` file sorted.

alex-l-kong (commented):
A few comments to add.

ark/utils/spatial_lda_utils.py
bcollica: It seems single elements on work for verify_in_list if it's a single string, and even then the error message gets funky because it breaks up the string into individual characters when it calls 
`bad_vals = ','.join([str(val) for val in test_list if val not in good_values])`

Also, integers and floats get hung up on this line and the code throws TypeError anyway because ints and floats are not iterable.

ark/utils/misc_utils.py
bcollica: See Noah's comment above

requirements.txt
bcollica: Sorted alphabetically you mean?

alex-l-kong: @bcollica yup.

ark/utils/spatial_lda_utils.py
ngreenwald: Can also cast the inputs to a list in the helper function if they're not supplied as one to avoid this

bcollica: Already done lol

alex-l-kong (commented):
Great work so far!

ngreenwald (approved):
Awesome work

bcollica:
🥳

PR #440, opened 2021-08-03
Create GoogleDrivePath Usage Documentation

ackagel:
Instead of an example notebook, a full documentation page on using the `GoogleDrivePath` API is added. 

This mitigates the confusion users would have trying to run the example notebook, and lets them gradually familiarize themselves with `GoogleDrivePath` usage as they need it.


docs/_rtd/google_drive_usage.md
```diff
@@ -0,0 +1,166 @@
...
+    elif not os.path.exists(directory):
+        os.makedirs(directory)
+```
+
+Furthermore, we may not want all of these to be `GoogleDrivePath`s.  In the context of the 
+`Segment_Image_Data` notebook, `deepcell_input_dir` only stores the Mesmer-compatable formatted
+data, while `deepcell_output_dir` stores the results of the Mesmer segmentation.  
+
+We probably don't need to upload/download this re-formatted data, and wouldn't mind if it's stored
+locally since it's very small relative to the size of the dataset.  On the other hand, we'll
```
alex-l-kong, line 100:
```suggestion
locally since it's very small relative to the size of the dataset.  On the other hand, we
```

docs/_rtd/google_drive_usage.md
```diff
@@ -0,0 +1,166 @@
...
+
+ 1. We don't necesarily want all paths to be `GoogleDrivePath`s
+ 2. We should always create new folders regardless of if they're local or Google Drive folders
+
+### Running notebooks
+
+Now that your paths are formatted, most `ark-analysis` data-processing functions should work
+out of the box, as if your `GoogleDrivePath`s were actual local files.  If you come across an issue
+using a `GoogleDrivePath` within an `ark-analysis` function, please submit an issue here
+(TODO: add a link here...).
```
alex-l-kong, line 128: https://github.com/angelolab/ark-analysis/issues

docs/_rtd/google_drive_usage.md
```diff
@@ -0,0 +1,166 @@
...
+For the generalized example above, we can make it `GoogleDrivePath` compatable via:
+```
+google_drive_utils.drive_write_out( some_path, lambda x: df.to_csv(x, index=False) )
+```
+Note that this lambda function could also be a regular function:
+
+```
+def to_csv_wrapper(x):
+    df.to_csv(x, index=False)
+```
```
alex-l-kong, line 165: This part shouldn't be necessary.

alex-l-kong (changes requested):
Looks good, just a few minor changes.

docs/_rtd/google_drive_usage.md
```diff
@@ -0,0 +1,158 @@
+## Using Google Drive
+
+### Motivation
+
+Image data can be very large, and it can be difficult to keep all of it on you local hard drive.
```
bcollica, line 5:
```suggestion
Image data can be very large, and it can be difficult to keep all of it on your local hard drive.
```

bcollica (commented):
This is great, I think it's much more intuitive presented this way.

ngreenwald (changes requested):
Overall looks great

docs/_rtd/google_drive_usage.md
```diff
@@ -0,0 +1,158 @@
...
+For now, using the `GoogleDrivePath` API requires a universal passcode.  Email
+`ackagel@stanford.edu` to request access for this features.
+
+Once you have this passcode, you can make all of your notebooks `GoogleDrivePath`-compatible by
+running this code cell at the top
+
+```
+from ark.utils import google_drive_utils
+
+google_drive_utils.init_google_drive_api("passcode")
```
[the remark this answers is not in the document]
ackagel: They would replace the `"passcode"` string with the actual passcode yep.  I can clarify that if need be.

Also, now that I think about it, I should add some details for signing in to their google drive account after running that code cell.

docs/_rtd/google_drive_usage.md
```diff
@@ -0,0 +1,158 @@
...
+```
+for directory in [deepcell_input_dir, deepcell_output_dir, single_cell_dir, viz_dir]:
+    if type(directory) is GoogleDrivePath:
+        directory.mkdir()
+    elif not os.path.exists(directory):
+        os.makedirs(directory)
+```
+
+Furthermore, we may not want all of these to be `GoogleDrivePath`s.  In the context of the 
+`Segment_Image_Data` notebook, `deepcell_input_dir` only stores the Mesmer-compatable formatted
```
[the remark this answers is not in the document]
ackagel: Only non-negligible hinderance on performance would be in the case of poor upload **_or_** download speed.

With that being the case, maybe I just wrap this comment around an **'in the case of slow upload/download speed'** section?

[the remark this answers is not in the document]
ngreenwald: Yeah, move this to the bottom, and change from "this is what you should do" to "this is one way to remedy slow upload/download"

alex-l-kong (commented):
Looks good to me!

ngreenwald (approved):
Looks good

PR #441, opened 2021-08-09
Add randomization module to utils

alex-l-kong:

Much of `ark-analysis` requires random sampling, and in light of future requests for random samples of the data, we should create a separate module containing several different randomization methods.

We add `rand_utils.py` to the `utils` folder.

**Remaining issues**

I have written other randomization methods but for now, the only one I've added is a simple one to do a fully random sample.


alex-l-kong:
Should be a pretty easy review, just want to get a sense of:

- The current structure
- Any other randomization functions related to pixel/cell clustering that we need to add?

alex-l-kong:
Should be a pretty easy review, just want to get a sense of:

- The current structure
- Any other randomization functions related to pixel/cell clustering that we need to add?

ngreenwald (changes requested):
Nice bite-sized piece!

There are two main types of randomization I see people wanting to do in the near future. 

1) Train a SOM on a dataset that is imbalanced in some way, and correct that imbalance based on image-level labels. For example, 100 images, 20 of them from disease, 80 from healthy. The SOM should be trained on 10 million pixels, 5 million from disease and 5 million from healthy. 

2) Train a spatial LDA model on an imbalanced dataset. For example, 20 images from disease and 80 from healthy. The spatial LAD model should be trained on 20 images from disease and 20 from healthy. 

For both cases, the first step will be reading in a pandas array that has the FOV names as one column, and the disease label as a second column. However, for the SOM application we'll want the output to be an additional column in the pandas array specifying an image-specific weighting for how many pixels to subset. For the spatial LDA application we'll want the output to be the specific FOVs which will be included in training the model, since we don't subset.

ackagel (commented):
Looking good so far

bcollica (commented):
Something else useful might include a wrapper for bootstrapping operations.  It comes up in the spatial enrichment analysis and also for a kmeans cluster metric I'm looking at implementing.  Plus I can see it being useful in other areas.

PR #446, opened 2021-08-16
Add random sampling module

alex-l-kong:

Closes #441. Reopen of a previous PR.


alex-l-kong:
Per Noah's last review:

Nice bite-sized piece!

There are two main types of randomization I see people wanting to do in the near future.

Train a SOM on a dataset that is imbalanced in some way, and correct that imbalance based on image-level labels. For example, 100 images, 20 of them from disease, 80 from healthy. The SOM should be trained on 10 million pixels, 5 million from disease and 5 million from healthy.

Train a spatial LDA model on an imbalanced dataset. For example, 20 images from disease and 80 from healthy. The spatial LAD model should be trained on 20 images from disease and 20 from healthy.

For both cases, the first step will be reading in a pandas array that has the FOV names as one column, and the disease label as a second column. However, for the SOM application we'll want the output to be an additional column in the pandas array specifying an image-specific weighting for how many pixels to subset. For the spatial LDA application we'll want the output to be the specific FOVs which will be included in training the model, since we don't subset.

alex-l-kong:
I've removed anything from `som_utils` as this branch is getting merged directly into `master`. After this is merged into `master`, I'll merge that into `new_som` and we can add the randomization calls to the SOM stuff then.

ngreenwald:
Closing this, we ended up not needing this in pixel clustering

PR #447, opened 2021-08-24
Create datasets that integrate with remapping GUI

alex-l-kong:

Zak has added an interactive visualization of cluster assignment data. This will require two datasets:

1. A table that computes the average marker expression and total pixel count per pixel cluster
2. A table that computes the weighted expression of each marker in each cell.

1. `compute_pixel_cluster_channel_avg` already computes the average marker expression. We can add an option to create a column with the total pixel count as well.
2. `compute_cell_cluster_channel_avg` computes the weighted expression of each marker in each cell.


alex-l-kong:
Tests are failing because a new version of `cryptography` was released under 15 minutes ago. I'll open an issue regarding this soon.

ark/phenotyping/som_utils.py
```diff
@@ -166,6 +171,75 @@ def compute_cell_cluster_avg(cluster_path, column_prefix, cluster_col):
     return mean_count_totals
 
 
+def compute_cell_cluster_channel_avg(pixel_channel_avg, cell_counts,
```
ngreenwald, line 174: What is the point of doing the calculation this way? Why take a weighted average of an average, rather than just calculating the expression of each marker in each cell? Is this faster?

ngreenwald (changes requested):

ackagel (commented):
Implementation looks good.  Seconding @ngreenwald's comment, this seems like a round about way of getting/approximating that data when we can just get it directly.

Further, I also don't know why we'd weight marker expression by cluster counts.  We've only weighted raw marker expression by individual cell size in the past to get 'marker density/concentration' within the cell.  I can't immediately see the physical interpretation for expression weighted by a global cluster's prevalence, but I could just be confused.

ngreenwald:
@cliu72 was there a reason you set it up this way?

templates/example_cell_clustering.ipynb
```diff
@@ -0,0 +1,500 @@
+{
```
ngreenwald, line 1: <p>I don't see a None here anymore, I don't think this comment applies</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/447/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

ngreenwald (changes requested):
Looks good to me, just one comment in the notebook that (I think) needs to be updated. @cliu72, will these two different functions both be used to generate inputs to the interactive reclustering? Or is one needed for pixel clustering, and the other for cell clustering?

templates/example_cell_clustering.ipynb
alex-l-kong: This should be `None`, I just forgot to reassign it back to `None` since I was testing it locally.

cliu72:
Two functions are needed because to generate inputs for Zak's tool, expression is summed for each pixel for pixel clustering, while weighted sum of pixel clusters is calculated for each cell for cell clustering. Hope that answered the question, but I'm not sure which comment you're talking about.

alex-l-kong:
@cliu72 it was a comment unrelated to these two functions that was confusing because I forgot to change the default assignment of `pixel_mat_consensus` in `example_cell_clustering.ipynb.`

ngreenwald:
Got it, that's what I thought. In that case Alex, no need to run the pixel sum function in the cell clustering notebook.

ngreenwald (changes requested):
See above

cliu72:
The ranges for the heatmaps in both notebooks look funky. Values should be z-scores across the markers, range should be around -3 to 3. Not sure why the values in the cell clustering notebook goes up to the hundreds. Can you also add in code so that 0 in the heatmaps is always white? And blue is negative values and red is positive.

Also, Zak's notebook takes CSVs as input. It's not a big deal, but since these are averaged tables that are being stored, I don't think there's a memory issue with storing them as CSVs. If you want to keep the feather format, we need to add a few lines to Zak's code to make the conversion.

alex-l-kong:
@cliu72 I'll be opening a PR right after this to address all the visualization requests, so that'll be addressed then too.

alex-l-kong:
@ngreenwald I'm assuming `compute_pixel_cluster_channel_avg` is the pixel sum function you're talking about. So I'm not actually running `compute_pixel_cluster_channel_avg` in `example_cell_clustering.ipynb` (it runs in `example_pixel_clustering.ipynb`), though I do load the results computed in the pixel clustering notebook in since `compute_cell_cluster_channel_avg` needs that to do its computations. Did you want this loading cell removed or a different cell?

ngreenwald:
Ah, you're right, I saw the loading and assumed it was being computed. Nope, it's fine as is.

ngreenwald (approved):

PR #451, opened 2021-08-25
Update spatial analysis

ackagel:
## **Purpose**

The spatial analysis notebook doesn't support the current saved label format, doesn't support batching (although that might be hard to add conceptually), and is generally pretty lackluster feature/documentation wise.  This PR hopes to update the notebook, clarify implementation details, and add QoL features, such as an example cell-cluster/expression image plot.

## **Implementation**

So far just extracting tifs from the `.xr` labels file, and converting the notebook to load those tiffs, instead of the `.xr` file.  More changes to come from separate branches.

## **Remaining issues**

Only the labels loading interface has been updated so far.  The rest of the features mentioned still need to be implemented.


alex-l-kong:
@ackagel Marte and I have been running into an issue with importing `ark.utils._bootstrapping` with the current Docker. Suspect it's a Cython issue. While this is being addressed, what can we do to set this up locally?

ackagel:
Depends what the issue is; this might be that 'oldest-supported-numpy' requirement that rears its head every once and awhile, in which case, I have a perma-fix for that in mibi-bin-tools I should add here.  Can you post a screenshot?

However if it's just a regular 'import not found' type error, then you should be able to manually install it via the jupyter lab terminal

PR #458, opened 2021-08-30
Fov Batching for Spatial Analysis

ackagel:
## **Purpose**

Batches spatial enrichment computations over FOVs to accommodate large datasets.

## **Implementation**

Wrapper functions for the high level channel and cluster spatial enrichment functions.


ark/analysis/spatial_analysis.py
```diff
@@ -1,12 +1,77 @@
...
+
+    batching_strategy = \
+        [all_label_names[i:i + batch_size] for i in range(0, len(all_label_names), batch_size)]
+
+    # create containers for batched return values
+    values = []
+    stats_datasets = []
+
+    # extract 'included_fovs' kwarg
+    included_fovs = kwargs.get('included_fovs', None)
```
ngreenwald, line 46: We should filter `all_label_names` instead, right? Then we don't have to compute/batch over FOVs we don't care about?

ark/analysis/spatial_analysis.py
```diff
@@ -116,6 +181,69 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds,
     return values, stats
 
 
+def batch_cluster_spatial_enrichment(label_dir, all_data, batch_size=5, **kwargs):
```
ngreenwald, line 184: We could combine this function with previous batching function. We could add an "enrichment_type" argument, and if channels, have thresholds be a required argument. 

If you think that would be too messy can keep as is, but almost all the lines are identical

ngreenwald (changes requested):
Nice small PR!
See comments below

alex-l-kong (commented):
Nothing else to add on top of Noah's comments.

ark/analysis/spatial_analysis.py
ackagel: So, it looks like there's a way to generate fov-batching versions of a bunch of functions (cell table generation, spatial enrichment, neighborhoods, etc) by using 'decorator-like' function factories.  It does look a little abstract, and is a little outside of the focus/scope of this PR, so I might leave these as separate for now, and submit a 'generic fov-batching' PR with that solution.

ngreenwald (approved):

PR #459, opened 2021-09-03
Generic batching decorators

ackagel:
## **Purpose**

There are a lot of operations that will need to be batch processed over fovs for large cohorts.  This PR seeks to add a way to easily add batch support for 'per-fov' operations.

## **Implementation**

Adds the `batch_utils.py` file, which holds functions for creating generic 'batching strategies' as well as a function decorator `@make_batch_function(func, strategy)`.   So far, this successfully implements batching for the spatial enrichment calculation functions, and is able to automatically determine the appropriate function signature to compare with the user provided docstring.  See `spatial_analysis.batch_channel_spatial_enrichment`.

## **Remaining issues**

Not really sure how to test `batch_utils`.  Also, getting the autodocs working for these functions is the largest reason for the added complexity involving `exec()`.  If the complexity doesn't seem worth clean docs, then setting the docs `conf.py` to ignore argument checks on functions prefixed with `batch_` may be in order, but this necessitates stronger review of documentation on `batch_` functions...


ngreenwald:
Stale, closing for now

PR #462, opened 2021-09-08
Context dependent spatial randomization

ackagel:
## **Purpose**

Add context based spatial randomization to spatial analysis package.  This restricts the positional randomization to occur between pairs of contexts (catagorical data, e.g. metaclusters), as opposed to global, whole fov positional randomization.

## **Implementation**

`calculate_*_spatial_enrichment` functions have an additional `context_col` argument for a context column in the provided cell table.  If this argument is None, global randomization is used, but otherwise, all unique context pairings are calculated, and randomizations are repeated and summed for each pairing.

Note that this makes no changes to the `compute_close_cell_num_random` function, and instead alters the given arguments.  This prevents requiring a new randomization function with several new arguments to accommodate context-dependence.

## **Remaining issues**

Only `calculate_cluster_spatial_enrichment` has been templated.  There are no test functions at the moment.  Currently, this change is not represented in the spatial analysis notebook.


ackagel:
This is the general idea for adding context dependence.  Lmk if anything looks blatantly wrong.  Also, @alex-l-kong if you have some example/test data, I'm going to need that, to make sure everything's working as expected.

ark/analysis/spatial_analysis.py
```diff
@@ -324,7 +324,14 @@ def calculate_cluster_spatial_enrichment(all_data, dist_matrices_dict, included_
...
     # Only include the columns with the patient label, cell label, and cell phenotype
-    all_pheno_data = all_data[[fov_col, cell_label_col, cluster_id_col]]
+    all_pheno_data_cols = [fov_col, cell_label_col, cluster_id_col]
+    all_pheno_data_cols += [] if context_col is None else [context_col]
+
+    all_pheno_data = all_data[all_pheno_data_cols]
+
+    if context_col is not None:
+        context_names = all_data[context_col].unique()
+        context_pairings = combinations_with_replacement(context_names, 2)
```
alex-l-kong, line 334: Nice!

alex-l-kong (commented):
Looks good so far. Give me some time to sort through the junk on my computer and I'll have the test data over.

ark/analysis/spatial_analysis.py
ngreenwald: How does this handle empty rows? I assume np.unique() will treat empty as a unique value?

Lets say the output from cell clustering is 20 different cell types. There will be one column where each row (i.e. cell) is assigned to one of those possible cell types. For context-dependent randomization, we might only be interested in the interactions between the 7 different immune celltypes that are there. We could make a separate column that only includes the immune celltypes, but then the other cell types won't be labeled. 

The other option would be to pass in a list of values that we want to context randomize over, but that could also get messy.

ark/analysis/spatial_analysis.py
```diff
@@ -342,12 +349,40 @@ def calculate_cluster_spatial_enrichment(all_data, dist_matrices_dict, included_
...
             dist_mat=dist_mat, dist_lim=dist_lim, analysis_type="cluster",
             current_fov_data=current_fov_pheno_data, cluster_ids=cluster_ids)
 
-        close_num_rand = spatial_analysis_utils.compute_close_cell_num_random(
-            pheno_nums, dist_mat, dist_lim, bootstrap_num)
+        # subset distance matrix with context
+        if context_col is not None:
+            close_num_rand = np.zeros((*close_num, bootstrap_num), dtype=np.uint16)
+
+            context_nums_per_id = \
```
ngreenwald, line 360: What is this one-liner doing?

ngreenwald (changes requested):
Overall looks very clean. See below for my question about the best way to format the requirement for context-dependent randomization inputs.

ark/analysis/spatial_analysis.py
ngreenwald: Actually I rescind my comment. The way this context randomization was originally done by Leeat wasn't just for one specific subtype, but across a couple different celltypes. I think this is a good way to set things up for now.

ark/analysis/spatial_analysis.py
ackagel: This creates a dictionary with context names as keys and lists of member cell's label values as values.  i.e:
```
context_nums_per_id = {
    'immune': [13, 490, 793, ...],
    'epithelial': [17, 289, 391, ...],
    ...
}
```

ark/analysis/spatial_analysis.py
ackagel: In the futre, we can add a 'context filter' to filter cell tables to only compute spatial enrichment between cells in the given list of contexts.  That's a simple as adding this line:
```
filtered_cell_table = cell_table[cell_table[context_col].isin(contexts)]
```

ngreenwald:
Got it, looks good to me

ackagel:
previously, the context based randomization functioned as follows:
 - for all cells in each context pairing, all cells of lineage (i, j) were randomly placed within the allowable context labels

now, the added randomization asymmetry warrants a revaluation of what we want from the context.  The interpretations I can see:
 - a cell (lineage=i, context='square') chooses |num_cells[j]| from all cells in context='square' (doesn't feel right, plus what if `|'square'| < |num_cells[j]|`)
 - a cell (lineage=i) chooses |num_cells[j]| where all cells have been randomized within their respective contexts (feels right).

ngreenwald:
Okay, neither of those is that intuitive to me. We can discuss tuesday. The original implementation of context-random should be in the matlab code too

ackagel:
kk, we'll talk Tuesday.  Reading over the matlab version, I'm pretty sure the original implementation is symmetric in the sense that each marker/lineage pair was randomized, not just marker/lineage `j`.

ackagel:
Looks like the diff is broken for some reason, so to see the actual difference between this branch and the target, you can use this link:

https://github.com/angelolab/ark-analysis/compare/spatial-analysis-nb-update...context-dependent-randomization

ngreenwald:
Updated diff is super concise, no complaints on my end. Maybe once you merge this into the main branch github will update? Probably the result of it not being refreshed in a while

PR #463, opened 2021-09-09
Close num rand speedup

ackagel:
## **Purpose**

Despite our best efforts,`spatial_analysis_utils.compute_close_cell_num_random` is still slow.  It's so slow that even a bootstrap number of 100 causes the call to `compute_channel_spatial_enrichment` in the example notebook to take several minutes.  The fix would be simple if a more efficient algorithm existed, but none has been verified so far (though several have been tested).

This PR adds a ~10-30x speedup to the bootstrapping calculation by implementing the bulk of the iterations in cython.  In more quantifiable terms, a realistic `bootstrap_num=10000`, `calculate_cluster_spatial_enrichment` call in the demo notebook takes ~5 minutes, compared to the 30-90 mins it previously took.

## **Implementation**

cython compiles the code in `ark/utils/_bootstrapping/_close_cell_num_random.pyx` to c code which can be called just like a python module.  Currently, the same algorithm as the python version is used, only making edits to minimize python interpreter interaction. 

As for travis, the cython code should be automatically compiled, so that all tests work as the normally did.  Furthermore, since we're just embedding previous functionality into C, we can just test the wrapping python code, as we have been; i.e no new tests are needed.

### Messy cython
Firstly, the cython code is a little messy and could be cleaned up (stray comment, useless function arguments, stricter type declaration/binding, etc).

### Optimal algorithm
The line by line profiling show that the deepest iteration takes ~99.7% of the computation time.  This isn't because the accumulation or array index is slow, it's simply because there are on average `O(C^2 * m^4 * r)` additions, where `C` is number of cells, `m` is number of 'markers' (channels or clusters), and `r` is the bootstrapping number.  Because `C` will be massive, this is bad (this also explains why channel randomization is much slower than cluster randomization as num_channels > num_markers).

Theoretically, the most optimal time complexity for this task that I can think of is `O(C * D * m^4 * r)` where `D` is the average count of cells within the set distance limit.  Implementing an algorithm with this complexity requires a low-level 'hash-table-like' container for randomized columns, to search for positive columns per row within the binarized distance matrix.  Because `D` is constant for increasing `C, m, r`, this essentially moves the algorithm to be linear in number of cells, as opposed to quadratic.

I haven't tried implementing this solution yet and want to do that to test if the speedups are substantial

### DIFF
--- a/ark/utils/spatial_analysis_utils.py
+++ b/ark/utils/spatial_analysis_utils.py
@@ -11,6 +11,7 @@ from scipy.spatial.distance import cdist
 
 import ark.settings as settings
 from ark.utils import io_utils, misc_utils
+from ark.utils._bootstrapping import compute_close_num_rand
 
 
 def calc_dist_matrix(label_maps, save_path=None):
@@ -246,33 +247,15 @@ def compute_close_cell_num_random(marker_nums, dist_mat, dist_lim, bootstrap_num
             Large matrix of random positive marker counts for every permutation in the bootstrap
     """
 
-    # Get marker counts
-    marker_count = len(marker_nums)
-
-    # Create close_num_rand
-    close_num_rand = np.zeros((
-        marker_count, marker_count, bootstrap_num), dtype=np.uint16)
-
     # Generate binarized distance matrix
     dist_mat_bin = (dist_mat.values < dist_lim).astype(np.uint16)
 
     # static choice array
-    choice_ar = list(range(dist_mat.shape[0]))
+    _marker_nums = np.array(marker_nums, dtype=np.uint16)
+    _choice_ar = np.array(list(range(dist_mat.shape[0])), dtype=np.uint16)
 
-    for j, m1n in enumerate(marker_nums):
-        for k, m2n in enumerate(marker_nums[j:], j):
-            for r in range(bootstrap_num):
-                # Select same amount of random cell labels as positive ones in close_num
-                marker1_labels_rand = np.random.choice(a=choice_ar, size=m1n, replace=False)
-                marker2_labels_rand = np.random.choice(a=choice_ar, size=m2n, replace=False)
-
-                # Record the number of interactions and store in close_num_rand in the index
-                # corresponding to both markers, for every permutation
-                close_num_rand[j, k, r] = \
-                    np.sum(dist_mat_bin[np.ix_(marker1_labels_rand, marker2_labels_rand)])
-
-            # System should be symetric
-            close_num_rand[k, j, :] = close_num_rand[j, k, :]
+    close_num_rand = compute_close_num_rand(dist_mat_bin, _marker_nums, _choice_ar,
+                                            int(bootstrap_num))
 
     return close_num_rand
 
--- a/setup.py
+++ b/setup.py
@@ -1,15 +1,23 @@
 import setuptools
 from os import path
-from distutils.command.build_ext import build_ext as DistUtilsBuildExt
-from setuptools import setup, find_packages
+from setuptools import setup, find_packages, Extension
+import numpy as np
+from Cython.Build import cythonize
 
 VERSION = '0.2.9'
 
+with open(path.join(path.dirname(__file__), 'requirements.txt')) as req_file:
+    requirements = req_file.read().splitlines()
 
 # set a long description which is basically the README
 with open(path.join(path.abspath(path.dirname(__file__)), 'README.md')) as f:
     long_description = f.read()
 
+extensions = [Extension(
+    name="ark.utils._bootstrapping",
+    sources=["ark/utils/_bootstrapping/_close_cell_num_random.pyx"],
+    include_dirs=[np.get_include()]
+)]
 
 setup(
     name='ark-analysis',
@@ -20,25 +28,8 @@ setup(
     author='Angelo Lab',
     url='https://github.com/angelolab/ark-analysis',
     download_url='https://github.com/angelolab/ark-analysis/archive/v{}.tar.gz'.format(VERSION),
-    install_requires=['cryptography>=3.4.8,<4',
-                      'google-api-python-client>=2.7.0,<3',
-                      'google-auth-httplib2>=0.1.0,<1',
-                      'google-auth-oauthlib>=0.4.4,<1',
-                      'jupyter>=1.0.0,<2',
-                      'jupyter_contrib_nbextensions>=0.5.1,<1',
-                      'jupyterlab>=3.1.9,<4',
-                      'matplotlib>=2.2.2,<3',
-                      'numpy>=1.16.3,<2',
-                      'pandas>=0.23.3,<1',
-                      'requests>=2.25.1,<3',
-                      'scikit-image>=0.14.3,<=0.16.2',
-                      'scikit-learn>=0.19.1,<1',
-                      'scipy>=1.1.0,<2',
-                      'seaborn>=0.10.1,<1',
-                      'statsmodels>=0.11.1,<1',
-                      'umap-learn>=0.4.6,<1',
-                      'xarray>=0.12.3,<1',
-                      'tqdm>=4.54.1,<5'],
+    ext_modules=cythonize(extensions),
+    install_requires=requirements,
     extras_require={
         'tests': ['pytest',
                   'pytest-cov',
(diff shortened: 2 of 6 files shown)


ackagel:
The alg should be pretty much as fast as possible now, ignoring parallelization.  I've set it up so that if we end up needing that extra speed, it wouldn't be too hard do multithread stuff.

Also, the code should be relatively clean now, so restructuring comments, etc are welcome

ark/utils/spatial_analysis_utils.py
```diff
@@ -246,33 +247,43 @@ def compute_close_cell_num_random(marker_nums, dist_mat, dist_lim, bootstrap_num
...
-    # Create close_num_rand
-    close_num_rand = np.zeros((
-        marker_count, marker_count, bootstrap_num), dtype=np.uint16)
-
     # Generate binarized distance matrix
     dist_mat_bin = (dist_mat.values < dist_lim).astype(np.uint16)
 
-    # static choice array
-    choice_ar = list(range(dist_mat.shape[0]))
+    for mn in marker_nums:
```
ngreenwald, line 253: Same thing here, more comments to help future you or future us

setup.py
```diff
@@ -1,15 +1,35 @@
 import setuptools
 from os import path
-from distutils.command.build_ext import build_ext as DistUtilsBuildExt
-from setuptools import setup, find_packages
+from setuptools import setup, find_packages, Extension
```
ngreenwald, line 3: Once this gets merged in, we'll need to make a new release and double check that all of this was formatted formatted appropriately. I don't know if there's a better way to check that all of this is correct

ngreenwald (changes requested):
Did you check the outputs against the data from Alex?

setup.py
ackagel: For sure.  We should be good, since I followed the cython guides plus some other example cython projects to see what they do, but a double check on that never hurts.

Also, if we want stuff like autodoc strings in .pyx files, there's a little more overhead with a 'declaration' file and configuring some things in `ark/utils/__init__.py` which can make that happen, but that's a little extra imo

ackagel:
Z-score comparison for the test data in the notebooks:

## Channel spatial enrichment stats (`bootstrap_num=100`)

### Original code run and duplicate run
channels_original_recipe
channels_orignal_duplicate

###  Speedup version and duplicate

channels_speedup

channels_speedup_duplicate

### Notable differences

That one weird CD45-NaKATPase negative enrichment that wasn't duplicated; `bootstrap>=1000` probably makes that not a thing

## Cluster spatial enrichment stats

### Original code run and duplicate run (`bootstrap=100`)
clusters_original_duplicate
clusters_original_recipe

### Speedup version and duplicate (`bootstrap=1000`)
clusters_speedup
 
clusters_speedup_duplicate

### Notable differences

None, nada, zip, nilch, big O

ark/utils/_bootstrapping.pyx
```diff
@@ -0,0 +1,147 @@
...
+                            rand_rows, rand_cols, rand_cols_flags, num_choices, m1n, m2n,
+                            bootstrap_num)
+            close_num_rand_view[k, j, :] = close_num_rand_view[j, k, :]
+
+    PyMem_Free(rand_rows)
+    PyMem_Free(rand_cols)
+    PyMem_Free(rand_cols_flags)
+    PyMem_Free(m2n_small)
+
+    return close_num_rand
```
alex-l-kong, line 142: Along with Noah's comment above, it would be helpful to add:

- Documentation describing the arguments and return type
- What `m2n_small` is and how it determines whether to use `_list_accum` or `_dict_accum`
- Etc. for anything that might cause confusion

setup.py
```diff
@@ -1,15 +1,35 @@
...
+
+    directive_defaults['linetrace'] = True
+    directive_defaults['binding'] = True
+
+CYTHON_MACROS = [('CYTHON_TRACE', '1')] if CYTHON_DEBUG else None
 
 VERSION = '0.2.9'
 
+with open(path.join(path.dirname(__file__), 'requirements.txt')) as req_file:
+    requirements = req_file.read().splitlines()
```
alex-l-kong, line 21: Heads up, I tried this before with `setup.py` and it failed to deploy to PyPI correctly. If it fails again, a quick fix would be to revert to the whole "list out everything" method.

alex-l-kong (commented):
Just a couple more things.

ackagel:
I think this is good to be merged into the spatial analysis update branch; lmk if there's anything I'm forgetting for this pass (the context stuff is in a different PR for reference).

ark/utils/_bootstrapping.pyx
```diff
@@ -0,0 +1,276 @@
...
+
+
+@boundscheck(False) # Deactivate bounds checking
+@wraparound(False)  # Deactivate negative indexing
+@cdivision(True)    # Ignore modulo/divide by zero warning
+cdef inline void _list_accum(DTYPE_t[:] close_num_rand_view,
+                             const DTYPE_t[:, :] dist_mat_bin, MAXINDEX_t[:] pos_labels,
+                             Py_ssize_t* rand_cols, Py_ssize_t num_choices, int m1n, int m2n,
+                             int bootstrap_num):
+    """ List based accumulation for small secondary marker size
```
ngreenwald, line 62: What is secondary marker size?

ngreenwald (changes requested):
Looks great, one question about your docstring

ark/utils/_bootstrapping.pyx
ackagel: just the number of cells 'of lineage'/'with marker' `j` (i.e `m2n`).  it's faster than `dict_accum` in the case where `m2n` is smaller than the average `m2n` over all markers/lineages.  I'm open to tweaking the doc string to be more clear.

ngreenwald: Okay got it. If you're consistent with that I think it's fine to use it

ark/utils/spatial_analysis_utils.py
```diff
@@ -246,33 +244,54 @@ def compute_close_cell_num_random(marker_nums, dist_mat, dist_lim, bootstrap_num
...
+    cols_in_row = [[] for i in range(dist_mat_bin.shape[0])]
+    for pair_idx, row in enumerate(rows):
+        cols_in_row[row].append(cols[pair_idx])
+
+    # flattens list-of-list representation into 1D array and stores the index keys
+    row_indicies = [0]
+    cols_in_row_flat = []
+    for cell_indx, row in enumerate(range(dist_mat_bin.shape[0])):
+        row_indicies.append(len(cols_in_row[row]) + row_indicies[cell_indx])
+        cols_in_row_flat.extend(cols_in_row[row])
```
alex-l-kong, line 266: After `cols_in_row` is created it's then immediately used to create `cols_in_row_flat`. Any chance this can be done in one loop and not two consecutive ones?

ark/utils/_bootstrapping.pyx
```diff
@@ -0,0 +1,276 @@
...
+    cdef Py_ssize_t m1_label, m2_label
+
+    if not m1n or not m2n:
+        return
+
+    for r in range(bootstrap_num):   
+        accum = 0
+        _c_permutation(rand_cols, num_choices)
+        memset(rand_cols_flags, 0, num_choices * sizeof(UINT8_t))
+        _init_flag_table(rand_cols_flags, rand_cols, m2n)
```
alex-l-kong, line 155: This function seems pretty short, can we move the code in `_init_flag_table` here or is this something we expect to see elsewhere in `_bootstrapping.pyx` in the future (or is there something about Cython that makes this advantageous)?

alex-l-kong (commented):
Just a few minor nitpicky comments.

ark/utils/spatial_analysis_utils.py
ackagel: yep yep yep, good catch

ark/utils/_bootstrapping.pyx
ackagel: I figured that it would look somewhat out of place in that function.  In particular, I think it improves the readability for it the be a separate function.  Plus, I could see this flag-table approach being useful in other calls in the future, since it's basically the closest thing I can have to a python dictionary in cython (hash table, no hash).

ackagel: open to thoughts though

ngreenwald (approved):

PR #464, opened 2021-09-09
added dtype arg to generate_deepcell_input and create_overlay functions

awedwards:

This PR allows high level functions in the jupyter notebook to handle images that are of a different datatype than the default int16.

I added ```dtype``` as an argument for the ```data_utils.generate_deepcell_input``` and ```plot_utils.create_overlay functions```. The dtype argument is then passed to the load_imgs functions from load_utils.


ngreenwald (changes requested):
Looks great. Did you test that this solves the problem? I think it's overkill to add a test just for this as long as you've checked that both work now with this change.

awedwards:
Yes, I tested it on both int8 and int32 images and both functions work now.

PR #465, opened 2021-09-09
Spatial-LDA exploratory metrics and test utils.

bcollica:

This PR adds functions which compute various exploratory metrics for determining (1) a potential radius to use for spatial-LDA featurization and (2) an appropriate number of topics for training/inference.  In addition, a function was added which generates a typical cell table with default column names.  Some columns also have data which is randomly generated for testing purposes.

To get a rough idea of an appropriate number of topics, we leverage the speed of K-means clustering and compute diagnostic metrics on the result.  This includes standard clustering metrics like inertia and silhouette score.  It also includes more specialized metrics like the gap statistic which requires bootstrapping the data.  

For making a test cell table, standard column names and their ordering are taken from the documentation of `marker_quantification.generate_cell_table()`.  The new function is in `test_utils.py` and replaces `make_segmented_csv()`.  

**Remaining issues**

The initial build will fail because of a `ImportError` thrown by trying to import `logsumexp` from an outdated source.  The fixed have been merged into Calico's spatial_lda repo but have not been updated in pypi yet.  Because of that, some the tests have not been able to run yet.  Also the data for generating a test cell table is arbitrary and not realistic.  While this is mostly for testing functionality anyway, it may be nice to have data which is representative of an actual cell table.

### DIFF
--- a/ark/spLDA/processing.py
+++ b/ark/spLDA/processing.py
@@ -2,7 +2,7 @@ import functools
 
 import numpy as np
 import pandas as pd
-from scipy.spatial.distance import cdist, pdist
+from scipy.spatial.distance import pdist
 from sklearn.cluster import KMeans
 from sklearn.metrics import silhouette_score
 from sklearn.model_selection import train_test_split
@@ -279,7 +279,7 @@ def compute_topic_eda(features, topics, num_boots=25):
     return stats
 
 
-def fov_density(cell_table, total_pix=1024**2):
+def fov_density(cell_table, total_pix=1024 ** 2):
     """Computes cellular density metrics for each field of view to determine an appropriate
     radius for the featurization step.
 
@@ -301,7 +301,7 @@ def fov_density(cell_table, total_pix=1024**2):
     density_stats = {
         "average_area": {k: v["cell_size"].mean() for (k, v) in cell_table.items()},
         "cellular_density": {
-            k: np.sum(v["cell_size"]) / total_pix for (k,v) in cell_table.items()}
-                     }
+            k: np.sum(v["cell_size"]) / total_pix for (k, v) in cell_table.items()}
+    }
 
     return density_stats
--- a/ark/spLDA/processing.py
+++ b/ark/spLDA/processing.py
@@ -38,8 +38,10 @@ def format_cell_table(cell_table, markers=None, clusters=None):
 
     # Only keep columns relevant for spatial-LDA
     if markers is not None:
-        BASE_COLS.append(markers)
-    drop_columns = [c for c in cell_table.columns if c not in BASE_COLS]
+        keep_cols = BASE_COLS + markers
+    else:
+        keep_cols = BASE_COLS
+    drop_columns = [c for c in cell_table.columns if c not in keep_cols]
     cell_table_drop = cell_table.drop(columns=drop_columns)
 
     # Rename columns
@@ -189,7 +191,7 @@ def create_difference_matrices(cell_table, features, training=True, inference=Tr
     return matrix_dict
 
 
-def gap_stat(features, k, clust_inertia, num_boots):
+def gap_stat(features, k, clust_inertia, num_boots=25):
     """Computes the Gap-statistic for a given k-means clustering model as introduced by
     Tibshirani, Walther and Hastie (2001).
 
--- a/ark/spLDA/processing.py
+++ b/ark/spLDA/processing.py
@@ -296,10 +296,12 @@ def fov_density(cell_table, total_pix=1024 ** 2):
         by cells divided by the total number of pixels in each field of view.
 
     """
-    density_stats = {
-        "average_area": {k: v["cell_size"].mean() for (k, v) in cell_table.items()},
-        "cellular_density": {
-            k: np.sum(v["cell_size"]) / total_pix for (k, v) in cell_table.items()}
-    }
+    average_area = {}
+    cellular_density = {}
+    for i in range(1, 5):
+        average_area[i] = cell_table[i].cell_size.mean()
+        cellular_density[i] = np.sum(cell_table[i].cell_size) / total_pix
+
+    density_stats = {"average_area": average_area, "cellular_density": cellular_density}
 
     return density_stats
--- a/ark/spLDA/processing.py
+++ b/ark/spLDA/processing.py
@@ -165,8 +165,6 @@ def create_difference_matrices(cell_table, features, training=True, inference=Tr
     """
     if not training and not inference:
         raise ValueError("One or both of 'training' or 'inference' must be True")
-    if training and features["train_features"] is None:
-        raise ValueError("train_features cannot be 'None'")
 
     cell_table = {
         k: v for (k, v) in cell_table.items() if k not in ["fovs", "markers", "clusters"]
@@ -256,12 +254,10 @@ def compute_topic_eda(features, topics, num_boots=25):
     # Check inputs
     if num_boots < 25:
         raise ValueError("Number of bootstrap samples must be at least 25")
-    if min(topics) <= 2:
-        raise ValueError("Number of topics must be greater than 2")
-    if max(topics) >= features.shape[0] - 1:
-        raise ValueError("Number of topics must be less than {}".format(features.shape[0] - 1))
+    if min(topics) <= 2 or max(topics) >= features.shape[0] - 1:
+        raise ValueError("Number of topics must be in [2, %d]" % (features.shape[0] - 1))
 
-    inertias, silhouettes, gap_stats, gap_sds, pct_vars = {}, {}, {}, {}, {}
+    inertias = silhouettes = gap_stats = gap_sds = pct_vars = {}
 
     # Compute the total sum of squared pairwise distances between all observations
     total_ss = np.sum(pdist(features) ** 2) / features.shape[0]
--- a/ark/spLDA/processing.py
+++ b/ark/spLDA/processing.py
@@ -294,7 +294,7 @@ def fov_density(cell_table, total_pix=1024 ** 2):
     """
     average_area = {}
     cellular_density = {}
-    for i in range(1, 5):
+    for i in cell_table["fovs"]:
         average_area[i] = cell_table[i].cell_size.mean()
         cellular_density[i] = np.sum(cell_table[i].cell_size) / total_pix
 
--- a/ark/spLDA/processing.py
+++ b/ark/spLDA/processing.py
@@ -184,3 +184,4 @@ def create_difference_matrices(cell_table, features, training=True, inference=Tr
 
     matrix_dict = {"train_diff_mat": train_diff_mat, "inference_diff_mat": inference_diff_mat}
     return matrix_dict
+
--- a/ark/spLDA/processing.py
+++ b/ark/spLDA/processing.py
@@ -267,7 +267,7 @@ def compute_topic_eda(features, topics, num_boots=25):
         stats['inertia'][k] = cluster_fit.inertia_
         stats['silhouette'][k] = silhouette_score(features, cluster_fit.labels_, 'euclidean')
         stats['gap_stat'][k], stats['gap_sds'][k] = gap_stat(features, k, cluster_fit.inertia_,
-                                                       num_boots)
+                                                             num_boots)
         stats['percent_var_exp'][k] = (total_ss - cluster_fit.inertia_) / total_ss
 
     return stats
--- a/ark/spLDA/processing.py
+++ b/ark/spLDA/processing.py
@@ -1,12 +1,15 @@
 import functools
 
 import numpy as np
+import pandas as pd
+from scipy.spatial.distance import cdist, pdist
+from sklearn.cluster import KMeans
+from sklearn.metrics import silhouette_score
 from sklearn.model_selection import train_test_split
 from spatial_lda import featurization as ft
 
+import ark.utils.spatial_lda_utils as spu
 from ark.settings import BASE_COLS
-from ark.utils.spatial_lda_utils import check_format_cell_table_args, \
-    check_featurize_cell_table_args
 
 
 def format_cell_table(cell_table, markers=None, clusters=None):
@@ -31,7 +34,7 @@ def format_cell_table(cell_table, markers=None, clusters=None):
     """
 
     # Check function arguments
-    check_format_cell_table_args(cell_table=cell_table, markers=markers, clusters=clusters)
+    spu.check_format_cell_table_args(cell_table=cell_table, markers=markers, clusters=clusters)
 
     # Only keep columns relevant for spatial-LDA
     if markers is not None:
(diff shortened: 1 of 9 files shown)


ark/spLDA/processing_test.py
```diff
@@ -0,0 +1,46 @@
...
+    # Check that columns were retained/renamed
+    verify_in_list(
+        cols1=["x", "y", "cluster_id", "cluster", "is_index"], cols2=list(TEST_FORMAT[1].columns))
+    # Check that columns were dropped
+    assert len(TEST_CELL_TABLE.columns) > len(TEST_FORMAT[1].columns)
+
+
+def test_featurize_cell_table():
+    # Check for consistent dimensions
+    assert TEST_FEATURES["featurized_fovs"].shape[0] == TEST_CELL_TABLE.shape[0]
```
ngreenwald, line 33: Since there is logic in featurize_cell_table that changes based on input arguments, we should check this. Specifically, the featurization type. We want to make sure that passing `marker`, `avg_marker`, `cluster`, etc actually works, and also that the behavior changes appropriately.

We should also add input validation to `featurize_cell_table` to make sure only valid values are provided for this argument

ark/spLDA/processing_test.py
```diff
@@ -0,0 +1,46 @@
...
+    assert TEST_FEATURES["featurized_fovs"].shape[0] == N_CELLS
+    assert TEST_FEATURES["train_features"].shape[0] == TRAIN_CELLS
```
ngreenwald, line 35: Should also test that train_frac is being used appropriately

ark/spLDA/processing_test.py
```diff
@@ -0,0 +1,46 @@
...
+    verify_in_list(
+        fovs1=list(np.unique(TEST_CELL_TABLE[settings.FOV_ID])), fovs2=fov_list)
+    # Check that columns were retained/renamed
+    verify_in_list(
+        cols1=["x", "y", "cluster_id", "cluster", "is_index"], cols2=list(TEST_FORMAT[1].columns))
+    # Check that columns were dropped
+    assert len(TEST_CELL_TABLE.columns) > len(TEST_FORMAT[1].columns)
+
+
+def test_featurize_cell_table():
```
ngreenwald, line 31: Same thing here, I don't think it's necessary to have TEST_FEATURES as a global

ark/spLDA/processing_test.py
```diff
@@ -0,0 +1,46 @@
...
+
+def test_compute_topic_eda():
+    # at least 25 bootstrap iterations
+    with pytest.raises(ValueError, match="Number of bootstrap samples must be at least"):
+        pros.compute_topic_eda(TEST_FEATURES["featurized_fovs"], topics=[5], num_boots=20)
+    # appropriate range of topics
+    with pytest.raises(ValueError, match="Number of topics must be greater than 2"):
+        pros.compute_topic_eda(TEST_FEATURES["featurized_fovs"], topics=[2], num_boots=25)
+    with pytest.raises(ValueError, match=r"Number of topics must be less than"):
+        pros.compute_topic_eda(TEST_FEATURES["featurized_fovs"], topics=[1000], num_boots=25)
```
ngreenwald, line 46: These are great. Should also add a test that the output contains the correct items. This will ensure that the function runs from start to end without errors

ark/spLDA/processing_test.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def test_featurize_cell_table():
+    # Check for consistent dimensions
+    assert TEST_FEATURES["featurized_fovs"].shape[0] == TEST_CELL_TABLE.shape[0]
+    assert TEST_FEATURES["featurized_fovs"].shape[0] == N_CELLS
+    assert TEST_FEATURES["train_features"].shape[0] == TRAIN_CELLS
+
+
+def test_compute_topic_eda():
```
ngreenwald, line 38: Add a gap stat test function that does similar stuff to this one: make sure errors are raised appropriately, make sure that the expected output is generated.

ngreenwald (changes requested):
Awesome stuff! See below for comments

ark/spLDA/processing.py
```diff
@@ -184,3 +187,121 @@ def create_difference_matrices(cell_table, features, training=True, inference=Tr
...
+        within-cluster sum of squares.
+
+    """
+    # Calculate the range of each feature column
+    mins, maxs = features.apply(min, axis=0), features.apply(max, axis=0)
+    n, p = features.shape
+    w_kb = []
+    # Create bootstrapped reference data
+    boots = np.zeros(shape=(n * num_boots, p))
+    boot_ind = np.repeat(np.array(range(num_boots)), n)
```
ackagel, line 222: This could gobble up memory and cause a slow down if the number of cells is high/dense.  Might be worth checking the speed for ~30k spoofed cells.

ark/spLDA/processing.py
```diff
@@ -184,3 +187,121 @@ def create_difference_matrices(cell_table, features, training=True, inference=Tr
 
     matrix_dict = {"train_diff_mat": train_diff_mat, "inference_diff_mat": inference_diff_mat}
     return matrix_dict
+
+
+def gap_stat(features, k, clust_inertia, num_boots):
```
ackagel, line 192: `num_boots` should be given a default value.  I'm not sure what that value would be, but what ever you think is appropriate.

ackagel (commented):
Looks good.  Just a few notes on the bootstrapping function, and the stats container

ark/spLDA/processing.py
bcollica: Was actually going to touch base with you about your cython code and see if we can find a solution that way.  What say you?

ackagel: Depends.   Quick mini essay:

The time complexity looks like it's fixed at `O(n * r)` where `n` is the cell count and `r` is the bootstrap number, which shouldn't be too bad actually.  The issue can comes with this implementation's spatial complexity, which is currently `O(n * r * p)` where `p` is the feature-marker count.  Python can end up taking a lot more time when it's dealing with a lot of memory.  In particular, the `boots` variable for 30k cells, 100 bootstrap_num, and 30 markers is like 90 million entries, so with the default float64's, that's about 0.7 GB's of allocated memory.  That's almost a whole GB of allocated data that Python has to not throw a tantrum over.

So, if the code is slow in the 30k cell case, that's probably the reason, which means the first fix is to try implementing this with lower spatial complexity.  One version is something like this:

```
# this replaces 'bootstrapped reference data' creation
boot_array = np.zeros((n, p))

# looping looks like this
for b in range(num_boots):
    for i in range(p):
        boot_array[:, i] = np.random.uniform(low=mins[i], high=maxs[i], size=n)
    boo_clust = ...
    ...
```

while this is worse time complexity wise ( `O(n * r * p)` vs `O(n * r)` ) this removes `r` from the spatial complexity, which can end up (surprisingly) saving time.

If it's still slow after something like that, then it's cython time.

ark/spLDA/processing.py
```diff
@@ -184,3 +189,121 @@ def create_difference_matrices(cell_table, features, training=True, inference=Tr
...
+        provided.
+
+    """
+    # Check inputs
+    if num_boots < 25:
+        raise ValueError("Number of bootstrap samples must be at least 25")
+    if min(topics) <= 2:
+        raise ValueError("Number of topics must be greater than 2")
+    if max(topics) >= features.shape[0] - 1:
+        raise ValueError("Number of topics must be less than {}".format(features.shape[0] - 1))
```
alex-l-kong, line 266: I think you can condense this into one error message:

```
if min(topics) <= 2 or max(topics) >= features.shape[0] - 1:
    raise ValueError("Number of topics must be in [2, %d)" % (features.shape[0] - 1))
```

ark/spLDA/processing.py
```diff
@@ -184,3 +189,121 @@ def create_difference_matrices(cell_table, features, training=True, inference=Tr
...
+
+    inertias, silhouettes, gap_stats, gap_sds, pct_vars = {}, {}, {}, {}, {}
```
alex-l-kong, line 268: Python allows for a nice shortcut here:

```
intertias = silhouettes = gap_stats = gap_sds = pct_vars = {}
```

ark/spLDA/processing.py
```diff
@@ -184,3 +189,121 @@ def create_difference_matrices(cell_table, features, training=True, inference=Tr
...
+
+        - A dictionary containing the average cell size and the cellular density for each field
+        of view.  Cellular density is calculated by summing the total number of pixels occupied
+        by cells divided by the total number of pixels in each field of view.
+
+    """
+    density_stats = {
+        "average_area": {k: v["cell_size"].mean() for (k, v) in cell_table.items()},
+        "cellular_density": {
+            k: np.sum(v["cell_size"]) / total_pix for (k, v) in cell_table.items()}
```
alex-l-kong, line 306:
```suggestion
        "cellular_density": {
            k: np.sum(v["cell_size"]) / total_pix for (k, v) in cell_table.items()
        }
```

alex-l-kong (changes requested):
A few more comments on top of what everyone else has said.

ark/spLDA/processing.py
bcollica: Using 30k cells, 20 features, and 100 bootstrap samples, the original code took about 200 seconds and the one you suggested took about 184 seconds.

ackagel: That doesn't sound too bad then.  Since those would probably be some of our max settings, I think 180-200s is reasonable.  

That being said, if we really really want this to be lightning fast, the speed improvements from a cython-ization depend on where the 'slow' part is.  Using the `cProfile` module, you can see what function calls are taking up the most cumulative time.  If that ends up being the `np.random.uniform` initializations, then cython could speed that up quite a bit.  If it's the `Kmeans` call (that's my suspicion), then idk if there's much to be done on cython's end.

ark/spLDA/processing.py
bcollica: I ended up having to change this whole part anyway because python got mad at me

ark/spLDA/processing.py
bcollica: I always forget this python hack!

ark/spLDA/processing.py
```diff
@@ -184,3 +187,121 @@ def create_difference_matrices(cell_table, features, training=True, inference=Tr
...
+    for k in topics:
+        cluster_fit = KMeans(n_clusters=k).fit(features)
+        inertias[k] = cluster_fit.inertia_
+        silhouettes[k] = silhouette_score(features, cluster_fit.labels_, metric='euclidean')
+        gap_stats[k], gap_sds[k] = gap_stat(features, k, cluster_fit.inertia_, num_boots)
+        pct_vars[k] = (total_ss - cluster_fit.inertia_) / total_ss
+
+    stats = {"inertia": inertias, "silhouette": silhouettes,
+             "gap_stat": gap_stats, "gap_sds": gap_sds, "percent_var_exp": pct_vars}
+    return stats
```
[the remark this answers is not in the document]
bcollica: Is there a particular benefit to doing it this way?

ark/spLDA/processing.py
bcollica: Yeah most of the work is happening in the KMeans call.  I guess it's fine the way it is for now.  It's just an exploratory tool so I don't think there's much value in doing more than 100 bootstrap samples anyway.

ark/spLDA/processing.py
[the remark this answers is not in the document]
ackagel: the problem with `inertias = silhouettes = gap_stats = gap_sds = pct_vars = {}` is that they will all be the same dictionary.  Try adding a key to one of them and it'll appear in all of them.  The benefit of this method is that adding or removing stats in the future will be easier, since it's just adding/removing a member to a list, as opposed to initializing a new variable.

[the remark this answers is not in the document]
bcollica: Damn you python and you're crazy variable scope and pointers. 💪  FYI this is me shaking my fist at python, not flexing.

ark/spLDA/processing.py
```diff
@@ -184,3 +186,118 @@ def create_difference_matrices(cell_table, features, training=True, inference=Tr
...
+        dict:
+
+        - A dictionary containing the average cell size and the cellular density for each field
+        of view.  Cellular density is calculated by summing the total number of pixels occupied
+        by cells divided by the total number of pixels in each field of view.
+
+    """
+    average_area = {}
+    cellular_density = {}
+    for i in range(1, 5):
```
ngreenwald, line 297: Is there a reason this is hardcoded for 1-5?

ark/spLDA/processing_test.py
```diff
@@ -0,0 +1,154 @@
...
+def test_format_cell_table():
+    # call formatting function
+    all_clusters = list(np.unique(TEST_CELL_TABLE[settings.CLUSTER_ID]))
+    all_markers = ['A', 'B', 'C', 'D', 'E', 'F', 'G']
+    some_clusters = all_clusters[2:]
+    some_markers = all_markers[2:]
+
+    cluster_format1 = pros.format_cell_table(cell_table=TEST_CELL_TABLE, clusters=all_clusters)
+    marker_format1 = pros.format_cell_table(cell_table=TEST_CELL_TABLE, markers=all_markers)
+    cluster_format2 = pros.format_cell_table(cell_table=TEST_CELL_TABLE, clusters=some_clusters)
```
ngreenwald, line 24: Giving these more informative names will help increase readability

ark/spLDA/processing_test.py
```diff
@@ -0,0 +1,154 @@
...
+    all_clusters = list(np.unique(TEST_CELL_TABLE[settings.CLUSTER_ID]))
+    all_markers = ['A', 'B', 'C', 'D', 'E', 'F', 'G']
+    cluster_names = list(np.unique(TEST_CELL_TABLE["cluster_labels"]))
+    cell_table_format = pros.format_cell_table(cell_table=TEST_CELL_TABLE, clusters=all_clusters,
+                                               markers=all_markers)
+
+    # call featurization
+    features1 = pros.featurize_cell_table(cell_table=cell_table_format, featurization='cluster',
+                                          train_frac=0.75)
+    features2 = pros.featurize_cell_table(cell_table=cell_table_format, featurization='cluster',
```
ngreenwald, line 63: Same here for more informative names

ngreenwald (changes requested):
Nice stuff, a few more comments

ark/spLDA/processing.py
bcollica: Because I got confused and thought I was writing a test using a generated cell table 😂

alex-l-kong (changes requested):
A couple extra comments.

ark/spLDA/processing.py
```diff
@@ -31,12 +33,14 @@ def format_cell_table(cell_table, markers=None, clusters=None):
...
-    check_format_cell_table_args(cell_table=cell_table, markers=markers, clusters=clusters)
+    spu.check_format_cell_table_args(cell_table=cell_table, markers=markers, clusters=clusters)
 
     # Only keep columns relevant for spatial-LDA
     if markers is not None:
-        BASE_COLS.append(markers)
-    drop_columns = [c for c in cell_table.columns if c not in BASE_COLS]
+        keep_cols = BASE_COLS + markers
+    else:
+        keep_cols = BASE_COLS
```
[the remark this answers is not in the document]
bcollica: Sounds good

ark/spLDA/processing.py
```diff
@@ -184,3 +186,118 @@ def create_difference_matrices(cell_table, features, training=True, inference=Tr
...
+    # Calculate the range of each feature column
+    mins, maxs = features.apply(min, axis=0), features.apply(max, axis=0)
+    n, p = features.shape
+    w_kb = []
+    # Create bootstrapped reference data
+    boot_array = np.zeros((n, p))
+    # Cluster each bootstrapped sample to get the inertia
+    for b in range(num_boots):
+        for i in range(p):
+            boot_array[:, i] = np.random.uniform(low=mins[i], high=maxs[i], size=n)
```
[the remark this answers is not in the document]
bcollica: Nice, that's way easier!

PR #471, opened 2021-09-30
Allow generate_deepcell_input to accept other img_sub_folder values (including None) for non-MIBItiff inputs

### LINKED ISSUE #470: generate_deepcell_input in data_utils does not allows for img_sub_folder to be None or empty str
I have non-MIBITiff inputs for segmentation. There is no image sub-folder. Directory goes tiff_dir/fov/*tiff files* The function generate_deepcell_input in data_utils appears to force subfolder to be 'TIFs' if is_mibitiff=False and img_sub_dir = None (an empty str). Would like it to accept 'None' if sub-directory does not exist.

alex-l-kong:

Closes #470. A bug in `generate_deepcell_input` defaulted the `img_sub_folder` value passed into `load_imgs_for_tree` to always be `'TIFs'`. We need it to accept the `img_sub_folder` value passed into `generate_deepcell_input` instead.

In addition to changing this, we also change the default value for `img_sub_folder` of `generate_deepcell_input` to be `None` instead of `'TIFs'`.

### DIFF
--- a/templates/Segment_Image_Data.ipynb
+++ b/templates/Segment_Image_Data.ipynb
@@ -188,7 +188,7 @@
    "outputs": [],
    "source": [
     "# generate and save deepcell input tifs\n",
-    "# set img_sub_folder param to None if tiff_dir does not contain one for fov data\n",
+    "# set img_sub_folder param to None if the image files in tiff_dir are not in a separate sub folder \n",
     "data_utils.generate_deepcell_input(\n",
     "    deepcell_input_dir,\n",
     "    tiff_dir,\n",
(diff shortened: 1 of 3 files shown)


templates/Segment_Image_Data.ipynb
```diff
@@ -188,9 +188,16 @@
    "outputs": [],
    "source": [
     "# generate and save deepcell input tifs\n",
+    "# set img_sub_folder param to None if tiff_dir does not contain one for fov data\n",
```
ngreenwald, line 191: I would change this to:
"set img_sub_folder param to None if the image files in tiff_dir are not in a separate sub folder "

ngreenwald (changes requested):
Two small changes

ngreenwald (approved):

PR #476, opened 2021-10-13
Add QC metrics and corresponding visualizations

alex-l-kong:
**Section 1: Design details**

**Relevant background**

Images extracted directly from MIBI may not always produce the best segmentations and post-segmentation analysis (ex. spatial analysis, pixel and cell clustering, etc). Before a user runs the segmentation pipeline, it is imperative that the image data quality is not significantly skewed by noise, impurities, etc.

Manual inspection can reveal significant flaws in the images extracted, but more subtle problems may not be as easy to catch. A more detailed quantitative analysis of the images are required to get a more complete understanding. This can help users isolate problems in the MIBI extraction process, saving time hunting down errors, improving image quality (ex. using tools such as AMP), and producing better segmentations.

Tools in MATLAB and R (`MIBIevalMeanNonZeroPixelIntensityAllMarkersFromTIFs.m` and `NHPTBevalMarkersOverRun.R`) compute and visualize quality control (QC) metrics such as non-zero mean pixel intensity to assess extracted image quality. However, in line with compartmentalizing tools into ark-analysis, we will be porting these scripts over into Python. Users will be able to easily run an interactive script that will allow them to easily and quickly compute and visualize these QC metrics, giving them qualitative and quantitative evidence to move forward with segmentation or run more pre-segmentation processing.

**Design overview**

A notebook, `example_qc_comp_eval.ipynb`, will run the full QC computation and visualization process. Like our other notebooks in ARK, this can be accessed and run by starting up the Docker, so no extra installations will be needed.

While most image preprocessing is done in the MIBI extraction and AMP pipelines, the user will be allowed to specify additional Gaussian blurring as outlier noise may be difficult to remove earlier in the MIBI pipeline.

Prior to visualization, the notebook will save the results of each QC metric separately to a CSV file for reference. 

Each QC metric will be visualized as barplots with the x-axis containing the fov number and the y-axis containing the value of the QC metric, faceted by channel.

As with the other notebooks, the underlying implementation will be encapsulated away in ARK.

**Design list/flowchart**

qc_metric_roadmap (1)

**Required inputs**

The user must specify a path to a data directory of extracted MIBI images. The directory must be in 1 of 2 formats:

- Non-MIBItiff: must contain sub-folders named after the specific FOVs imaged. **Each FOV directory must have the same structure: the images must either all be in the root of the sub-folders, or have a consistent sub-directory structure.**
- MIBItiff: must contain the raw images extracted as MIBItiffs. These must be contained in the root of the data directory. **Note that it is not enough to download the raw images off the commercial instrument. The images must contain a specific encoding according to Ionpath standards**.

If the user wishes, they may explicitly specify a list of FOVs to limit the QC metric extraction to. By default, all of the FOVs in the data directory are used.

If the user wishes, they may explicitly specify a list of channels to limit the QC metric extraction to. By default, all of the channels are used. **Note that every image in the data directory must contain the same channels**!

If the user wishes, they may add Gaussian blurring to each image prior to QC metric extraction. A flag to turn on Gaussian blurring can be set in the notebook, as well as the standard deviation of the Gaussian kernel used.

**Output files**

Prior to visualization, the notebook will save the results of each QC metric separately to a CSV file for reference. As outlined by `MIBIevalMeanNonZeroPixelIntensityAllMarkersFromTIFs.m`, an example output format table will look like:

| FOV | chan1_qc_metric | chan2_qc_metric | ... | chan{n}_qc_metric |
| ---- | ----------------- | ----------------- | -- | ------------------- |
| 1 | ... | ... | ... | ... |
| 2 | ... | ... | ... | ... |
| ... | ... | ... | ... | ... |
| m | ... | ... | ... | ... |

where `m` represents the number of FOVs and `n` represents the number of channels. In this table, `chan{n}_qc_metric` defines the computed value of the specified QC metric for channel `n` in fov `m`. All QC metrics will be computed on a per-channel, per-fov basis.

Visualizations will be produced as outlined by `NHPTBevalMarkersOverRun.R`. An example visualization for total sum of all pixels:

[meanpos.pdf](https://github.com/angelolab/ark-analysis/files/7470827/meanpos.pdf)

As outlined in the design documentation, the user-facing process will be contained in example_qc_comp_eval.ipynb. This notebook will contain 3 main sections:

1. File path/hyperparameter setting: allows the user to define the following
    1. Path to extracted MIBI data
    2. If the extracted images are in MIBItiff format, and if so, what file suffix is used
    3. List of fovs to use
    4. List of channels/markers to use
    5. If additional Gaussian blurring is desired, and if so, what standard deviation to use for the Gaussian kernel. Larger standard deviations will be more aggressive in removing additional noise.
2. QC metric extraction
    1. D dataframes (D = number of QC metrics) will be extracted as separate Pandas DataFrames. Each DataFrame will have dimensions m x n (m = number of fovs, n = number of channels).
    2. The D DataFrames will be saved to separate and appropriately named .csv files. See the Design Overview section for details on the DataFrame format
3. Visualization
    1. Barplots as outlined in the Design Overview section will be visualized using seaborn

**Control flow**

_QC Metric Extraction Details_

A file, `qc_comp.py`, will contain the main MIBI QC extraction process, `compute_qc_metrics`, which will generate the DataFrames associated with each metric.

Prior to extracting the QC metrics, a Gaussian blur will be specified should the user desire it. The Gaussian blur will be applied over the entire FOV with all channels included (as opposed to each channel separately, as in the pixel clustering preprocessing case). The `scipy` package offers a convenient function: `ndimage.gaussian_filter`, which can accomplish this.

The following metrics are currently defined and computed separately for each channel per fov:

- Non-zero mean pixel intensity
- Total pixel intensity
- 99.9% pixel intensity value

Each of these metrics will have separate helper functions (appropriately named) in `qc_comp.py`. These functions will be masked to the user.

To support the addition of future metrics, `compute_qc_metrics` will return a Python dictionary with the metric name mapped to its respective DataFrame.

_Visualization Details_

The barplot visualization helper will be added directly to `qc_comp.py`, as opposed to `visualize.py`. There is a generic function that draws barplots in `visualize.py`, however, for QC metric visualization, much customization is needed that will not be applicable elsewhere. Additionally, a faceted barplot requires the use of a specific type of plot (`catplot`) to streamline customization.

**Milestones and timeline**

1. Port computation of QC metrics (`MIBIevalMeanNonZeroPixelIntensityAllMarkersFromTIFs.m`) to Python (10/22/21, DONE)
    - 3 metrics so far: non-zero mean pixel intensity, total pixel intensity, 99.9 percentile value of pixel intensity (on a per fov, per channel basis)
2. Port visualization of QC metrics (`NHPTBevalMarkersOverRun.R`) to Python (10/26/21, DONE)
3. Create a Jupyter notebook that runs the full computation and visualization of QC metrics process (10/28/21, DONE)


ark/mibi/qc_comp.py
```diff
@@ -0,0 +1,337 @@
...
+    return qc_data_batch
+
+
+def compute_qc_metrics(tiff_dir, img_sub_folder="TIFs", is_mibitiff=False,
+                       fovs=None, chans=None, batch_size=5, gaussian_blur=False,
+                       blur_factor=1, dtype='int16'):
+    """Compute the QC metric matrices
+
+    Args:
+        tiff_dir (str):
```
ngreenwald, line 159: This part will eventually need to change, but I think it's fine for now.

ngreenwald (approved):
Looks great, just want to see what Erin thinks

templates/example_qc_metric_eval.ipynb
```diff
@@ -0,0 +1,380 @@
+{
```
ngreenwald, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.    sort_function = lambda row: sorted(re.findall(r'fov(\d+)|point(\d+)', row['fov'].lower())[0])[1]
</pre><p>We're going to have all kinds of weird naming schemes for the points, we won't be able to rely on this. We'll probably need to read the JSON that was created to figure out the right order. Can this be adapted so that given a list of the weird names in the correct order, it will order them accordingly? Can put a TODO here, and do that in the next PR</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/476/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

alex-l-kong: <p>@ngreenwald we could create a map of each FOV name to a numerical index and have the user refer to that. It would make things much simpler for us downstream too: instead of working with long string names we can instead use a numerical key to identify them instead.</p>

PR #483, opened 2021-11-19
Add process to automatically download data from MIBItracker prior to QC analysis

### LINKED ISSUE #482: Add automatic MIBI image downloading from MIBItracker for QC analysis
**Section 1: Design details**

**Relevant background**

During the MIBI QC pipeline, we need a way to automate the process of downloading files users extract from MIBI off from MIBItracker. Right now, users have to manually download the files off of the MIBItracker themselves. This requires a significant amount of interruption for those running the pipeline as users have to gather the files, save it to their own device, then run it.

The way forward is to use Ionpath's API to access MIBItracker images scanned by the user. With the user some pre-defined fields from their run, an API endpoint can be created to download the data directly from running the QC pipeline (done in a Jupyter notebook) into a specified path. This removes the bottleneck of the user needing to access MIBItracker and downloading the image data directly. It also allows the data to be created for future tasks in the `ark` pipeline (ex. image segmentation, pixel clustering, etc.).

**Design overview**

This process will be put inside the QC notebook (`example_qc_eval_comp.ipynb`) prior to running QC metric analysis. Please consult the design doc for this notebook (https://github.com/angelolab/ark-analysis/pull/476) to learn more about its placement in `ark` and its structure.

At the beginning, the user will be allowed to specify the username and password to their MIBItracker account as well as their particular run information and which fovs and channels they wish to download. Using this information, we can use Ionpath's `mibilib/mibitracker` library to create an endpoint URL and make an API request to retrieve the data the user wishes to retrieve.

With this information, we can loop through each fov and channel, extract the image data as an `m x n` `numpy` array (where `m x n` is the dimension of the image scanned by MIBI), and write the data to `.tiff`. While downloading the data, the directory structure will be created in a format that `ark` can work with (see the `Output files` section for a visual example, or consult `ark/data/example_dataset` for an example in `ark`). With this in mind, the user can also specify the name of their output folder (`data_write_dir`) and an `img_sub_folder` if they wish to store the image data in a particular subdirectory structure of each folder.

As with the QC metric extraction, the vast majority of the processes (ex. validating the `fovs` and `chans`, extracting the images as `numpy` arrays, saving the data), will be abstracted away in `ark/mibi`. The notebook will remain barebones with a minimal number of function calls visible to the user.

_Note and clarification_

~~This scheme assumes the user wants to download in non-MIBItiff format. Will there be applications where users will need the images downloaded in MIBItiff format?~~

~~If yes, it is possible to do with a different function call to the `mibilib/mibitracker` library and a custom `tiff` writer provided in `mibilib/mibidata`. In this case, we'll also need to provide the user with a way to specify the extraction type (MIBItiff or not).~~

**Design list/flowchart**

qc_mibitracker_download drawio

**Required inputs**

The user will need to provide the following information:

- MIBItracker email
- MIBItracker password
- ~~Extraction type (if we choose to support both non-MIBItiff and MIBItiff downloads)~~
- Run name and run label (NOTE: these are different!)
- The name of the directory to write the image data to
- The name of the sub folder to place the image data into ~~(if so desired for non-MIBItiff extraction)~~

The run name and run label can be found on the particular run's page in Ionpath:

Screen Shot 2021-11-11 at 3 43 54 PM

**Output files**

Data for a particular `fov`/`channel` will be extracted as a `numpy.ndarray` in the dimensions extracted during the run. The data type of each image is `int`.

While saving each `numpy.ndarray` to an appropriately named `.tiff` file, we'll need to create the appropriately named directory structure as such:

Untitled Diagram drawio

~~Note that if MIBItiff extraction is supported, the `.tiff` files will always be placed in the root of the `fov` folder.~~

What an extracted image channel looks like will vary based on the specifics of the data extracted. Please consult your own MIBI-scanned data for an example of what a channel file looks like.

Because this will be run in conjunction with QC metric evaluation, the encapsulated functions needed for MIBI downloading will be placed in `qc_comp.py`. The notebook, `example_qc_metric_eval.ipynb`, will also be updated to include the MIBI download process at the beginning.

**Control flow**

_Input specification_

The user will need to provide their `email`, `password`, `run_name`, and `run_label` in appropriately named variables in `example_qc_metric_eval.ipynb`. See the Design list/flowchart section for information on how to retrieve the run name and run label from the Ionpath site.

If they wish, the user can also explicitly set a `fovs` and `chans` variable, which defines a specific subset of FOVs and channels to download. These variables default to `None`, meaning all of them will be downloaded.

Additionally, the user will need to specify `data_dir`, the location of the data folder they wish to extract the FOV image data into. If they wish, the user may explicitly set `img_sub_dir`, which creates an additional subfolder inside each FOV folder where the image data will be placed (by default, `img_sub_dir` is `None`, meaning the data will be saved in the root of each FOV folder).

_Retrieve the MIBI run information_

The notebook defers this work to a function, `download_mibitracker_data`, in `qc_comp.py` which takes in the user-specified email, password, run name, and run label.

This function first creates a MIBItracker `request_helper` instance: this is needed to call Ionpath's API to interactive with, upload, and download images. For the purposes of this project, only the latter will be needed. This `request_helper` instance will be created by calling a hidden function in `qc_comp.py`, appropriately named `create_mibitracker_request_helper`.

With the `request_helper` instance, we can retrieve the run info by calling `request_helper.search_runs` with both the run name and run label. This returns a Python `dict` that contains the name of the XML file in storage used to generate the file name, the FOV names, as well as the channels used for each FOV, etc. Let's call this `dict` `run_info`.

Before actually downloading and creating the image dataset, we can use `run_info` to locate the path of the XML run file in Ionpath storage. Using `request_helper.download_file` with the full XML run path in Ionpath storage as a param, we can retrieve this XML data as an `io.BytesIO` object, which we can then save to an XML file in the root of `data_dir` with the same name. This will be needed to determine run order and will be especially useful during the actual QC metric extraction process.

Note that the XML data needs the intermediate saving step because `io.BytesIO` cannot be converted from binary to XML otherwise.

_Create the image dataset_

Using `run_info`, we can identify the names of each FOV, their internal image ID, and their respective channels. Note that if the user explicitly specifies `fovs` and `chans`, an additional layer of verification is needed to ensure the provided FOV and channel subsets exist within the run.

When looping over each desired FOV, a specific folder for the FOV will be created inside `data_dir` (and an `img_sub_folder` if specified). The desired channel data for each FOV can be downloaded one-by-one using `request_helpers.get_channel_data`, passing both the internal image ID of the FOV and the channel name retrieved from `run_info`. This returns a `numpy` array, which can then be saved to the desired FOV folder using `skimage.io.imsave` as `{chan_name}.tiff`.

**Additional considerations**

Because downloading from MIBItracker requires heavy use of the `mibitracker` library in `mibilib` (developed by Ionpath), this library will need to be included in `requirements.txt`. We have run into `mibilib` dependency issues with our own repo, and there are two possible ways around this:

1. Resolve the dependencies: this will likely have to be on our end. The downside of this method is that there is no way to know ahead of time if these updates will break certain parts of the `ark` pipeline.
2. Copy snippets of the `mibitracker` library into our code base: a hacky but effective way around this, since `mibitracker` is just a wrapper around the `requests` library in Python in many aspects. If we go this route, these wrapper functions will also be placed in `qc_comp.py`.

Because QC metric analysis is going to be run on the commercial instrument computer, it may not be useful to create the full image dataset, directory structure and all. Alternatively, the extracted images can be stored in an `xarray.DataArray`, with dimensions specifying the `fov`, `x`, `y`, and `channel` names. This `DataArray` can be passed directly to `compute_qc_metrics` in `example_qc_metric_eval.ipynb` (as opposed to the function needing to load in the image dataset using `load_utils`). If this is the case, the current process in `example_qc_metric_eval.ipynb` will need to be updated.

**Milestones and timeline**

1. Update `qc_comp.py` with helper functions to download images from MIBItracker and create the image dataset, fully tested (2-3 days, depending on whether we'll be directly calling `mibilib` or copying their wrapper functions to our repo).
2. Update `example_qc_metric_eval.ipynb` with MIBItracker download process included prior to QC metric evaluation (1-2 days, depending on if we'll be saving the data to a directory structure or using a raw `xarray.DataArray`).

alex-l-kong:

Addresses and closes #482.

 Please read the design doc in the link to the issue for more info.


alex-l-kong:
Coverage has gone down because we do not test `mibitracker_utils`. This is just a hotfix while we wait for Ionpath to resolve their pinning issue in using their library.

ark/mibi/qc_comp.py
```diff
@@ -1,14 +1,160 @@
...
+        channels (lsit):
+            A list of channels to subset over. If `None`, uses all channels.
+    """
+
+    # verify that base_dir provided exists
+    if not os.path.exists(base_dir):
+        raise FileNotFoundError("base_dir %s does not exist" % base_dir)
+
+    # create the MIBItracker request helper
+    mr = create_mibitracker_request_helper(email, password)
```
ngreenwald, line 75: does this automatically retry if wrong pw is entered, or fails? Is the error message interpretable, or do we need to add our own?

ark/mibi/qc_comp.py
```diff
@@ -1,14 +1,160 @@
...
+
+    # get the run info using the run_name and the run_label
+    # NOTE: there will only be one entry in the 'results' key with run_name and run_label specified
+    run_info = mr.search_runs(run_name, run_label)
```
ngreenwald, line 79: Does this give an interpretable error message if no such run/label exists, or do we need to add our own?

ark/mibi/qc_comp.py
```diff
@@ -1,14 +1,160 @@
...
+            The user's MIBItracker email address
+        password (str):
+            The user's MIBItracker password
+        run_name (str):
+            The name of the run (specified on the user's MIBItracker run page)
+        run_label (str):
+            The label of the run (specified on the user's MIBItracker run page)
+        base_dir (str):
+            Where to place the created `tiff_dir`
+        tiff_dir (str):
```
ngreenwald, line 56: We should auto name tiff_dir. I think yyyy_mm_dd_run_name makes sense.

ark/mibi/qc_comp.py
```diff
@@ -1,14 +1,160 @@
...
+
+    # if the desired tiff_dir exists, remove it
+    if os.path.exists(os.path.join(base_dir, tiff_dir)):
+        rmtree(os.path.join(base_dir, tiff_dir))
+
+    # make the image directory
+    os.mkdir(os.path.join(base_dir, tiff_dir))
+
+    # write the run metadata as XML to specified xml_name in tiff_dir
+    with open(os.path.join(base_dir, tiff_dir, xml_name), 'wb') as outfile:
```
ngreenwald, line 122: What are we using the xml file for?

ark/mibi/qc_comp.py
```diff
@@ -1,14 +1,160 @@
...
+    # create the MIBItracker request helper
+    mr = create_mibitracker_request_helper(email, password)
+
+    # get the run info using the run_name and the run_label
+    # NOTE: there will only be one entry in the 'results' key with run_name and run_label specified
+    run_info = mr.search_runs(run_name, run_label)
+
+    # extract the name of the FOVs and their associated internal IDs
+    run_fov_names = [img['number'] for img in run_info['results'][0]['imageset']['images']]
+    run_fov_ids = [img['id'] for img in run_info['results'][0]['imageset']['images']]
```
ngreenwald, line 83: What is `run_fov_ids` for? Is this so that the sorting method by FOV names works to put the FOVs in correct order, or is that not implemented. I'm assuming one of these is the actual FOV number and one is the user-generated name?

ngreenwald (changes requested):
Nice start, minor comments only

ark/mibi/qc_comp.py
alex-l-kong: Ah good point, I'll add a check for this.

ark/mibi/qc_comp.py
alex-l-kong: Same as above, need to add a check for this.

ark/mibi/qc_comp.py
alex-l-kong: `xml_name` contains the run order, which we'll need to update the QC FOV sorting on the next patch.

alex-l-kong:
<p>It connects to Jaiswal's data (Mouse Heart: Aortic Root).</p>

---

 <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/483/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

alex-l-kong:
<p>That's done in a prior cell in the notebook.</p>

---

 <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/483/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ark/mibi/qc_comp.py
alex-l-kong: `run_fov_ids` is needed to call `get_channel_data`, while `run_fov_names` extracts the actual name of the FOV (ex. `Point1`) for creating the folder.

And you're right, `run_fov_ids` can be used to order the FOVs in acquisition order as well. I assumed the user would still want to see the XML file containing the acquisition order regardless, but we can easily forego that if needed.

alex-l-kong:
<p>Sounds good.</p>

---

 <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/483/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ark/mibi/qc_comp.py
```diff
@@ -1,14 +1,169 @@
...
+
+    Returns:
+        ark.mibi.mibitracker_utils.MibiRequests:
+            A request helper module instance to access a user's MIBItracker info
+    """
+
+    try:
+        return MibiRequests(settings.MIBITRACKER_BACKEND, email, password)
+    except HTTPError:
+        print("Invalid MIBItracker email or password provided")
```
ackagel, line 41: does the `HTTPError` here contain info on whether it was the email or pw that was wrong?

ark/mibi/qc_comp.py
```diff
@@ -1,14 +1,169 @@
...
+
+    # ensure all of the channels are valid (important if the user explicitly specifies channels)
+    misc_utils.verify_in_list(
+        provided_chans=channels,
+        mibitracker_run_chans=run_channels
+    )
+
+    # if the desired tiff_dir exists, remove it
+    if os.path.exists(os.path.join(base_dir, tiff_dir)):
+        rmtree(os.path.join(base_dir, tiff_dir))
```
ackagel, line 121: maybe prompt for overwrite?  or have an `overwrite_tiff_dir=False` default arg, which would error out here if `False`?

ark/mibi/qc_comp.py
```diff
@@ -1,14 +1,169 @@
...
+            chan_file = '%s.tiff' % chan
+
+            # write the data to a .tiff file in the FOV directory structure
+            imsave(
+                os.path.join(base_dir, tiff_dir, img['number'], img_sub_folder, chan_file),
+                chan_data
+            )
+
+        # append the run name and run id to the list
+        run_order.append((img['number'], img['id']))
```
ackagel, line 163: was it tested if iterating over the `run_info` will always yield fovs in the run order?

ark/mibi/qc_comp_test.py
```diff
@@ -22,6 +39,117 @@
     (['fov0', 'fov1'], ['chan0', 'chan1'], True)
 ]
 
+MIBITRACKER_EMAIL = 'qc.mibi@gmail.com'
+MIBITRACKER_PASSWORD = 'The_MIBI_Is_Down_Again1!?'
```
ackagel, line 43: *cough *cough *cough (set these as env variables on the travis server; otherwise anyone can just see this and use the qc mibitracker account)

also, let's change that pw now lol

ackagel (changes requested):
Looks largely fine to me; basically just one security thing

start_docker.sh
```diff
@@ -8,7 +8,7 @@ while test $# -gt 0
 do
   case "$1" in
     -d|--develop-notebook-templates)
-      JUPYTER_DIR='templates'
+      JUPYTER_DIR='templates_ark'
```
ngreenwald, line 11: Looks pretty minor, but did you check that these show up as expected when starting the docker?

ngreenwald (changes requested):
Looking good!

start_docker.sh
alex-l-kong: It works 👍.

ark/mibi/qc_comp_test.py
alex-l-kong: @ngreenwald can we add these to the Travis server as global environment variables?

ark/mibi/qc_comp.py
alex-l-kong: It doesn't unfortunately.

ark/mibi/qc_comp.py
alex-l-kong: The sorting scheme in the notebook accounts for the fact `fovs` will not necessarily be retrieved in the run order. This is done by keeping track of the image ID retrieved for each image, which are sorted in run order. By using this image ID as a key for sorting, the actual FOV run order can be maintained.

ark/mibi/qc_comp_test.py
alex-l-kong: @ackagel also, have you changed the `MIBITRACKER_PASSWORD`?

alex-l-kong: UPDATE: we've changed `qc.mibi@gmail.com` to be read-only and have access to just Jaiswal's project, resolving.

ngreenwald (approved):

PR #488, opened 2021-12-08
Create remapping interface for tiling script

### LINKED ISSUE #484: Update TMA tiling script to match against user-generated points

**Section 1: Design details**

**Relevant background**

The current tiling script reads in a `fov_tile_info` params dict from the user defining the start and end FOV for each region and, based on user params, generates the location of every FOV in the region based on additional user parameters such as the number of FOVs and the size of each FOV along the x and y coordinates.

However, these do not necessarily correspond to the actual desired points in question. On BeTTy, this wasn't a problem; users could easily adjust the point if it was set incorrectly. Unfortunately, the commercial instrument is not so forgiving in this regard, making it practically impossible to adjust the point after being set in the JSON params set in `example_fov_grid_generate.ipynb`. There is no way in the script currently to handle this.

With this in mind, we need an extra line of defense against the commercial instrument's inflexibility. The user will now be able to provide their own tiling parameter JSON (in a similar structure to the one returned by `create_tiled_regions`). In doing so, the user-provided JSON can then be compared against the `create_tiled_regions` JSON as a verification step prior to doing a run on the commercial instrument.

**Design overview**

This process will be appended to `example_fov_grid_generate.ipynb` after the call to `create_tiled_regions` (and subsequent saving).

We will need the user to specify an additional parameter which defines the path to the JSON they wish to verify against. This data will be read in, and both the user-proposed JSON and the automatically-generated JSON (returned by `create_tiled_regions`) will be passed into a function called `assign_closest_tiled_regions`.

Additionally, the user will need to provide a path to the region they wish to tile. This should be in the form of a `.tiff` file. This path will be used to visualize the results over (as later defined in the description of `compare_user_and_script_tiles`).

`assign_closest_tiled_regions`, in a nutshell, takes each of the user-proposed FOVs and maps them onto the closest automatically-generated FOV (in Euclidean space). This mapping contains both the name of the automatically-generated FOV each user-proposed FOV maps onto as well as information regarding each FOV.

Using this mapping returned by `assign_closest_tiled_regions`, we can then pass it to a function called `compare_user_and_script_tiles`, which visualizes on the region space the location of each user-proposed FOV compared to its mapped automatically-generated FOV. The former will be visualized as a black circle and the latter will be visualized as a blue circle. Additional coloring may be needed to adjust for different FOV mapping pairs, different levels of agreement, etc.

This visualization will also contain a mechanism to remap user-proposed FOVs to automatically-generated FOVs. This will be in the form of a drop-down menu of the user-proposed FOVs and the automatically-generated FOVs. The latter will update automatically based on the user-proposed FOVs list. It is also possible to change the mapping by selecting a different automatically-generated FOV to map to a user-proposed FOV.

**Design list/flowchart**

tma_update drawio (2)

**Required inputs**

For this section, it is assumed that the user has already ran `create_tiled_regions`.

The user will need to specify an additional proposed JSON file of tiled regions they've created. An example JSON file should look like:

```
{
    'exportDateTime': '2021-11-24T21:14:01',
    'fovFormatVersion': '1.5',
    'fovs': [
     {
           'scanCount': 1,
           'centerPointMicrons': {'x': 400, 'y': 100},
           'timingChoice': 7,
           'frameSizePixels': {'width': 2048, 'height': 2048},
           'imagingPreset': {'preset': 'Normal',
           'aperture': '2',
           'displayName': 'Fine',
           'defaults': {'timingChoice': 7}},
           'sectionId': 8201,
           'slideId': 5931,
           'name': 'row100_col400',
           'timingDescription': '1 ms'
      },
      {
          'scanCount': 1,
          'centerPointMicrons': {'x': 100, 'y': 400},
          'timingChoice': 7,
          'frameSizePixels': {'width': 2048, 'height': 2048},
          'imagingPreset': {'preset': 'Normal',
          'aperture': '2',
          'displayName': 'Fine',
          'defaults': {'timingChoice': 7}},
          'sectionId': 8201,
          'slideId': 5931,
          'name': 'row400_col100',
          'timingDescription': '1 ms'
      },
      ...
      ]
}
```

Additionally, the user will need to specify the region they will be visualizing over. This will vary greatly depending on the slide used, but it should be a `.tiff` file of the desired slide to be tiled.

**Output files**

The output will be a visualization comparing the locations of each user-proposed FOV in the region along with its mapped automatically-generated FOV. A basic overlay might look like:

tma_verification_viz drawio

As with most visualizations in `ark`, it will be possible to save this if desired.

Once you have completed section 1, please tag the relevant parties and iterate on the initial design details until everyone is satisfied. Then, 
proceed to section 2

**Control flow**

Provide additional, more granular details (if necessary) about how the proposed coding logic will be put together

**Milestones and timeline**

List each of the major components of the project, and provide an estimated completion date for each one.

alex-l-kong:

Addresses and closes #484. 

Please consult the design doc in the link above for more info.


ark/mibi/tiling_utils.py
```diff
@@ -75,28 +83,28 @@ def generate_region_info(region_params):
     return region_params_list
 
 
-def _read_tma_region_input(fov_tile_info, region_params):
-    """Reads input for TMAs from user and fov_tile_info
+def _read_tma_region_input(fov_list_info, region_params):
```
ngreenwald, line 86: Now that we're only using the automatically generated TMA to label the FOVs from the manually generated one, we don't need the rest of these user inputs, like moly QC points, fov size, etc. The only thing that matters is constructing the grid of the appropriate size.

ark/mibi/tiling_utils.py
```diff
@@ -290,7 +309,7 @@ def set_tiling_params(fov_list_path, moly_path, tma=False):
 
     # read in the fov list data
     with open(fov_list_path, 'r') as flf:
-        fov_tile_info = json.load(flf)
+        fov_list_info = json.load(flf)
```
ngreenwald, line 312: I'm not sure if we need `set_tiling_params` to support TMAs anymore

ark/mibi/tiling_utils.py
```diff
@@ -382,27 +405,27 @@ def generate_x_y_fov_pairs(x_range, y_range):
     return all_pairs
 
 
-def create_tiled_regions(tiling_params, moly_point, tma=False):
-    """Create the tiled regions for each fov
+def generate_fov_list(tiling_params, moly_point, tma=False):
```
ngreenwald, line 408: Same here, I think we can remove the TMA logic

ark/mibi/tiling_utils.py
```diff
@@ -446,22 +474,672 @@ def create_tiled_regions(tiling_params, moly_point, tma=False):
...
+            The coordinate in microns to convert
+
+    Returns:
+        tuple:
+            The converted coordinate from microns to pixels
+    """
+
+    # NOTE: all conversions are done using the fiducials
+    # convert from microns to stage coordinates
+    stage_coord_x = (coord[0] * 0.001001 - 0.3116)
```
ngreenwald, line 517: Can these hard-coded constants be turned into variables with informative names? That way if the co-registration changes in the future it will be more obvious what needs to be updated here.

ark/mibi/tiling_utils.py
```diff
@@ -446,22 +474,672 @@ def create_tiled_regions(tiling_params, moly_point, tma=False):
...
+          coordinates and size
+    """
+
+    # define the centroid and size info for manual_fovs and auto_fovs
+    manual_fovs_info = {
+        fov['name']: {
+            'centroid': convert_microns_to_pixels(
+                (fov['centerPointMicrons']['x'], fov['centerPointMicrons']['y'])
+            ),
+            'size': (fov['frameSizePixels']['width'], fov['frameSizePixels']['height'])
```
ngreenwald, line 558: Why do we need the size included here?

ark/mibi/tiling_utils.py
```diff
@@ -446,22 +474,672 @@ def create_tiled_regions(tiling_params, moly_point, tma=False):
...
+    stage_coord_y = (coord[1] * 0.001018 - 0.6294)
+
+    # convert from stage coordinates to pixels
+    pixel_coord_x = (stage_coord_x + 27.79) / 0.06887
+    pixel_coord_y = (stage_coord_y - 77.40) / -0.06926
+
+    return (int(pixel_coord_y), int(pixel_coord_x))
+
+
+def assign_closest_fovs(manual_fovs, auto_fovs, moly_point_name):
```
ngreenwald, line 527: This function can be simplified now that we're only relabeling existing TMAs. Instead of taking in a fully-formed JSON, I think it can just take in the names and centroids of the automatically generated FOVs. 

We can assume that there aren't any moly points in the manually generated TMA, and remove that logic from here

ark/mibi/tiling_utils.py
```diff
@@ -446,22 +474,672 @@ def create_tiled_regions(tiling_params, moly_point, tma=False):
...
+        (auto_fovs_info[fov]['centroid'][0], auto_fovs_info[fov]['centroid'][1]): fov
+        for fov in auto_fovs_info
+    }
+
+    # define numpy arrays of the manual and auto centroids
+    manual_centroids = np.array(
+        [list(centroid) for centroid in manual_centroid_to_fov]
+    )
+
+    # define numpy arrays of the manual and auto centroids
```
ngreenwald, line 592: redundant comment

templates_qc/example_fov_grid_generate.ipynb
```diff
@@ -1,5 +1,12 @@
 {
  "cells": [
+  {
+   "cell_type": "markdown",
```
ngreenwald, line 4: Lets rework the notebook to assume it is specifically for annotating existing TMAs, not for generating tiled images. All of the functions and arguments for options between the two should be removed. 

Organization wise, all of the filepaths should be defined at once in the top notebook. Similar to pixel clustering, let's have the default be a single identifier, like "example_TMA", and then everything named based on that: identifier + '_manual_fovs.json', identifier + '_automatic_fovs.json', identifier + '_mapping.json', etc.

ngreenwald (changes requested):
Cool, now it's time to update this to be TMA-specific

ark/mibi/tiling_utils.py
ngreenwald: Changed my mind. Cleaning up the tiling-specific functions can be for the next PR. For this one, just make sure that all of the TMA functions are updated for the new, simplified logic

ark/mibi/tiling_utils.py
alex-l-kong: Ok I agree, it was weird needing to include FOV size. When the user specifies the number of FOVs along the x- and y-axes now, we can assume the FOVs will be equally spaced along both axes according to the top-left and bottom-right FOVs defined.

ark/mibi/tiling_utils.py
alex-l-kong: Since TMAs only contain 2 FOVs, one defining the top-left FOV and one defining the top-right FOV, I think it makes sense. However, if we nuke TMA support in `set_tiling_params`, we will still need a way to generate the FOV list (the automatically-generated FOVs) without the `tiling_params` returned by this function for TMAs. See the next comment for an idea.

ark/mibi/tiling_utils.py
alex-l-kong: Since we'll need to generate a FOV list in both the TMA and non-TMA case, I think we should have separate functions for both:

- `generate_fov_list` will become `generate_fov_list_non_tma` and basically be as is without the control-flow logic for TMAs.
- Create a new function `generate_fov_list_tma` which uses the two FOVs defined to define the centroids of each FOV. This function should take in user-input defining the number of x- and y-FOVs along each axis and verify that the FOVs contained in the file are valid. Using the user-defined params, generate the list of (x, y) pairs needed and associate a FOV with each.

ark/mibi/tiling_utils.py
alex-l-kong: Yeah I think we can remove this now.

ark/mibi/tiling_utils.py
alex-l-kong: Yeah I agree, don't need to pass in the whole FOV list for either `manual_fovs` and `auto_fovs`. Nuking `set_tiling_params` and having a dedicated FOV JSON creator for TMAs will remove the need for Moly point checking here.

templates_qc/example_fov_grid_generate.ipynb
alex-l-kong: Once I'm done addressing the other comments above, I'll work on separating this into two separate notebooks.

ark/mibi/tiling_utils.py
ngreenwald: Yup, agreed. Given that there is no longer any logic about which parameter are required (it always be number of X and number of Y) probably it doesn't need to be an interactive prompt, it can just be two variables that are passed directly as arguments to the new function.

ark/mibi/tiling_utils.py
alex-l-kong: Alternatively, since `assign_closest_fovs` currently creates `manual_fovs_info` and `auto_fovs_info` which is just a `dict` mapping from each FOV to its centroid, I could just have the notebook do that and pass those in directly as arguments. Would this work or is there a reason we just want separate lists of names and centroids for both manual and auto FOVs?

ngreenwald: I think it's okay to leave the dict creation inside of the function, rather than have it in the notebook, to keep things clean. 

My comment was more related to what the automatic FOV logic needs to be. We don't need to create an entire fake JSON containing the automatically generated FOVs, since that JSON will never be saved anywhere. The only thing we need to do is create the centroids in the right spot with the right names, so that they can be used to relabel the manual JSON.

alex-l-kong:
I see, we’ll just return the automatic FOV names mapped to their centroids then when generating the automatic FOVs 👍.

alex-l-kong:
Changes:

- Major house cleaning on both non-TMA and TMA tiling processes
- Notebooks have been separated into FOV generation (which supports non-TMAs and TMAs) and remapping (just TMAs)
- Generating the automatically-generated FOVs for TMA tiles now just returns a `dict` mapping each FOV name to its centroid coordinates rather than a complete run file of FOVs
- Along the previous point, some TMA-specific functions got nuked and `assign_closest_fovs` adjusted to account for the new TMA automatically-generated FOVs format
- Fiducial mapping multiplier and offset values have been added to `settings.py`
- File paths specified at the top of the notebook with a prefix set by the user (both notebooks)
- More notebook documentation
- Change 'run' to 'region' for non-TMA tiling

ngreenwald (changes requested):
Overall looks good. For the notebooks, I think it makes more sense to have a dedicated TMA notebook, and a dedicated tiling notebook. The tiling notebook does everything needed to generating a tiled json, and the TMA notebook does everything needed for a TMA json.

ark/mibi/tiling_utils.py
```diff
@@ -382,86 +271,858 @@ def generate_x_y_fov_pairs(x_range, y_range):
...
+
+    # create all pairs between two lists
+    x_y_pairs = generate_x_y_fov_pairs(x_interval, y_interval)
+
+    # name the FOVs according to MIBI conventions
+    fov_names = ['R%dC%d' % (y + 1, x + 1) for x in range(num_fov_x) for y in range(num_fov_y)]
+
+    # get the metadata values to copy over
+    metadata_keys = list(fov_list_info.keys())
+    metadata_keys.remove('fovs')
```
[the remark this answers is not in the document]
alex-l-kong: Whoops, I forgot to push my most recent commit removing that. We don't actually want to copy these over since they're not needed.

ark/mibi/tiling_utils.py
```diff
@@ -382,86 +299,840 @@ def generate_x_y_fov_pairs(x_range, y_range):
...
-            tiled_regions['fovs'].append(moly_point)
+            fov_regions['fovs'].append(moly_point)
+
+    return fov_regions
+
+
+def tma_generate_fov_list(fov_list_path, num_fov_x, num_fov_y):
+    """Generate the list of FOVs on the image using the TMA input file in `fov_list_path`
+
+    NOTE: unlike non-TMAs, the returned list of FOVs is just an intermediate step to the
```
ngreenwald, line 408: Update docstring

ark/mibi/tiling_utils.py
```diff
@@ -382,86 +299,840 @@ def generate_x_y_fov_pairs(x_range, y_range):
...
+    start_fov_x = upper_left['centerPointMicrons']['x']
+    end_fov_x = bottom_right['centerPointMicrons']['x']
+    start_fov_y = upper_left['centerPointMicrons']['y']
+    end_fov_y = bottom_right['centerPointMicrons']['y']
+
+    # the coordinates have to be valid: upper-left cannot be below or to the right of bottom-right
+    if start_fov_x > end_fov_x:
+        err_msg = ("Coordinate error for region %s: upper-left x coordinates cannot be"
+                   " greater than bottom-right coordinates")
+        raise ValueError(err_msg % upper_left['name'])
```
ngreenwald, line 458: Nice, much simpler

ark/mibi/tiling_utils_test.py
```diff
@@ -465,125 +429,328 @@ def test_create_tiled_regions_non_tma(randomize_setting, moly_run, moly_interval
...
+    # generate the mapping from manual to automatically-generated
+    manual_to_auto_map, manual_fovs_info, auto_fovs_info = \
+        tiling_utils.assign_closest_fovs(
+            manual_sample_fovs, auto_sample_fovs
+        )
+
+    # for each manual fov, ensure the centroids are the same in manual_fovs_info
+    for fov in manual_sample_fovs['fovs']:
+        # skip the Moly points
+        if fov['name'] != 'MoQC':
```
ngreenwald, line 567: We got rid of this for the real data, but did we ever have moQC points in the test data?

ark/mibi/tiling_utils_test.py
```diff
@@ -465,125 +429,328 @@ def test_create_tiled_regions_non_tma(randomize_setting, moly_run, moly_interval
...
+    # assert the mapping is correct
+    actual_map = {
+        'row0_col25': 'row0_col0',
+        'row50_col25': 'row0_col0',
+        'row50_col50': 'row0_col100',
+        'row75_col50': 'row100_col100',
+        'row100_col25': 'row100_col0'
+    }
+
+    assert manual_to_auto_map == actual_map
```
ngreenwald, line 592: We should add an automatic FOV that isn't mapped to any manual FOV because it isn't the closest. 

We should also add two manual FOVs that map to the same automatic FOV. This indicates a problem, but in this case we want to make sure that it gets propagated forward, so it can be identified by the user.

ark/mibi/tiling_utils_test.py
```diff
@@ -465,125 +429,328 @@ def test_create_tiled_regions_non_tma(randomize_setting, moly_run, moly_interval
...
+        moly_insert, moly_interval
+    )
+
+    # assert id, name, and status are the same
+    assert new_manual_sample_fovs['id'] == manual_sample_fovs_copy['id']
+    assert new_manual_sample_fovs['name'] == manual_sample_fovs_copy['name']
+    assert new_manual_sample_fovs['status'] == manual_sample_fovs_copy['status']
+
+    # assert same number of FOVs exist in both original and scrambled (excluding Moly points)
+    assert len(manual_sample_fovs['fovs']) == \
```
ngreenwald, line 720: We can instead check that the `set` of FOV names between the two is equivalent except for Moly, right?

ngreenwald (changes requested):
Looking quite nice, a few minor comments

templates_qc/example_tma_fov_generate.ipynb
```diff
@@ -0,0 +1,430 @@
+{
```
ngreenwald, line 1: <p>This is nice. Before, I would say the notebooks had too little information. Now, they have a bit too much information. We'll find the goldilocks zone next time. The goal is to provide the user enough information to understand what they need to do, but not so much that they get overwhelmed or miss key details. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/488/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

ark/mibi/tiling_utils_test.py
alex-l-kong: We don't, and since neither `manual` nor `auto` will include any for TMAs, we can remove this test now.

ark/mibi/tiling_utils_test.py
alex-l-kong: I can add a couple extra centroid coordinates in the `auto` list for the former.

For the latter, if I'm not mistaken, I think in the mapping above both `'row0_col25'` and `'row50_col25'` maps to `'row0_col0'`. Is this the test we're looking for or something else?

ark/mibi/tiling_utils_test.py
alex-l-kong: Also change the comment to make it more clear what is being tested.

templates_qc/example_tma_fov_generate.ipynb
alex-l-kong: <p>Sounds good.</p>

ark/mibi/tiling_utils_test.py
ngreenwald: Great, sounds good. Yeah that's what I had in mind.

ark/mibi/tiling_utils_test.py
alex-l-kong: Yup, we can do this by using `misc_utils.verify_same_elements`

templates_qc/example_tma_fov_generate.ipynb
alex-l-kong: Per our discussion today, before merging in this branch @ngreenwald will push a commit with the new notebook description at the top of `example_tma_fov_generate.ipynb`.

ngreenwald (approved):

PR #490, opened 2022-01-04
Add new FOV dot visualization method

alex-l-kong:

Previously, we highlighted the names of each FOV on top of their corresponding circle for the interactive remapping visualization. This is problematic because:

1. Long FOV names may need to be displayed outside of the circle
2. If circles overlap, the FOV names will display on top of each other and become unreadable

With this in mind, we remove the text on each FOV and instead use a green circle to highlight which are selected. This works because FOVs follow a specific numbering scheme such that it's easy to resolve which one names which.

On the initial draw, change the way the circles are colored so that the highlighted pair is green.

On the dropdown-menu updates, make a similar change.


alex-l-kong:
Addressed:

1. Make interactive figure display size smaller
2. Decrease the size of the dots indicating each tile centroid
3. Change the color of the highlighted FOV pair mapping to dark red and dark blue

ngreenwald (changes requested):
The notebook doesn't run, `remap_and_reorder_tiles` fails

alex-l-kong:
@ngreenwald did you run it after you clicked `Save mapping`? The way the notebook works, the interactive visualization won't be able to stop any cells below it from running, which means some file paths may not exist yet (ex. `mapping_path`).

I don't think `ipywidgets` allows for this, so I specify in the comments for the interactive visualization to ignore any error messages in the final section (under `Use mapping to rename...`) before the user saves their mapping.

ngreenwald:
Yup, this was after saving

image

alex-l-kong:
@ngreenwald OK I see, I assumed that every JSON tile file would contain an `id`, `name`, and `status` field. I can add a check to ensure that they're only added if they exist.

alex-l-kong:
@ngreenwald can you see if the new patch works?

ngreenwald:
Yup, it works now

ark/mibi/tiling_utils.py
```diff
@@ -1150,15 +1102,36 @@ def remap_and_reorder_tiles(tiled_regions_proposed, proposed_to_auto_map,
...
         raise ValueError("moly_interval must be at least 1")
 
-    # rename the FOVs based on the mapping
+    # define a new tiled regions proposed dict
+    tiled_regions_remapped = {}
+
+    # if the user specifies additional metadata, make sure they're copied over
+    # specifically, id, name, and status
+    if 'id' in tiled_regions_proposed:
+        tiled_regions_remapped['id'] = tiled_regions_proposed['id']
```
ngreenwald, line 1111: We should have a helper function that copies over whatever the metadata is at the top of the JSON instead of hard coding it. So anything that's just a single string.

ark/mibi/tiling_utils.py
```diff
@@ -1150,15 +1102,36 @@ def remap_and_reorder_tiles(tiled_regions_proposed, proposed_to_auto_map,
...
+        # from running if a mapping_path provided already exists
+        fov_data = copy.deepcopy(fov)
+
+        # remap the name and append to tiled_regions_remapped
+        fov_data['name'] = proposed_to_auto_map[fov['name']]
+        tiled_regions_remapped['fovs'].append(fov_data)
 
     # randomize the order of the FOVs
-    tiled_regions_proposed['fovs'] = shuffle(tiled_regions_proposed['fovs'])
+    tiled_regions_remapped['fovs'] = shuffle(tiled_regions_remapped['fovs'])
```
ngreenwald, line 1131: This should be controlled by a boolean flag for whether randomization happens or not

ark/mibi/tiling_utils.py
```diff
@@ -1150,15 +1102,36 @@ def remap_and_reorder_tiles(tiled_regions_proposed, proposed_to_auto_map,
...
 
     # insert Moly points at the specified interval
-    for mi in range(moly_interval, len(tiled_regions_proposed['fovs']), moly_interval + 1):
-        tiled_regions_proposed['fovs'].insert(mi, moly_point)
+    for mi in range(moly_interval, len(tiled_regions_remapped['fovs']), moly_interval + 1):
```
ngreenwald, line 1134: Same here

ngreenwald (changes requested):
First two comments about naming apply to the entire branch, the rest are pretty minor

ark/mibi/tiling_utils.py
alex-l-kong: To clarify, does this mean it's OK to leave out `dict` fields such as `opticalImage` if specified, or should we copy those over too?

ngreenwald: Hmm, can't hurt to include those I guess. It's not needed for loading

ngreenwald (approved):
Looks good for previous changes, I'll leave comments on main PR

PR #491, opened 2022-01-07
removes self-counting and asymmetrizes randomization scheme

### LINKED ISSUE #485: Bootstrapping in the spatial enrichment analysis
Request to modify the spatial enrichment analysis (calculate_cluster_spatial_enrichment, specifically compute_close_cell_num_random. For a for a pair of cell phenotypes (i, j), the bootstrap permutations should be done by only permuting the labels of j NOT both. This way the analysis becomes the enrichment of phenotype j around i. Otherwise this analysis is roughly symmetrical (aside from minor fluctuations from the bootstrapping).

ackagel:
## **Purpose**

Corrects bootstrapping randomization to asymmetric pairing scheme.  Closes #485.

## **Implementation**

Currently building off of the speed up PR.  Most changes are to the arguments of the relevant functions.  Still runs at about the same speed, although there's still a small speed up which could be gained by assuring that the marker/cell_type undergoing dictionary storage is the larger of the two.  This was guaranteed in the symmetric version, but now requires (at a minimum) a `rand_rows_or_cols` flag to reap the full benefits of said speed-up.

## **Remaining issues**

Still need to correct the test functions' arguments

### DIFF
--- a/ark/utils/_bootstrapping.pyx
+++ b/ark/utils/_bootstrapping.pyx
@@ -56,7 +56,7 @@ cdef inline void _init_flag_table(UINT8_t* flags, const Py_ssize_t* perm,
 @wraparound(False)  # Deactivate negative indexing
 @cdivision(True)    # Ignore modulo/divide by zero warning
 cdef inline void _list_accum(DTYPE_t[:] close_num_rand_view,
-                             const DTYPE_t[:, :] dist_mat_bin, Py_ssize_t* rand_rows,
+                             const DTYPE_t[:, :] dist_mat_bin, MAXINDEX_t[:] pos_labels,
                              Py_ssize_t* rand_cols, Py_ssize_t num_choices, int m1n, int m2n,
                              int bootstrap_num):
     """ List based accumulation for small secondary marker size
@@ -73,8 +73,8 @@ cdef inline void _list_accum(DTYPE_t[:] close_num_rand_view,
             typed memory view of the close_num_rand_view datastructure
         dist_mat_bin (np.ndarray[np.uint16]):
             binarized distance matrix
-        rand_rows (Py_ssize_t*):
-            pointer to mutable row randomization memory block
+        pos_labels (np.ndarray[np.uint64]):
+            marker indidcies within dist_mat_bin    
         rand_cols (Py_ssize_t*):
             pointer to mutable column randomization memory block
         num_choices (Py_ssize_t):
@@ -90,14 +90,18 @@ cdef inline void _list_accum(DTYPE_t[:] close_num_rand_view,
     cdef DTYPE_t accum
     cdef Py_ssize_t m1_label, m2_label
 
+    #print(m1n, m2n)
+
     if not m1n or not m2n:
         return
 
+    #print(pos_labels, m1n, m2n)
     for r in range(bootstrap_num):
+        #print(r)
         accum = 0
-        _c_permutation(rand_rows, num_choices)
+        #_c_permutation(rand_rows, num_choices)
         _c_permutation(rand_cols, num_choices)
-        for m1_label in rand_rows[:m1n]:
+        for m1_label in pos_labels:
             for m2_label in rand_cols[:m2n]:
                 accum += dist_mat_bin[m1_label, m2_label]
         close_num_rand_view[r] = accum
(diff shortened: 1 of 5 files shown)


ackagel:
for some reason, fixing a set seed for the tests no longer guarantees exact consistency between batching strategies.  I'm removing the tests for exact batch consistency and will only validate that the 'close_num' counts are the same.  I could do some stats comparisons, but since the batching shouldn't effect the outcome of the stats, I'm going to ignore it for now.

ark/utils/_bootstrapping.pyx
```diff
@@ -90,14 +90,18 @@ cdef inline void _list_accum(DTYPE_t[:] close_num_rand_view,
     cdef DTYPE_t accum
     cdef Py_ssize_t m1_label, m2_label
 
+    #print(m1n, m2n)
```
ngreenwald, line 93: leftover comments here and below

ark/utils/_bootstrapping.pyx
```diff
@@ -200,10 +208,10 @@ cdef _compute_close_num_rand(DTYPE_t[:, :] dist_mat_bin, DTYPE_t[:] cols_in_row_
     cdef DTYPE_t[:, :, :] close_num_rand_view = close_num_rand
     
     # allocate marker_label randomization containers
-    cdef Py_ssize_t* rand_rows = <Py_ssize_t*> PyMem_Malloc(num_choices * sizeof(Py_ssize_t))
+    #cdef Py_ssize_t* rand_rows = <Py_ssize_t*> PyMem_Malloc(num_choices * sizeof(Py_ssize_t))
```
ngreenwald, line 211: here too

ark/utils/spatial_analysis_utils.py
```diff
@@ -248,7 +245,7 @@ def compute_close_cell_num_random(marker_nums, dist_mat, dist_lim, bootstrap_num
     """
 
     # Generate binarized distance matrix
-    dist_mat_bin = (dist_mat.values < dist_lim).astype(np.uint16)
+    dist_mat_bin = ((dist_mat.values < dist_lim) & (dist_mat.values > 0)).astype(np.uint16)
```
ngreenwald, line 248: This greater than 0 is to avoid self-counting right?

ark/utils/spatial_analysis_utils_test.py
```diff
@@ -144,22 +143,35 @@ def test_compute_close_cell_num():
         dist_mat=example_dist_mat, dist_lim=100, analysis_type="cluster",
         current_fov_data=all_data, cluster_ids=cluster_ids)
 
-    assert example_closenum[0, 0] == 16
-    assert example_closenum[1, 1] == 25
-    assert example_closenum[2, 2] == 1
+    assert example_closenum[0, 0] == 12
+    assert example_closenum[1, 1] == 20
+    assert example_closenum[2, 2] == 0
```
ngreenwald, line 148: We should add a test that checks for asymmetry

ngreenwald (changes requested):
Some minor stuff

ark/utils/spatial_analysis_utils.py
ackagel: yep yep. I've gotten word from Inna and Erin that self-counting is officially a no-no (which is strongly represented in the enrichment z-scores as well).  

As for this method, I suppose there's theoretically a case where two cells share a centroid, but I doubt that ever happens in practice.  If you think it's a concern, I can just zero the diagonals after the initial binarization.

ark/utils/spatial_analysis_utils_test.py
ackagel: the regular close-num calculation is still symmetric, it's just the randomization that has the asymmetry.

In particular, cells of lineage `i` are fixed in place while `num_cells_of_linage[j]` cells are randomly selected and checked to be within the distance limit.  Previously, the cells of lineage `i` where also randomized, which, _while informative_, is more reflective of global cell 'topography' than local lineage organization.

Regardless, I can add a check for asymmetry within the `close_num_rand` tests

ark/utils/_bootstrapping.pyx
```diff
@@ -73,8 +73,8 @@ cdef inline void _list_accum(DTYPE_t[:] close_num_rand_view,
             typed memory view of the close_num_rand_view datastructure
         dist_mat_bin (np.ndarray[np.uint16]):
             binarized distance matrix
-        rand_rows (Py_ssize_t*):
-            pointer to mutable row randomization memory block
+        pos_labels (np.ndarray[np.uint64]):
+            marker indidcies within dist_mat_bin
```
alex-l-kong, line 77:
```suggestion
            marker indices within dist_mat_bin
```

ark/utils/_bootstrapping.pyx
```diff
@@ -127,8 +126,8 @@ cdef inline void _dict_accum(DTYPE_t[:] close_num_rand_view,
             flattened list-of-lists representation of binarized distance matrix
         row_indicies (np.ndarray[np.uint64]):
             'deflattening' index array for `cols_in_row_flat`
-        rand_rows (Py_ssize_t*):
-            pointer to mutable row randomization memory block
+        pos_labels (np.ndarray[np.uint64]):
+            marker indidcies within dist_mat_bin
```
alex-l-kong, line 130:
```suggestion
            marker indices within dist_mat_bin
```

alex-l-kong (commented):
Just a couple typo fixes

ngreenwald (approved):

PR #493, opened 2022-01-18
remove hardcoded variable_sizes arg

ngreenwald:

Updates load_imgs_from_tree to support variable image sizes of arbitrary size, as specified by the user

Changes the variable_sizes boolean arg to a max_img_size integer arg specifying the length of the largest image that will be loaded. This is in contrast to hardcoding in 1024 as the largest image size, which is what the function currently does

**Remaining issues**


alex-l-kong (approved):
Looks good to me. Can't remember off the top of my head but do we have any calls to `load_imgs_from_tree` with `variable_sizes` (which should now be `max_image_size`)?

PR #497, opened 2022-02-03
Integrate interactive remapping workflow into the pixel and cell clustering pipeline

### LINKED ISSUE #496: Merge interactive reclustering into the pixel and cell clustering process
**Section 1: Design details**

**Relevant background**

When running the pixel and cell clustering process, some meta clusters may end up needing to be merged with each other. The static heatmaps and barplots we currently have in the pixel and cell clustering process don't handle that for us and require a lot of awkward visualization code.

We have an interactive remapping interface for the meta clusters that also handles the z-scored average channel expression per pixel cluster and average pixel SOM/meta cluster count per cell cluster heatmaps. This issue will focus on integrating this into both `example_pixel_clustering.ipynb` and `example_cell_clustering.ipynb`.

**Design overview**

**Adding the interactive visualization**

`example_manually_adjust_metaclusters.ipynb` contains the libraries and the call to the metaclustering process we need to add in both the pixel and cell clustering notebooks. This should require just one Jupyter notebook cell and it will be placed immediately after the respective consensus clustering processes finish (in other words, the first cell in section 3).

The only real addition other than this will be documentation. The visualization can be a bit confusing to use at first, and while `example_manually_adjust_metaclusters.ipynb` does provide some documentation, this will need to be expanded upon in the pixel and cell clustering notebooks.

**Integrating with downstream processes**

_Updating meta cluster names in all dependent files_

The interactive reclustering process saves a file containing each pixel/cell SOM mapped to its new meta cluster label (plus a 3rd column for custom names of meta clusters if the user desires, defaults to just the meta cluster label). Previous PRs ensured that this file format is compatible with the downstream processes of reassigning the meta cluster labels and recomputing aggregations on meta clusters in certain cases.

Additionally, this file format will be supported by the two static visualizations that need to remain in the cell clustering notebook: weighted channel average per cell SOM and meta cluster. It may be best to extract the creation of this heatmap to a dedicated visualization function since its pretty involved for a Jupyter notebook.

Both of the above can be resolved by replacing the dummy mapping we currently create in the notebook by programmatically interacting with the metacluster GUI interface, which we can use to easily extract the filename where the mapping is stored.

_Standardizing meta cluster colors_

In future visualizations, such as the meta cluster overlay (and, for cell clustering, the weighted channel average per cell SOM and meta cluster heatmaps), we want the meta cluster colors assigned in the interactive visualization to remain consistent. Currently, the notebooks use a default `hls` color palette to generate the meta cluster colors; this will obviously need to change.

More research will need to be done into the interactive visualization setup to extract the colors used for each metacluster.

**Design list/flowchart**

pixel_cell_clustering_diagram drawio

**Required inputs**

The only inputs needed are those specified at the beginning of the pixel and cell clustering process (the image directory, segmentation directory, subset proportion, etc. for pixel clustering, the cell table, pixel clustering file param file, etc. for cell clustering). All other intermediate files needed by interactive reclustering are created during the process and will not need to be specified by the user.

**Output files**

The interactive reclustering portion of the pipeline outputs a file in the format:

| `cluster` | `metacluster` | `mc_name` |
| 1 | 1 | meta_1 |
| 2 | 1 | meta_1 |
| ... | ... | ... |

The columns are as follows:

* `cluster`: the pixel/cell SOM cluster
* `metacluster`: the remapped pixel/cell meta cluster
* `mc_name`: the custom name for the value set in `metacluster` defined by the user during the remapping process

This table can then be passed to `apply_pixel_meta_cluster_remapping` and `apply_cell_meta_cluster_remapping` for pixel and cell clustering respectively. These will relabel the meta clusters and append the custom names to all dependent files.

Once you have completed section 1, please tag the relevant parties and iterate on the initial design details until everyone is satisfied. Then, 
proceed to section 2

**Control flow**

Provide additional, more granular details (if necessary) about how the proposed coding logic will be put together

**Milestones and timeline**

List each of the major components of the project, and provide an estimated completion date for each one.

alex-l-kong:

Closes #496. This is the final major component of pixel clustering 1.0.0: the interactive remapping addition to the pipeline.

Consult the issue page (linked above) for more details.

The one unresolved issue, integrating the interactive remapping colors across the pipeline, was addressed by using the `im_cl.cmap` property of the `MetaClusterGui` class.

**Remaining issues**

Mostly just in terms of the structure of the pipeline. I've discussed with @ngreenwald but after pixel clustering 1.0.0 is merged in, we'll work on turning the pipeline into an OOP approach.


templates/example_cell_clustering.ipynb
```diff
@@ -39,12 +39,15 @@
     "\n",
```
cliu72, line 39: <p>I don't think the note about NA/NaN/Inf is needed here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/497/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

cliu72, line 39: <p>Such a small comment, but I think line 57 in cell_consensus_cluster.R should be deleted (the next line down is probably the one you want to keep).</p><p> </p><p>Line that says:</p><pre class="ql-syntax" spellcheck="false">clusterAvgsScale &lt;- clusterAvgs[,clusterCols]
</pre><p> </p><p>That file wasn't changed in this pull request so can't comment on the line directly, but figured I would just add it here since I'm looking at it now.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/497/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

cliu72:
Looks great! Some small things. The cell clustering interactive tool looks like this. Any way to remove "pixel_meta_cluster_rename" on the text on the left for the display?

image

cliu72:
Also, when I try to generate the cell overlays, I'm getting this error. The reason is because the segmentation masks are not labelled "_feature_0.tif" in this test dataset I'm using. In the pixel clustering notebook, I specified the segmentation suffix, but I don't think there is that option in the cell clustering notebook. Perhaps add that to the json file.

image

templates/example_cell_clustering.ipynb
alex-l-kong: <p>Yup, this was a line I forgot to remove in the previous PR.</p>

alex-l-kong:
Issues addressed:

- Put `seg_suffix` from pixel clustering into the params file read in for cell clustering, pass that in as an argument to `generate_cell_cluster_mask` to allow for flexible suffix names
- Add a column trimming feature to `metacluster_from_files`, preventing the `pixel_meta_cluster_rename_` prefix from hiding the pixel cluster names in the interactive heatmap for cell clustering
- Remove extraneous comments and lines of code

cliu72:
In the pixel clustering notebook, I'm getting this error in the pixel cluster overlay section. Not sure what changed since last time, but looks like there's a problem when there's a sub folder.

image

cliu72 (commented):
Integration of the tool looks good, just one error in the pixel clustering notebook.

alex-l-kong:
@cliu72 I apologize, this was part of a patch I added as part of a request from Marte to prevent loading in `.DS_Store` files as channels, it seems the wrong version made it in. I'll patch this up.

Did you get a `.DS_Store` (or some other hidden file not a channel file) error when no `img_sub_dir` was specified?

cliu72 (commented):
Still getting this error. Should probably be os.path.join(tiff_dir, fovs[0], img_sub_folder) when there is an img_sub_folder.

image

cliu72:
> @cliu72 I apologize, this was part of a patch I added as part of a request from Marte to prevent loading in `.DS_Store` files as channels, it seems the wrong version made it in. I'll patch this up.
> 
> Did you get a `.DS_Store` (or some other hidden file not a channel file) error when no `img_sub_folder` was specified?

I'm not getting a .DS_Store error, I think .DS_Store is a mac thing.

Issue #498, opened 2022-02-04
Cannot download the code base: failed to build pandas scikit-image numba llvmlite (M1 Mac)


mirayctn:
Hi everyone,

I am trying to install the code base according to the readme file but I am having some problems.

I have an M1 chip MacBook, and I downloaded Anaconda to use python according to your instructions on 
https://ark-analysis.readthedocs.io/en/latest/_rtd/development.html

I get the ark-analysis from github, then go to its folder, and after opening docker I run: 
docker build -t ark-analysis '.'

This is where I ran into problems:

Fails at:  => ERROR [6/8] RUN pip install -r /opt/ark-analysis/requirements.txt  

with this:

executor failed running [/bin/sh -c pip install -r /opt/ark-analysis/requirements.txt]: exit code: 1

and what I see from the long error message is that some of packages fail while building wheel for them.

For example:
#9 561.5   error: command 'gcc' failed with exit status 1
#9 561.5   ----------------------------------------
#9 561.5   ERROR: Failed building wheel for pandas
#9 561.5   Running setup.py clean for pandas

#9 562.1   Building wheel for scikit-image (setup.py): started
#9 562.2   Building wheel for scikit-image (setup.py): finished with status 'error'
#9 562.2   ----------------------------------------
#9 562.2   ERROR: Failed building wheel for scikit-image

#9 691.2   ModuleNotFoundError: No module named 'numpy'
#9 691.2   ----------------------------------------
#9 691.2   ERROR: Failed building wheel for numba
#9 691.2   Running setup.py clean for numba

#9 691.3   ModuleNotFoundError: No module named 'numpy'
#9 691.3   ----------------------------------------
#9 691.3   ERROR: Failed cleaning build dir for numba
#9 691.3   Building wheel for llvmlite (setup.py): started
#9 691.5   Building wheel for llvmlite (setup.py): finished with status 'error'#9 

775.6 Failed to build pandas scikit-image numba llvmlite

When I try only  "pip install -r requirements.txt ", it downloads all the packages without any problems. This only occur during the " docker build -t ark-analysis '.'  ".

Any ideas about how to fix this?

Cheers,
Miray

ngreenwald:
Hi Miray, 
[...]

PR #499, opened 2022-02-04
Add more extensive OS and Python version testing

alex-l-kong:

Partially addresses #498. Sometimes, people will post an issue that we can't duplicate because our personal machines don't run the same operating systems. This is especially the case for Mac, which as of today includes 3 versions (Catalina, Big Sur, and Monterey) that are supported. It is important that we ensure we have a way to test our codebase on different operating systems.

Additionally, it will be useful to test our codebase on different Python versions too. This can help isolate issues when supporting multiple Python versions or when upgrading to a different Python version.

We update the `.travis.yml` file to include builds for multiple jobs and multiple Python versions. We can discuss which ones we want to include.

**Remaining issues**

- Unsure if Travis can replicate the M1-chip computing environment
- Travis doesn't support Python on Windows build environments, so these will still need to be reported directly by the user

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 os:
   - linux
   - osx
+  - windows
 
 include:
   jobs:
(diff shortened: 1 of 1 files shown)


ngreenwald:
@alex-l-kong, is this something you're still working on, or can @srivarra take over?

alex-l-kong:
@srivarra can take over

ngreenwald:
Migrating to GA instead

PR #507, opened 2022-02-24
Added exact_match to list_folders

### LINKED ISSUE #360: Need better substring matching for list_folders
**Describe the bug**

The same issues that we ran into with `list_files` will appear in `list_folders` if the user wants a more exact match. For example, if there are 2 folders: "Point1" and "Point100", if the user wants to exactly match "Point1", the current logic will match both "Point1" and "Point100".

**Expected behavior**

While not as big of an issue because of the way `list_folders` is used, just as a precaution we should duplicate the `exact_match` argument in `list_folders`.

srivarra:

This PR closes #360. It adds an `exact_match` function parameter to the `list_folders` function. In addition it adds specific tests for
the case where `exact_match` is set to True.

1. The same case from `list_files` was used.
2. Instances of using `.sort()` within `assert` statements were replaced with `sorted()`, as `.sort()` returns `None`. 

**Remaining issues**

There could be more instances of using asserts with `.sort()`
instead of `sorted()`.

For example:

```python

a = [6,5,4,3]

assert a.sort() == [3,4,5,6]
```

Would not pass as `a.sort()` returns `None`.

However:

```python

a = [6,5,4,3]

assert sorted(a) == [3,4,5,6]
```

Would pass as `sorted()` returns the sorted iterable.

### DIFF
--- a/ark/utils/io_utils_test.py
+++ b/ark/utils/io_utils_test.py
@@ -169,6 +169,7 @@ def test_extract_delimited_names():
 
 
 def test_list_folders():
+    # Tests "Fuzzy Substring Matching",`exact_match` = False
     with tempfile.TemporaryDirectory() as temp_dir:
         # set up temp_dir subdirs
         dirnames = [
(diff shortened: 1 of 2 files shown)


ngreenwald (changes requested):
The build is failing due to the linter being unhappy with your code. When you run the tests locally, make sure to add the --pycodestyle flag to pytest so that you'll see the same errors as github. You can google pep8 style guide to read up best practices

image

ark/utils/io_utils_test.py
```diff
@@ -184,13 +185,61 @@ def test_list_folders():
...
+        dirnames.sort()
+        for dirname in dirnames:
+            os.mkdir(os.path.join(temp_dir, dirname))
+
+        # add extra file
+        pathlib.Path(os.path.join(temp_dir, 'test_badfile.txt')).touch()
+
+        # test substrs is None (default)
+        get_all = iou.list_folders(temp_dir, exact_match=True)
+        print(get_all.sort())
```
alex-l-kong, line 219: Remove this print statement.

ark/utils/io_utils_test.py
```diff
@@ -184,13 +185,61 @@ def test_list_folders():
...
+        assert sorted(get_all) == dirnames
+
+        # test substrs is not list (single string)
+        get_txt = iou.list_folders(temp_dir, substrs='_txt', exact_match=True)
+        assert len(get_txt) == 0
+
+        # test substrs is list
+        get_test_and_other = iou.list_folders(temp_dir, substrs=['test_', 'other'],
+                                              exact_match=True)
+        assert len(get_test_and_other) == 0
```
alex-l-kong, line 229: I would recommend writing tests where these don't return empty lists.

alex-l-kong (changes requested):
Just a couple testing comments.

alex-l-kong (approved):
Looks good, liking the new `test_list_folders` structure with separate sections for `exact_match = False` and `exact_match = True`.

ark/utils/io_utils_test.py
```diff
@@ -184,13 +176,48 @@ def test_list_folders():
...
         # test substrs is list
-        get_test_and_other = iou.list_folders(temp_dir, substrs=['test_', 'other'])
-        assert get_test_and_other.sort() == dirnames[1:].sort()
+        get_test_and_other = iou.list_folders(
+            temp_dir, substrs=['test_', 'other'], exact_match=False
+        )
+        assert sorted(get_test_and_other) == sorted(dirnames[1:])
+
+    # Tests "Exact Substring Matching", `exact_match` = True
+    with tempfile.TemporaryDirectory() as temp_dir:
```
ngreenwald, line 193: Why do we need to re-generate the context with the temp_dir. Can't we use the same files as above?

ark/utils/io_utils_test.py
```diff
@@ -184,13 +176,48 @@ def test_list_folders():
...
+            'othertf_txt',
+            'test_csv',
+            'test_out',
+        ]
+
+        dirnames.sort()
+        for dirname in dirnames:
+            os.mkdir(os.path.join(temp_dir, dirname))
+
+        # add extra file
```
ngreenwald, line 206: None of these tests actually test the condition we care about: the substring the user specifies is contained within multiple folders, and only the folder that exactly matches the substring, not the one that contains it, is returned when exact_match=True

ngreenwald (changes requested):
Syntax looks nice, see comments below

ark/utils/io_utils_test.py
srivarra: So if we have the following folders: `test_csv`, `test_csv2`, and `test_csv3`. And the user calls

```python

list_folders(temp_dir, substrs='test_csv', exact_match=True)
```

we want it to only return the folder `test_csv`?

ngreenwald: yup exactly. Similar to the testing logic for `list_files`, which is right above.

ngreenwald (approved):

PR #513, opened 2022-03-11
Don't check contrast for skimage save function.

### LINKED ISSUE #511: Don't check for contrast when saving images
Currently, when saving images, skimage throws a warning about low contrast for basically every image. These are annoying, especially when debugging, as it makes it hard to figure out which asserts failed

Either 1) add a `check_contrast=False` flag to every instance of io.imsave, or change the default behavior globally to not check for low contrast images.

srivarra:
- add an ignore_warnings function for imageio.core.util
- added reference

Closes #511. Removes the low contrast warnings for images.

Manually added `check_contrast=False` to all instances of `skimage.io.imsave` in ark.

In addition fixed a small typo and added a line to the `.gitignore` to ignore
`.python-version` files (pyenv).

[Reference](https://stackoverflow.com/questions/52486237/how-to-turn-off-skimage-warnings)

**Remaining issues**

None at the moment.


requirements.txt
```diff
@@ -21,3 +21,4 @@ tables>=3.6.1,<4
 umap-learn>=0.4.6,<1
 xarray>=0.12.3,<1
 tqdm>=4.54.1,<5
+imageio>=2.3.0
```
ngreenwald, line 24: Do we still need this?

ngreenwald (changes requested):
Looks good. I assume the imageio import was for the initial version that was going to define a function? Or is there a reason to still pin it.

requirements.txt
srivarra: Yeah it was initially used for a function. It's not needed now.

ngreenwald (approved):
Looks good. Will you make the same change for toffy and mibi_bin_tools as well?

srivarra:
Yeah I can do that.

PR #530, opened 2022-04-29
Update pixel clustering pipeline

cliu72:

Update pixel clustering pipeline

- Set the random seed for subsetting globally (instead for each fov)
- Get 99.9% normalization values for each fov during preprocessing, average these values
- Use these values to do 99.9% normalization before training and clustering
- Read in fovs in batches during training to help with RAM issue

**Remaining issues**

Need to change test functions


alex-l-kong:
The hold up here is some of the `FlowSOM` dependencies upgraded its R requirement to `>= .4.1`, however our Docker image by default installs `4.0`, causing the image to fail to deploy properly to Dockerhub because those dependencies aren't available for install.

Configuration requires a lot more messing around with source files than I was hoping for.

ark/phenotyping/create_cell_som.R
```diff
@@ -16,6 +16,15 @@ library(arrow)
 library(data.table)
 library(FlowSOM)
 
+# a helper function for computing 99.9%
+percentile_99_9_helper <- function(x) {
+    if (quantile(as.numeric(x[x > 0]), 0.999) == 0) {
+        return(quantile(as.numeric(x[x > 0], 0.999), 1))
```
cliu72, line 22: I'm confused by this line. It seems like you are turning `x[x>0],0.999` into a numeric (parentheses seem to be in the wrong place)? If 99.9% is 0, should just return the max. So you can use either `return(quantile(x,1))` or `return(max(x))`. No need to have 0.999 in this return line.

alex-l-kong: Thanks for pointing that out, I forgot to commit that change.

alex-l-kong: I need to convert to `numeric` because `percentile_99_9_helper` reads `x` as a character type by default even if the values in `clusterCountsNorm` are numeric.

cliu72: Right, that part is fine. It's just that the parentheses for the as.numeric are around the 0.999 right now.

ark/phenotyping/create_cell_som.R
```diff
@@ -19,7 +19,7 @@ library(FlowSOM)
 # a helper function for computing 99.9%
 percentile_99_9_helper <- function(x) {
     if (quantile(as.numeric(x[x > 0]), 0.999) == 0) {
-        return(quantile(as.numeric(x[x > 0], 0.999), 1))
+        return(quantile(as.numeric(x[x > 0], 1))
```
cliu72, line 22: The parentheses for the as.numeric is still in the wrong place. I think what you meant was `quantile(as.numeric(x[x > 0]), 1)`. You can try some dummy code out like x=0:5 and see that this doesn't return what you think it does. Also, since this is essentially a max function, you don't need to do the x[x>0] part. So i think `quantile(as.numeric(x),1)` is fine.

alex-l-kong: Crap, it's been a long day.

cliu72: Haha no problem, looks good to me now.

alex-l-kong (approved):
OK, _finally_ deployment has been updated with parameters that play nicely with:

1. Updated FlowSOM requirements
2. Updated FlowSOM dependency requirements
3. R 4.2 (which many dependencies now refuse to install without, required a new installation method)
4. New C/C++ requirements (needed special configuration to play nicely with `boost`, which FlowSOM relies on)
5. Travis log errors
6. Travis timeout errors

Dockerfile
```diff
@@ -1,7 +1,34 @@
...
+RUN lsb_release -a
+
+# set up the key for adding the R repo
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 95C0FAF38DB3CCAD0C080A7BDC78B2DDEABC47B7
+
+# add the correct Linux R repo
+RUN add-apt-repository 'deb https://cloud.r-project.org/bin/linux/debian bullseye-cran40/'
+
+# ensure the previous command added the right line to the sources.list files
+RUN cat /etc/apt/sources.list
```
ngreenwald, line 19: Some of the stuff in this dockerfile looks like it's mostly for testing, and not needed now that it's working, correct?

ark/phenotyping/som_utils.py
```diff
@@ -562,15 +560,16 @@ def create_fov_pixel_data(fov, channels, img_data, seg_labels,
     pixel_mat = normalize_rows(pixel_mat, channels, seg_labels is not None)
 
     # subset the pixel matrix for training
-    pixel_mat_subset = pixel_mat.sample(frac=subset_proportion, random_state=seed)
+    pixel_mat_subset = pixel_mat.sample(frac=subset_proportion)
```
ngreenwald, line 563: Is there a reason we got rid of the random seed?

ark/phenotyping/som_utils.py
```diff
@@ -630,6 +631,12 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
     if not os.path.exists(os.path.join(base_dir, subset_dir)):
         os.mkdir(os.path.join(base_dir, subset_dir))
 
+    # create variable for storing 99.9% values
+    quant_dat = pd.DataFrame()
+
+    # set seed for subsetting
+    np.random.seed(seed)
```
ngreenwald, line 638: Does this set the seed for everything?

ngreenwald (changes requested):
Couple small things

ark/phenotyping/som_utils.py
alex-l-kong: We got rid of it because `random_state` in `pandas` selects the same rows indices for each `fov` when passed directly as an argument to `sample`. So we have to preset the seed beforehand so there's stochasticity in the row selection but the results are still replicable.

Dockerfile
alex-l-kong: I thought it would be helpful to keep these around for debugging purposes, but since we dump everything to `/dev/null` anyways (to prevent output log errors), I think we can remove them.

ark/phenotyping/som_utils.py
alex-l-kong: See above comment.

Dockerfile
ngreenwald: Okay cool, whatever you prefer

ngreenwald (approved):

PR #539, opened 2022-05-03
Verify warnings

camisowers:

Closes #539 
Adds optional warnings to verify functions, as well as the appropriate tests.

Edited `verify_in_list` and `verify_same_elements` to have new argument `warn` that defaults to False if left unspecified. This should not change the existing implementation of these functions in the rest of the repo.
Also fixed a logic error in the error message for list one in `verify_same_elements`.


ngreenwald (changes requested):
Nice catch! While you're at it, can you also make the `err_str` a bit more verbose? This is one that users are going to get a lot when they forget to provide the right variable name in a list. 

Right now, it will print something like: 
`Displaying 1 of 3  invalid value(s) provided for list channel names.`

However, this is a bit too cryptic. Something explaining that all of the values they provide are expected to be in the good list, but the following items from the bad list were not found in the good list?

Same thing for the other function

ngreenwald (approved):
Looks good

PR #542, opened 2022-05-09
Enforce stricter indexing of colormap used for pixel and cell cluster overlays

alex-l-kong:

**NOTE: this PR is high priority and should be done by Thursday at latest**

The current scheme of defining the color bar for the pixel and cell cluster overlays (`plot_pixel_cell_cluster_overlay`) is not robust to heavy remapping which scrambles the indices of the colors significantly. In certain situations, it will cause the color map indices to become misaligned, mapping colors to wrong meta cluster names.

This should not happen even if the user does significant remapping and renaming.

Fixing this requires explicitly setting an index value for each metacluster that corresponds to their position in `mc_colors`, which defines the colors to use. We can use the index defined in `metacluster_id_to_name` as a baseline.

**Remaining issues**

This issue has only been significantly tested on Marte's dataset. Please post an issue if a different indexing issue pops up for a different dataset.


cliu72:
Probably didn't spend as much time as Marte, but tested it a bit and it looks good to me. I think we're good to merge this version in.

cliu72 (commented):
Probably didn't spend as much time as Marte, but tested it a bit and it looks good to me. I think we're good to merge this version in.

cliu72 (approved):

PR #543, opened 2022-05-11
raise error if wrong channel is specified: load_imgs_from_tree

### LINKED ISSUE #517: load_imgs_from_tree doesn't raise error if wrong channel name supplied

**Describe the bug**
The `load_imgs_from_tree` function allows the user to specify a list of channels to load in. However, if the user specifies a channel that doesn't exist, the channel is silently skipped, rather than raising an error

**To Reproduce**
1. Provide a non-existent channel to the `channels` argument of the function

**Expected behavior**
The function should raise an error specifying the invalid channel

srivarra:

Closes #517. Channels that are provided are silently skipped if they do not exist. If an `fov` name is provided instead of a list, the `fov` name is iterated on.

Added a check to see if `fovs` is a string, if so make it a list of `fovs`.
In order to counter the silently skipping issue, used `misc_utils.verify_same_elements` on the input channels, and the channels found with respect to the input channels.

**Remaining issues**

None atm.

### DIFF
--- a/ark/utils/load_utils.py
+++ b/ark/utils/load_utils.py
@@ -6,7 +6,7 @@ import numpy as np
 import xarray as xr
 
 from ark.utils.tiff_utils import read_mibitiff
-from ark.utils import io_utils as iou
+from ark.utils import io_utils as iou, misc_utils
 from ark.utils.google_drive_utils import path_join
 
 
@@ -97,8 +97,8 @@ def load_imgs_from_tree(data_dir, img_sub_folder=None, fovs=None, channels=None,
             directory containing folders of images
         img_sub_folder (str):
             optional name of image sub-folder within each fov
-        fovs (list):
-            optional list of folders to load imgs from. Default loads all folders
+        fovs (str, list):
+            optional list of folders to load imgs from, or the name of a single folder. Default loads all folders
         channels (list):
             optional list of imgs to load, otherwise loads all imgs
         dtype (str/type):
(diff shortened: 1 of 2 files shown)


ark/utils/load_utils.py
```diff
@@ -147,6 +152,11 @@ def load_imgs_from_tree(data_dir, img_sub_folder=None, fovs=None, channels=None,
         # get the corresponding indices found in channels_no_delim
         channels_indices = [channels_no_delim.index(chan.split('.')[0]) for chan in all_channels]
 
+        # verify if channels from user input are present in `all_channels`
+        all_channels_no_delim = [channel.split('.')[0] for channel in all_channels]
+        misc_utils.verify_same_elements(channels=all_channels_no_delim,
```
ngreenwald, line 157: These argument names are accurate, but won't help the user understand what they did wrong. When the error message gets raised, we want someone who hasn't looked inside the function to be able to figure out what the problem is

ngreenwald (changes requested):
Nice, just one comment about naming

ark/utils/load_utils.py
```diff
@@ -147,6 +152,15 @@ def load_imgs_from_tree(data_dir, img_sub_folder=None, fovs=None, channels=None,
         # get the corresponding indices found in channels_no_delim
         channels_indices = [channels_no_delim.index(chan.split('.')[0]) for chan in all_channels]
 
+        # verify if channels from user input are present in `all_channels`
+        all_channels_no_delim = [channel.split('.')[0] for channel in all_channels]
+        try:
+            misc_utils.verify_same_elements(all_channels_in_folder=all_channels_no_delim,
+                                            all_channels_detected=channels_no_delim)
+        except ValueError as _:
```
ngreenwald, line 160: I don't think we need to catch the error, just specifying the names like you did above is good

ngreenwald (changes requested):

ark/utils/load_utils.py
srivarra: So just changing the variable names in verify_same_elements is alright?

ngreenwald: yup exaclty

ngreenwald (approved):

PR #546, opened 2022-05-16
Speed up generation of pixel cluster masks overlay

### LINKED ISSUE #535: generate_pixel_cluster_mask is really slow
This function is extremely slow.

We should look and see if using alternative indexing, similar to what regionprops does, can lead to a performance speedup

alex-l-kong:

Closes #535. `generate_pixel_cluster_masks` in `data_utils.py` runs very slowly and needs to be optimized.

The main issue is in the way the mask for each FOV is currently indexed. Currently, each FOV image is accessed in its original 2D format. Due to how 2D arrays are stored in memory, this leads to several cache misses that adds a significant bottleneck.

By flattening the array using `.ravel` and converting the coordinates so we can do 1D indexing, significant time is saved as 1D arrays are stored contiguously in memory. We get about a 3x speedup per FOV.

**Remaining issues**

If this implementation is still too slow, we should examine a different option. Unfortunately, the per-FOV iteration is a bottleneck we can't eliminate.


alex-l-kong:
@cliu72 would you mind testing this PR too for speed purposes? It may not be completely accurate yet, but I'd like to get a sense of the speed performance on a different machine.

cliu72:
For 12 1024x1024 FOVs:
Old version: 6.987326205000045
This version:  2.889587150999887

ngreenwald:
Is that on the cluster or your laptop?

cliu72:
Google cloud, but baby instance (4 CPU, 32 GB memory)

ngreenwald:
I found the problem: the example notebook creates masks for all fovs, not just the pixel_fovs. I thought it was taking forever on the 8 I was analyzing, but it was using all 30 in the directory 
image

cliu72:
Didn't run into problems in my tests

alex-l-kong:
@ngreenwald from benchmarking, we get a far more significant speedup using 1-D indexing.

Here are the times for each fov using tuple-based indexing on a 2-D array:

Screen Shot 2022-05-18 at 3 15 06 PM

These are the equivalent times for 1-D indexing into a flattened array, then reshaping back into a 2-D array:

Screen Shot 2022-05-18 at 3 15 43 PM

ngreenwald (approved):

PR #565, opened 2022-05-25
Address metacluster GUI height and width issues

### LINKED ISSUE #536: Cluster labels overlapping for cell GUI

**Describe the bug**
The labels on the left side of the GUI are hard to see because of how they overlap each other

**Expected behavior**
The labels on the right side, which say the same thing, don't have this problem.

**Screenshots**
image

### LINKED ISSUE #550: Width of metacluster heatmap

**Describe the bug**
The metacluster heatmap width seems to be hardcoded to a value that doesn't work well with a large number of metaclusters. This makes the metacluster names impossible to read

**Screenshots**
image

alex-l-kong:

Closes #536 and closes #550. There are issues with the current hard-coded width and height settings for the visualization that cause the channel names to overlap and/ or the metacluster heatmap to become small and illegible. For now, we hard code new values. Depending on how we move forward, we may also add a programmatic way to compute these.

The issue lies with:

```
width_ratios = [
    int(self.mcd.cluster_count / 7),
    self.mcd.cluster_count,
    self.mcd.metacluster_count,
]
height_ratios = [6, self.mcd.marker_count, 3, 3]
subplots = plt.subplots(
    4, 3,
    gridspec_kw={
    'width_ratios': width_ratios,
    'height_ratios': height_ratios},
    figsize=(self.width, 6),
)
```

We double the width of `self.mcd.metacluster_count`, double the height of `self.mcd.marker_count`, and double the height of the entire figure in the `figsize` param of `plt.subplots`.

### DIFF
--- a/ark/utils/metacluster_remap_gui/metaclustergui.py
+++ b/ark/utils/metacluster_remap_gui/metaclustergui.py
@@ -94,16 +94,16 @@ class MetaClusterGui():
         width_ratios = [
             int(self.mcd.cluster_count / 7),
             self.mcd.cluster_count,
-            self.mcd.metacluster_count,
+            self.mcd.metacluster_count * 2,
             ]
-        height_ratios = [6, self.mcd.marker_count, 1, 1]
+        height_ratios = [12, self.mcd.marker_count * 2, 3, 3]
 
         subplots = plt.subplots(
             4, 3,
             gridspec_kw={
                 'width_ratios': width_ratios,
                 'height_ratios': height_ratios},
-            figsize=(self.width, 6),
+            figsize=(self.width, 12),
             )
         with self.plot_output:
             plt.show()
--- a/ark/utils/metacluster_remap_gui/metaclustergui.py
+++ b/ark/utils/metacluster_remap_gui/metaclustergui.py
@@ -96,8 +96,10 @@ class MetaClusterGui():
             self.mcd.cluster_count,
             self.mcd.metacluster_count * 2,
         ]
-        marker_ratio = self.mcd.marker_count / 20
-        height_ratios = [6, self.mcd.marker_count * marker_ratio, marker_ratio, marker_ratio]
+        marker_ratio = max(self.mcd.marker_count / 20, 1)
+        height_ratios = [
+            6 * marker_ratio, self.mcd.marker_count * marker_ratio, marker_ratio, marker_ratio
+        ]
 
         subplots = plt.subplots(
             4, 3,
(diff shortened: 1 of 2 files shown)


alex-l-kong:
With the current parameters, this is how a "normal" visualization might look (Candace's test dataset)

Screen Shot 2022-05-25 at 4 47 02 PM

and how a more "involved" one (lots of channels, few metaclusters, tested on Noah's dataset) might look:

Screen Shot 2022-05-25 at 4 47 33 PM

cliu72 (commented):
I think it would be nice if the "Selected" and "Metacluster" colorbar at the bottom weren't so large (similar in height to what it was before) - how hard is that to do? If it's very involved, I don't think it's worth it and it's fine as is.

I agree it would be nice in the future to have some sort of dynamic scaling, but that's not high priority at all.

alex-l-kong:
@cliu72 I can try shortening the fixed height parameters for the selection bar, however, one potential problem is it will become too short when a dataset like Noah's (with a lot more channels) is used. I noticed this happening with the old height and width parameters with Noah's dataset too.

If you're OK with the selection bar becoming small for high-channel datasets, I can go ahead with the change suggested. Otherwise, if we want the selection bar to maintain a fairly consistent height regardless of the dataset passed in, we'll need to use a more dynamic calculation that would be best saved for a different issue.

cliu72:
Hmm I think it's okay as is then, with the large bars. Would dynamically changing the height actually be that difficult? I would assume it's just number of markers * some constant.

alex-l-kong:
@cliu72 I can play around with it a bit over the weekend to see how it turns out (a callback to my CSS classes again lol).

alex-l-kong:
> Hmm I think it's okay as is then, with the large bars. Would dynamically changing the height actually be that difficult? I would assume it's just number of markers * some constant.

@cliu72 just ran a quick test, turns out adding a multiplier based on 20 (the default number of metaclusters) worked surprisingly well. Here's the base of your dataset:

Screen Shot 2022-05-27 at 3 09 16 PM

and here's Noah's dataset with more channels (top is cut off a bit):

Screen Shot 2022-05-27 at 3 10 33 PM

alex-l-kong:
One thing to note, though, is that the visualization gets a bit stretched out when the number of channels is very low. I think it's fine because I haven't seen a use case of fewer than 15 or so channels. But just to really push the lower bounds, here's what 10 channels looks like:

Screen Shot 2022-05-27 at 3 58 36 PM

And here's what 7 channels looks like

Screen Shot 2022-05-27 at 4 06 58 PM

PR #567, opened 2022-05-26
Append cell meta cluster labels to cell table

### LINKED ISSUE #523: Defaults for location of cell_table
The current cell clustering notebook requires the user to copy the cell table into the top-level directory. 

Since most of the people using this notebook will have created the cell table using our pipeline, we can infer the location of both the cell table and the segmentation masks from the top-level segmentation directory. These can be modifiable in case the user didn't use our approach, but setting the default to segmentation_dir/cell_table_normalized and segmentation_dir/deepcell_output I think would make sense.

### LINKED ISSUE #551: Update cell table with cluster IDs
The current clustering pipeline produces a feather file with the pixel metacluster weights and cell metacluster assignments for each cell. However, this information is never added to the cell table generated from segmentation. 

A final step that updates the cell table with cell label generated during the clustering process so that these labels can actually be used for downstream analysis

alex-l-kong:

Closes #551 and closes #523. The cell table is a requirement for several downstream analysis scripts, not just pixel clustering. To ensure analysis can be performed based on the cell meta cluster labels, these will need to be appended to the cell table as well.

A function `add_consensus_labels_cell_table` will run this appending process, with the call in `example_cell_clustering.ipynb` placed at the end of the notebook. The original `cell_table_name` will be overwritten with the same data including the `'cell_meta_cluster_rename'` column in `cell_consensus_name` (simply `'cell_meta_cluster'` in the cell table).

Users will be able to save multiple times and can also re-run the cell SOM training with the newly-saved cell table if desired.

Additionally, we now allow more flexible locations for the cell table. Initially, it was forced to be in `base_dir`. Now, we default the location to `segmentation_dir`, with options to change the root. Note that the updated cell table will be saved in the same root directory as the original cell table.

### DIFF
--- a/ark/phenotyping/som_utils.py
+++ b/ark/phenotyping/som_utils.py
@@ -1200,7 +1200,7 @@ def train_cell_som(fovs, channels, base_dir, pixel_consensus_dir, cell_table_pat
         pixel_consensus_dir (str):
             Name of directory with the pixel-level consensus data (SOM and meta labels added)
             Created by `pixel_consensus_cluster`
-        cell_table path (str):
+        cell_table_path (str):
             Path of the cell table, needs to be created with `Segment_Image_Data.ipynb`
         cluster_counts_name (str):
             Name of the file to save the number of pixel SOM/meta cluster counts for each cell
--- a/ark/phenotyping/som_utils.py
+++ b/ark/phenotyping/som_utils.py
@@ -1177,7 +1177,7 @@ def apply_pixel_meta_cluster_remapping(fovs, channels, base_dir,
     pixel_channel_avg_som_cluster.to_csv(som_cluster_avg_path, index=False)
 
 
-def train_cell_som(fovs, channels, base_dir, pixel_consensus_dir, cell_table_name,
+def train_cell_som(fovs, channels, base_dir, pixel_consensus_dir, cell_table_path,
                    cluster_counts_name='cluster_counts.feather',
                    cluster_counts_norm_name='cluster_counts_norm.feather',
                    pixel_cluster_col='pixel_meta_cluster_rename',
@@ -1200,8 +1200,8 @@ def train_cell_som(fovs, channels, base_dir, pixel_consensus_dir, cell_table_nam
         pixel_consensus_dir (str):
             Name of directory with the pixel-level consensus data (SOM and meta labels added)
             Created by `pixel_consensus_cluster`
-        cell_table_name (str):
-            Name of the cell table, needs to be created with `Segment_Image_Data.ipynb`
+        cell_table path (str):
+            Path of the cell table, needs to be created with `Segment_Image_Data.ipynb`
         cluster_counts_name (str):
             Name of the file to save the number of pixel SOM/meta cluster counts for each cell
         cluster_counts_norm_name (str):
@@ -1233,7 +1233,6 @@ def train_cell_som(fovs, channels, base_dir, pixel_consensus_dir, cell_table_nam
     """
 
     # define the data paths
-    cell_table_path = os.path.join(base_dir, cell_table_name)
     consensus_path = os.path.join(base_dir, pixel_consensus_dir)
     cluster_counts_path = os.path.join(base_dir, cluster_counts_name)
     cluster_counts_norm_path = os.path.join(base_dir, cluster_counts_norm_name)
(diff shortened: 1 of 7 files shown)


cliu72 (approved):
Looks good to me

cliu72:
Sorry I already approved the changes, but I just thought of a possible use case that might cause an error with this implementation. For very small cells (or weird cells), it's possible they're not included in the cell clustering (if there are no pixel clusters in there). In that case, the cell table and the cell clustering output wouldn't have the same entries. It would probably best to have those say "Unassigned" or something like that in the metacluster column in the cell table.

ark/phenotyping/som_utils.py
```diff
@@ -1935,3 +1935,58 @@ def generate_weighted_channel_avg_heatmap(cell_cluster_channel_avg_path, cell_cl
...
+        cell_table_labels=cell_table['label'].values,
+        consensus_data_labels=consensus_data['segmentation_label'].values,
+        enforce_order=True
+    )
+
+    # append the consensus SOM cluster values to the cell table
+    cell_table['cell_meta_cluster'] = consensus_data['cell_meta_cluster_rename'].copy()
+
+    # resave cell table with new meta cluster column
+    cell_table.to_csv(cell_table_path, index=False)
```
ngreenwald, line 1992: Rather than overwriting the existing cell table, I think it would be better to save an updated version

ark/phenotyping/som_utils.py
```diff
@@ -1935,3 +1935,58 @@ def generate_weighted_channel_avg_heatmap(cell_cluster_channel_avg_path, cell_cl
...
     )
+
+
+def add_consensus_labels_cell_table(base_dir, cell_table_name, cell_consensus_name):
+    """Adds the consensus cluster labels to the cell table,
+    then resaves data to `base_dir/cell_table_name`
+
+
+    Args:
+        base_dir (str):
```
ngreenwald, line 1946: I think this is a good place to tackle the cell table location issue as well:
https://github.com/angelolab/ark-analysis/issues/523

ark/phenotyping/som_utils.py
```diff
@@ -1935,3 +1935,58 @@ def generate_weighted_channel_avg_heatmap(cell_cluster_channel_avg_path, cell_cl
...
+    if not os.path.exists(cell_consensus_path):
+        raise FileNotFoundError('Cell consensus file %s does not exist in base_dir %s' %
+                                (cell_consensus_name, base_dir))
+
+    # read in the data, ensure sorted by FOV column just in case
+    cell_table = pd.read_csv(cell_table_path)
+    consensus_data = feather.read_dataframe(cell_consensus_path)
+
+    # ensure the data are sorted by fov and segmentation_label for consistency
+    cell_table = cell_table.sort_values(by=['fov', 'label'])
```
ngreenwald, line 1972: To address Candace's comment, we can pd.merge the two dataframes and raise a warning based on the output shapes

ngreenwald (changes requested):

ark/phenotyping/som_utils.py
alex-l-kong: Currently, I'm checking for the same FOV and segmentation_label values in both `cell_table` and `consensus_data`. In the case `cell_table` contains cells that don't appear in `consensus_data`, we can no longer use `misc_utils.verify_in_list` since per @cliu72 having different sets of segmentation labels between the two tables might actually not indicate an error.

I'm not usually a fan of leaving it up to the user to verify the data themselves, but in this case, I think it's probably the easiest option. This will allow us to just run `pd.merge` and label anything it can't find in `cell_table` as `'Unassigned'`. If the user winds up with a ton of `'Unassigned'` values, it should be clear to them that they passed in the wrong cell table and/or consensus-labeled data.

Any thoughts?

ark/phenotyping/som_utils.py
cliu72: I think the second option is fine (maybe append like '_phenotypes.csv' or '_cell_labels.csv' to the end).

ark/phenotyping/som_utils.py
cliu72: I think just having cells as "Unassigned" is fine. I don't think there's a need to raise a warning. If they passed in the wrong cell table, that would cause problems upstream anyways since the cell sizes in the cell table are used during the cell clustering. So I think using pd.merge and label the other stuff as "Unassigned" is the way to go.

ngreenwald (approved):

PR #570, opened 2022-05-27
Update FlowSOM pipeline with updated `MapDataToCodes` optional flag in `SOM` function

### LINKED ISSUE #568: Update SOM training to return only the SOM weights
The current SOM function within FlowSOM both trains the SOM, and applies the trained SOM to the provided training data. However, this doubles the memory requirements of the function. Furthermore, we don't care about running the SOM on the provided data, since it is a subset of the whole dataset, which we batch process after the fact

We should add an option to the SOM function to skip the [MapDataToCodes](https://github.com/SofieVG/FlowSOM/blob/31bf74c131e0c1bee2584a060e99ceff8b74b5af/R/2_buildSOM.R#L228). The easiest option would be getting this merged in to the main FlowSOM repo. Given that it's only a couple lines change, I think there's a decent chance it will get merged in. We should open a PR explaining our use case, why this is helpful, etc. 

However, in the interim, I think we can create a fork and make the changes ourselves, and then update the imports.

alex-l-kong:

Closes #568. This addresses the issue caused by the current functionality of `SOM` in the `FlowSOM` library which causes `MapDataToCodes` to run by default.

1. We've created a fork of the main `FlowSOM` repo and added an optional flag `map` to the `SOM` function which allows us to prevent the call to `MapDataToCodes`. This parameter is also added to our `SOM` calls in `create_pixel_som.R` and `create_cell_som.R`.
2. For the time being, we add an explicit command to overwrite the default `FlowSOM` installation with our forked `FlowSOM` package. This will be removed whenever the devs of the main `FlowSOM` repo review the PR I've opened addressing the `map` parameter.

**Remaining issues**

1. It is possible there may be other memory-related issues we'll run into that will need further addressing.
2. It is possible the main `FlowSOM` devs do not approve our PR. In that case, we'll have to continue piggybacking on the forked `FlowSOM` repo for the foreseeable future. That might be a good idea considering we may want to make more optimizations to the C code and what not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ jobs:
         on:
           tags: true
     - stage: docker_deploy
-      if: tag IS present
+      if: branch = flex_som_map
       python: 3.6
       script:
         - "travis_wait 120 sleep 7200 &"
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ jobs:
         on:
           tags: true
     - stage: docker_deploy
-      if: branch = flex_som_map
+      if: tag IS present
       python: 3.6
       script:
         - "travis_wait 120 sleep 7200 &"
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,8 @@ RUN R -e "install.packages('arrow')"
 RUN R -e "install.packages('data.table')"
 RUN R -e "install.packages('BiocManager')"
 RUN R -e "BiocManager::install('FlowSOM')"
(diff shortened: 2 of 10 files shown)


ngreenwald:
Not working for me: 
image

ngreenwald:
After installing devtools it's working. If this resolves the first half of the memory issues we should move this logic to the docker

ngreenwald:
Yup, resolves memory issues

ngreenwald (approved):
Nice, looks good

PR #577, opened 2022-05-31
`remove_file_extensions` is more robust to various fov naming schemes

### LINKED ISSUE #537: `create_deepcell_output` fails if FOV names with `.` specified
**Describe the bug**

If the user specifies a list of FOVs with a `.` in them (ex. `fov.1` or `fov.2`), `create_deepcell_output` will fail to parse the FOV names correctly.

This is because of these lines:

```
# set fovs equal to input_files it not already set
if fovs is None:
    fovs = input_files

# now extract only the names of the fovs without the file extension
fovs = io_utils.remove_file_extensions(fovs)
```

If the user specifies a `fovs` list, the control statement will not be entered. However, since there are `fovs` with `.` in them, `remove_file_extensions` will then attempt to remove everything past the `.`. So `fov.1` and `fov.2` become just `fov` and `fov`.

**To Reproduce**

Specify a `fovs` list containing `.` like above.

**Expected behavior**

These `fov` names should be parsed correctly as `fov.1` and `fov.2` in the example above.

srivarra:

Closes #537.

Allows more complex fov name parsing in `remove_file_extensions`. 
Removes the extensions correctly for files named such as:

```sh
"fov.1" -> "fov.1"
"fov.2.tiff" -> "fov.2"
"fov3" -> "fov3"
"fov4.tar.gz" -> "fov4.tar"
```

Added explicit checks for the following file extensions:

`"tiff", "tif", "png", "jpg", "jpeg", "tar", "gz", "csv", "feather"`.

Added an extra test for more complicated fov filenames.

**Remaining issues**

Should we account for more file extensions?


ngreenwald (changes requested):
Not sure why coverage is dropping, can you see if this is real or an artifact?

ngreenwald (approved):

PR #579, opened 2022-05-31
Pix preprocessing

ngreenwald:

Adds in functionality to normalize each channel of image data data separately prior to pixel clustering. This helps to make sure that markers which have different intensity values are treated equally right from the beginning of the clustering process. 

In addition, it changes from removing pixels that have 0 total counts to removing pixels in the bottom 5% of total counts from the image. This better matches the format of the data following rosetta, where there are very few true zeros. 

**Remaining issues**
The testing I put together is very basic. @alex-l-kong, if you could go in and double check that everything is working as intended, and if needed adding more testing, that would be great. Also feel free to change the organization/saving structure if you think it would be better some other way. I didn't add any new tests for create_pixel_matrix, that will likely need to be checked as well.


alex-l-kong:
@ngreenwald ok I'll take a look at this.

alex-l-kong:
@ngreenwald just tested it on Candace's dataset on my end, had to make one change to account for all-zero channels in `calculate_channel_percentiles` but other than that it runs without error.

ngreenwald:
Okay cool, let me know once it's ready to look at

alex-l-kong (approved):

ark/phenotyping/som_utils.py
```diff
@@ -19,6 +20,95 @@
...
+        for fov in fovs:
+
+            # load image data and remove 0 valued pixels
+            img = load_utils.load_imgs_from_tree(data_dir=tiff_dir, img_sub_folder=img_sub_folder,
+                                                 channels=[channel], fovs=[fov]).values[0, :, :, 0]
+            img = img[img > 0]
+
+            # record and store percentile
+            # NOTE: if there are no non-zero pixels, assign directly to 0 to avoid empty list error
+            img_percentile = np.quantile(img, percentile) if len(img) > 0 else 0
```
ngreenwald, line 58: I think it makes more sense to completely skip a blank channel, rather than counting it as a zero for the purposes of constructing the average

ark/phenotyping/som_utils_test.py
```diff
@@ -835,39 +913,63 @@ def test_create_fov_pixel_data():
...
+        )
+
+        # assert the channel names are the same
+        misc_utils.verify_same_elements(flowsom_chans=sample_pixel_mat.columns.values[:-4],
+                                        provided_chans=chans)
+        misc_utils.verify_same_elements(flowsom_chans=sample_pixel_mat_subset.columns.values[:-4],
+                                        provided_chans=chans)
+
+        # assert no rows sum to 0
+        assert np.all(sample_pixel_mat.loc[:, ['chan0', 'chan1']].sum(axis=1).values != 0)
```
ngreenwald, line 933: Is there a more precise check we could add here to ensure that this addition is working? This test would pass with the previous version and this version

ark/phenotyping/som_utils_test.py
```diff
@@ -835,39 +913,63 @@ def test_create_fov_pixel_data():
...
+@parametrize('fovs', [['fov0', 'fov1', 'fov2'], ['fov0', 'fov1'], ['fov0']])
+@parametrize('chans', [['chan0', 'chan1', 'chan2'], ['chan0', 'chan1'], ['chan0']])
+@parametrize('sub_dir', [None, 'TIFs'])
+@parametrize('seg_dir_include', [True, False])
+@parametrize('channel_norm_include', [True, False])
+@parametrize('pixel_norm_include', [True, False])
+def test_create_pixel_matrix(fovs, chans, sub_dir, seg_dir_include,
+                             channel_norm_include, pixel_norm_include):
+    # # tests for all fovs and some fovs, and all chans and some chans
+    # fov_lists = [['fov0', 'fov1', 'fov2'], ['fov0', 'fov1'], ['fov0']]
```
ngreenwald, line 953: Old lists can be removed

ngreenwald (commented):
Couple comments

ark/phenotyping/som_utils_test.py
alex-l-kong: @ngreenwald because we're dividing by row sums, we can change this test to ensure that all rows sum to 1 to test normalization with different `pixel_norm_val` parameters. Is this what you were thinking about, or something more specific?

alex-l-kong:
@ngreenwald just one clarification question, otherwise should be good to go.

ark/phenotyping/som_utils_test.py
ngreenwald: I mean something that checks that passing `pixel_norm_val` is working as intended. For example, an expected decrease in the total amount of pixels included in the df or something like that

alex-l-kong: @ngreenwald oh ok. I can do something along the lines of:

```
assert sample_pixel_mat.shape[0] < (sample_img_data.shape[0] * sample_img_data.shape[1])
```

This will assert that we actually generated fewer pixels in `sample_pixel_mat` than there exist in `sample_img_data`.

The opposite test:

```
assert sample_pixel_mat.shape[0] == (sample_img_data.shape[0] * sample_img_data.shape[1])
```

would be good for the other 2 tests where no pixels are removed by `pixel_norm_val`.

alex-l-kong:
@ngreenwald OK the above comment about testing pixel filtering with `pixel_norm_val` should be addressed now.

ngreenwald:
gonna test this out myself a bit more before merging it in to make sure nothing got missed

ngreenwald:
Looks good, once @cliu72 approves I'll merge it in

ark/phenotyping/som_utils.py
```diff
@@ -637,6 +733,40 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
     # set seed for subsetting
     np.random.seed(seed)
 
+    # create path for channel normalization values
+    channel_norm_path = os.path.join(base_dir, pre_dir, 'channel_norm.json')
```
cliu72, line 737: Just a small comment. We are currently saving the 99.9% normalization values (that existed before) as prefix_norm_vals.feather in the base directory. Should we also change that to a json to be consistent? Or else save these as a feather too? And maybe move that file into the preprocessed directory as well?

Also, I think it would be helpful to change the names of both of these files, so users are not confused which file is corresponding to which normalization. Maybe like "channel_norm" and "post_pixelnorm_channel_norm" or something?

cliu72 (commented):
Just a few small comments. Otherwise looks good to me.

PR #581, opened 2022-06-01
Custom channel smoothing

ngreenwald:

Adds in functionality for different smoothing of individual channels in pixel clustering

Provides the user a separate step to specify individual channels for increased smoothing, along with custom smoothing coefficients. 

Also adds in an error check to `create_pixel_matrix` which identifies cases where the user provided the original channel name, but a modified channel name exists. This will be useful for other pre-processing steps we introduce which produce modified channels


ark/phenotyping/som_utils.py
```diff
@@ -55,6 +56,75 @@ def normalize_rows(pixel_data, channels, include_seg_label=True):
...
+    else:
+        raise ValueError("Variable smooth_vals must be either a single integer or a list")
+
+    for fov in fovs:
+        for idx, chan in enumerate(channels):
+            img = load_utils.load_imgs_from_tree(data_dir=tiff_dir, img_sub_folder=img_sub_folder,
+                                                 fovs=[fov], channels=[chan]).values[0, :, :, 0]
+            chan_out = ndimage.gaussian_filter(img, sigma=smooth_vals[idx])
+            imsave(os.path.join(tiff_dir, fov, img_sub_folder, chan + '_smoothed.tiff'),
+                   chan_out, check_contrast=False)
```
alex-l-kong, line 125: On average how many channels will need additional smoothing in a cohort? If it's frequently going to be a lot, might be more efficient to put this in `create_fov_pixel_data` (which is enclosed in the per-FOV loop in `create_pixel_matrix`) to prevent the inefficiency of looping over each `channel` once to add the first `gaussian_filter`, then apply the "standard" `gaussian_filter` again in `create_fov_pixel_data`.

If only a few channels will need additional smoothing at most, this is a moot point.

alex-l-kong (commented):
Just one comment about potentially condensing one of the loops.

cliu72 (commented):
When I was testing, I got these messages:
image

Could potentially get annoying when you have a lot of markers that are substrings of others. Maybe instead of checking for a substring match, check for "_smoothed"? Don't have strong opinions either way.

ark/phenotyping/som_utils.py
cliu72: In my opinion, it will usually be a small number of markers that need additional smoothing.

ngreenwald:
Alex, can you take a look at why RTD is failing? I copy pasted the bottom half of the docstring from the function below it to make sure nothing is different but it's still complaining

alex-l-kong:
@ngreenwald OK documentation has been fixed.

PR #595, opened 2022-06-10
Remove `channel_norm.feather` and `pixel_norm.feather` from `cluster_pixels` verification

alex-l-kong:

Because the `channel_norm` and `pixel_norm` files were changed from `.json` to `.feather` format in `pixel_mat_dir`, they now get read in along with the `fov` files when retrieving them for verification in this line:

```
data_files = io_utils.list_files(data_path, substrs='.feather')
```

In some cases, this can cause `channel_norm.feather` to be read in as a FOV file by mistake. This needs to be prevented.

Add two `.remove` statements after the aforementioned line to prevent `channel_norm.feather` and `pixel_norm.feather` from interfering. Because these are hard-coded names set in `create_pixel_matrix`, the naming will not cause any issues.

### DIFF
--- a/ark/phenotyping/som_utils.py
+++ b/ark/phenotyping/som_utils.py
@@ -1050,7 +1050,10 @@ def cluster_pixels(fovs, channels, base_dir, data_dir='pixel_mat_data',
                                 (weights_name, base_dir))
 
     # verify that all provided fovs exist in the folder
+    # ensure that channel_norm and pixel_norm files are ignored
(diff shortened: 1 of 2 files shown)


ngreenwald (changes requested):
Instead, lets change it so that these files are in the top-level directory, with the 99.9% norm feather file. That way all of the preprocessing related feathers are in the same spot, and we won't need to hardcode anything

ngreenwald (approved):

PR #607, opened 2022-06-20
Allow segmentation pipeline to handle blank images

ngreenwald:

Updates the segmentation pipeline to work with images without any cells. 

Check for number unique objects in segmentation mask. If there aren't any cells, skips the `regionprops` extraction and just returns an empty df. Also adds tests to prevent future reversions

### DIFF
--- a/ark/segmentation/marker_quantification.py
+++ b/ark/segmentation/marker_quantification.py
@@ -41,6 +41,12 @@ def get_single_compartment_props(segmentation_labels, regionprops_base,
         props_options=list(REGIONPROPS_FUNCTION.keys())
     )
 
+    # if image is just background, return empty df
+    if len(np.unique(segmentation_labels)) < 2:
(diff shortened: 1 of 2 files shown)


ark/segmentation/marker_quantification.py
```diff
@@ -41,6 +41,12 @@ def get_single_compartment_props(segmentation_labels, regionprops_base,
         props_options=list(REGIONPROPS_FUNCTION.keys())
     )
 
+    # if image is just background, return empty df
+    if len(np.unique(segmentation_labels)) < 2:
+        output_list = regionprops_base + regionprops_single_comp
+        blank_df = pd.DataFrame(columns=output_list)
+        return blank_df
+
```
srivarra, line 49: Should we add a warning to notify the user that there are no cells within the image they've supplied in addition to returning an empty dataframe?

PR #609, opened 2022-06-20
Removal of `force_ints` in all load functions.

### LINKED ISSUE #606: future warning
The segmentation pipeline has a future warning that is coming up during the segmentation process

```
/usr/local/lib/python3.6/site-packages/pandas/core/indexing.py:1418: FutureWarning: 
Passing list-likes to .loc or [] with any missing label will raise
KeyError in the future, you can use .reindex() as an alternative.

See the documentation here:
https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#deprecate-loc-reindex-listlike
  return self._getitem_tuple(key)
```
The main reason this is an issue is because in some FOVs, it's getting raised 10s of times, which makes it impossible to see anything else in the output. 
image

Determine if this is the same pandas error that @srivarra resolved in the migration to 3.7 branch. If not, we can add this to the list of things to get addressed.

srivarra:

Closes #606.

`load_imgs_from_tree`  and `load_imgs_from_dir` will no longer allow any type conversions. There are significantly fewer user accessible type conversion options throughout ark.

Removed the majority of front facing instances `dtype` in function arguments. In `deepcell_service_utils` the raw bytes output from `Deepcell` will now be converted from floating point to integer values.

**Remaining issues**

The JSON returned by Deepcell has the following metadata when running the `Segment_Image_Data.ipynb` notebook:
`{"shape": [1, 1024, 1024]}`. This is the same for this branch, master (after #605 was merged in), and with the last commit in #624. Seems like it was just something I got confused with when I was adjusting the reshaping earlier in the PR process.


ark/utils/load_utils.py
```diff
@@ -324,10 +294,6 @@ def load_imgs_from_dir(data_dir, files=None, match_substring=None, trim_suffix=N
     if channel_indices and multitiff:
         img_data = img_data[:, :, :, channel_indices]
 
-    # check to make sure that dtype wasn't too small for range of data
-    if np.min(img_data) < 0:
```
ngreenwald, line 328: We should keep this around, and add it to the other load functions, but make it a warning, not an error. In general it probably means something went wrong if you're loading images with negative values, so this would be a good thing to have a basic check. It's also possible that the test_img had a different dtype than the other images in the directory. 

There are some cases where the user might be loading z-scored images, where negatives are expected, so we don't automatically want to raise an error, but good to flag

ark/utils/load_utils_test.py
```diff
@@ -206,60 +186,72 @@ def test_load_imgs_from_dir():
...
-                                          dtype=np.float32)
+                                          trim_suffix='_', xr_dim_name='compartments')
+
+    # Test floating point xarray type consistency (i.e. creation type == load_imgs type)
+    with tempfile.TemporaryDirectory() as temp_dir:
+
+        fovs, _ = test_utils.gen_fov_chan_names(num_fovs=3, num_chans=0, use_delimiter=True)
+        filelocs, data_xr = test_utils.create_paired_xarray_fovs(temp_dir, fovs, [0],
+                                                                 img_shape=(10, 10), mode='labels',
+                                                                 delimiter='_', dtype=np.float32)
```
ngreenwald, line 230: This and the one below can get turned into a for loop over dtype creation values

ark/utils/deepcell_service_utils.py
```diff
@@ -136,8 +141,12 @@ def zip_write(zip_path):
 
         with ZipFile(zip_files[-1], "r") as zipObj:
             for name in zipObj.namelist():
-                with open(os.path.join(deepcell_output_dir, name), mode='wb') as f:
-                    f.write(zipObj.read(name))
+                mask_path = os.path.join(deepcell_output_dir, name)
```
ngreenwald, line 144: Did you actually run the notebook with real data to make sure this works as expected, as well as inspecting the segmentation mask output from before and after the change to make sure segmentation looks identical?

ngreenwald (changes requested):
Looks great, just a couple small changes

ngreenwald (approved):

PR #620, opened 2022-06-24
Add option to restart pixel clustering notebook from crash

### LINKED ISSUE #575: Allow pixel/cell clustering pipeline to restart from checkpoint in case of crash

Testing of `example_pixel_clustering.ipynb` and `example_cell_clustering.ipynb` has revealed instances of crashing. If the user needs to restart Docker (ex. to increase memory), it would be inconvenient for them to have to re-run the whole pipeline (or at least one whole step) because of a crash during one component. We should add a way to restart the pipeline at that point in time.

There are a couple of problematic scenarios I see this popping up where this will need to be addressed (potentially more to be added):

1. SOM training: this one could be the trickiest to implement, as the `SOM` function in `FlowSOM` does not currently support restarting from a checkpoint. This is a feature we may need to add in our local forked version of `FlowSOM` if desired. We can see if we can save a snapshot of the weights and the number of passes we've already computed.
2. Pixel clustering and consensus assignment: in this case, we should have a control statement in the assignment loop to prevent having to re-assign FOVs that have already been fully processed. We should also check to see if the weights in a particular run folder have been generated so we don't re-run the SOM training.

alex-l-kong:

Closes #575. `example_pixel_clustering.ipynb` includes very expensive processes such as preprocessing, SOM training, and SOM cluster assignment. This PR provides a way to prevent these from re-running (at least in full) for items that have already been checked off on a previous run that crashed midway through.

Add options to check for files, or specific columns in files, indicating that certain processes have finished or not. This ensures we only need to process FOVs that haven't been completed.


alex-l-kong:
@ngreenwald the remaining issue to solve on this PR is how to handle the case when the Arrow file becomes corrupted due to crashing mid-save. Unfortunately, `feather` doesn't offer a way to recover from this, and since we overwrite each `fov` feather file when assigning SOM cluster labels, we lose a way of reverting back to the original copy.

The simplest solution is to add a `_temp.feather` for each `fov` while running the SOM and meta cluster assignment process, allowing the data to be recovered in case of a mid-save crash. Open to other ideas too.

ngreenwald:
I just ran into this same issue, but it was because I stopped the processing, not because it crashed. I think the _temp file would be a good idea if we only had one function that manipulated the .feather files, but because we have multiple, it might not be obvious which one caused the corrupted file, and hence which temp file to look at. 

I think as a first pass, it's okay to just wrap the `feather.read` call in a try/except clause, and if it's corrupted, say which feather file is the problem. Then it's fairly straightforward to re-run the preprocessing just for that FOV. 

If lots of people end up having this issue come up we can revisit an automated way to handle it, but I think for now we can punt on it

ngreenwald:
Looks like this is happening a fair amount of the time, even without a crash. It looks like about 1% of feather files get saved with some sort of corrupted issue: here's a list of the exceptions that are getting raised when trying to load them
image

image

alex-l-kong:
@ngreenwald that is very strange, never encountered an error like this before. Was there a specific exception being raised without the `try`/`except` block (ex. `ArrowInvalid`)?

ngreenwald:
Yeah, it's getting raised with an OSerror, the details of which are in the screenshot above 
image

ngreenwald:
It's reproducible, unlike the previous bugs with corrupted feather files. I can read the same file again and get the same error. I'll send you a couple via slack so you can take a look

alex-l-kong:
@ngreenwald I've added an update to force the compression to `'uncompressed'` when saving the feather files in R. Didn't encounter any issues on my end, can you see if this fix does the trick for you?

ngreenwald:
You didn't encounter any issues before. Is there a reason to think it's the compression that's causing the problems?

alex-l-kong:
> You didn't encounter any issues before. Is there a reason to think it's the compression that's causing the problems?

I'm guessing the reason this issue pops up is due to what's written R `write_feather` documentation:

```
Name of compression codec to use, if any. Default is "lz4" if LZ4 is available in your build of the Arrow C++ library, otherwise "uncompressed". "zstd" is the other available codec and generally has better compression ratios in exchange for slower read and write performance See [codec_is_available()](https://arrow.apache.org/docs/r/reference/codec_is_available.html). This option is not supported for V1.
```

I know in Python, I had to force the compression method to `'uncompressed'`, however in `R` this was never an issue. We did update the Docker to use the Python 3.7 image very recently, and combined with a fairly recent release of the internal `arrow` library which provides the `feather` functionality, it's possible that the default compression method changed to LZ4. Hopefully forcing it to `'uncompressed'` in R does the trick for you.

It's still weird, though, because even with this new image, I was able to run it smoothly without needing to force-feed a compression method in `R`. Have you encountered the same issue with Candace's dataset and verified that it runs on your end?

alex-l-kong:
@cliu72 would you mind running the code on this branch to see if you encounter a similar error as @ngreenwald with compression?

alex-l-kong:
@ngreenwald @cliu72 there is a slight issue when it comes to restarting on the call to `create_pixel_matrix`. Because of the need for random indexing, even with a random seed, different random indices for subsetting will be mapped to the remaining FOVs, causing the SOM and meta cluster assignments to change (as opposed to if the process ran to completion).

While it may be worth noting in the documentation of this function, I don't think it's a big issue since I don't expect restarting to happen frequently. Let me know what you guys think.

ngreenwald:
You mean that after a restart, a different random subset will be taken for the subsequent FOVs? I agree, I don't think that's an issue

ngreenwald:
What's the difference in the size of the feather file and the speed of I/O for lz4 vs uncompressed?

alex-l-kong:
A few more tests, the timing can vary but in general it still seems like uncompressed outperforms LZ4 in terms of speed:

Screen Shot 2022-06-27 at 1 39 39 PM

Screen Shot 2022-06-27 at 1 40 56 PM

cliu72:
> @cliu72 would you mind running the code on this branch to see if you encounter a similar error as @ngreenwald with compression?

Tested it a bit and didn't run into issues (but I also wasn't running to any crashing issues before this on cloud).

ngreenwald:
@alex-l-kong can you create a fresh branch that just has the compression changes in it so I can test if that resolves the issue, without all of the checkpointing/file creation logic that's in here

alex-l-kong:
@ngreenwald this is mostly done, I haven't tested it on the MIBIAN yet (think the password I have is incorrect) but this should work.

Just one more thing needs to be tested out before officially requesting a review: ensuring that `quant_dat` updates properly in `create_pixel_matrix` if a restart is needed. Currently, it doesn't have a save state, meaning if `create_pixel_matrix` crashes, it only computes the channel normalized values on FOVs that come after the last FOV before crashing. This will need to be updated so 99.9% normalization can happen correctly even if after a crash.

alex-l-kong:
@ngreenwald there is one minor issue popping up with occasional hanging on R's end of multi-processing, which happens when the first FOV in a particular batch is corrupted. This issue should only pop up in extraordinary circumstances since we're now writing to a separate directory, instead of overwriting `.feather` files. Still, it's an issue that I'm going to look into for the time being.

ark/phenotyping/pixel_consensus_cluster.R
```diff
@@ -100,11 +109,46 @@ for (batchStart in seq(1, length(fovs), batchSize)) {
...
+            }
+        )
+
+        data.frame(fov=fovs[i], status=status)
+    }
+
+    # report any erroneous feather files
+    for (i in 1:nrow(fovStatuses)) {
+        if (fovStatuses[i, 'status'] == 1) {
+            print(paste("The data for FOV", fovStatuses[i, 'fov'], "has been corrupted, removing"))
```
ngreenwald, line 149: Does this file get removed somewhere? How often do these corruption issues come up now that we've removed the feather overwriting? Is it related to multi-processing, or inherent to writing the files

ark/phenotyping/run_pixel_som.R
```diff
@@ -87,11 +67,55 @@ for (batchStart in seq(1, length(fovs), batchSize)) {
...
+                print(paste('Error encountered for fov', fovs[i]))
+                1
+            }
+        )
+
+        data.frame(fov=fovs[i], status=status)
+    }
+
+    # report any erroneous feather files
+    for (i in 1:nrow(fovStatuses)) {
```
ngreenwald, line 114: Same comments as consensus_clustering R file

ark/phenotyping/som_utils.py
```diff
@@ -271,9 +273,13 @@ def compute_pixel_cluster_channel_avg(fovs, channels, base_dir, pixel_cluster_co
...
         # read in the fovs data
-        fov_pixel_data = feather.read_dataframe(
-            os.path.join(base_dir, pixel_data_dir, fov + '.feather')
-        )
+        try:
+            fov_pixel_data = feather.read_dataframe(
+                os.path.join(base_dir, pixel_data_dir, fov + '.feather')
+            )
+        except (ArrowInvalid, OSError, IOError):
+            print("The data for FOV %s has been corrupted, removing" % fov)
```
ngreenwald, line 281: I think "skipping" is more accurate, since we're not actually removing it. 

The user will need to re-run the whole pipeline for this specific FOV. We should give them an error message indicating that, and also delete this specific feather file so that when they re-generate it there's no overwriting issues

ark/phenotyping/som_utils.py
```diff
@@ -937,13 +943,50 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
     if not os.path.exists(os.path.join(base_dir, subset_dir)):
         os.mkdir(os.path.join(base_dir, subset_dir))
 
+    # create variable for storing 99.9% values
+    quant_dat = pd.DataFrame()
+
+    # define a path for storing the intermediate quant_dat values, needed in case of restart
+    # TODO: change this to saving in the pixel_output_dir after merging in master
+    quant_dat_path = os.path.join(base_dir, pixel_output_dir, 'quant_dat.feather')
```
ngreenwald, line 951: What is the scenario where creating the 99.9% normalized values would fail? We haven't generated the feather files yet. What type of error is this guarding against?

ark/phenotyping/som_utils.py
```diff
@@ -1337,8 +1526,16 @@ def pixel_consensus_cluster(fovs, channels, base_dir, max_k=20, cap=3,
         os.path.join(base_dir, clust_to_meta_name)
     ).astype(np.int64)
 
-    # read in the channel-averaged results across all pixel SOM clusters
-    pixel_channel_avg_som_cluster = pd.read_csv(som_cluster_avg_path)
+    # because FOVs may become corrupted, we also need to re-compute average channel expression
```
ngreenwald, line 1529: I think it's fine to use the previous version, 1 or 2 corrupted FOVs is not going to dramatically change this

ark/phenotyping/som_utils.py
alex-l-kong: Let's say the user runs a cohort of 100 FOVs, and `create_pixel_matrix` crashes 97 FOVs in. Previously, we would have lost any reference to the 97 FOVs in `quant_dat`. On restart, `quant_dat` would only contain the data for the last 3 FOVs, which would affect the calculation of `mean_quant` and 99.9% normalization prior to training the SOM.

Intermediate saving of `quant_dat` helps guard against this. It's mostly to ensure correctness, but I'll remove it if you don't anticipate this being an issue.

ark/phenotyping/som_utils.py
alex-l-kong: True, when re-running the meta cluster remap I noticed that only the pixel count barplot raw numbers changed slightly. The heatmaps didn't look different.

ark/phenotyping/pixel_consensus_cluster.R
```diff
@@ -78,11 +67,29 @@ clusterAvgsScale <- scale(clusterAvgs[,markers])
 clusterAvgsScale <- sapply(as.data.frame(clusterAvgsScale), pmin, cap)
 clusterAvgsScale <- sapply(as.data.frame(clusterAvgsScale), pmax, -cap)
 
-# run the consensus clustering
-print("Running consensus clustering")
-suppressMessages(consensusClusterResults <- ConsensusClusterPlus(t(clusterAvgsScale), maxK=maxK, seed=seed))
-som_to_meta_map <- consensusClusterResults[[maxK]]$consensusClass
-names(som_to_meta_map) <- clusterAvgs$pixel_som_cluster
+# define a temporary som_to_meta_map .feather path, used for checkpointing
+clustToMetaTempPath <- stri_replace(clustToMetaPath, '_temp.feather', fixed='.feather')
```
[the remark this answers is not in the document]
alex-l-kong: No downside, it's pretty fast. Agreed it's an unnecessary step considering this, I'll remove it.

ark/phenotyping/pixel_consensus_cluster.R
alex-l-kong: This could happen if the writing process for a FOV in `run_pixel_som.R` terminated early (ex. the kernel crashed). On restart, since the FOV is technically partially written, `run_pixel_som.R` will not check to re-generate this (it would be cumbersome to open up 100s of large FOVs to see which ones were invalid). So this will need to be handled by `pixel_consensus_cluster.R`.

Mostly, this exception handling is in place to prevent overnight crashing from stopping the entire pipeline. That being said, I agree that this issue shouldn't come up frequently with the new directory structure, so I can remove it.

alex-l-kong:
@ngreenwald per our discussion today:

1. Leave the `try`/`except` blocks/`tryCatch` blocks in. These are helpful for reporting corrupted FOV issues to the user, and they will not stop the pipeline from running.
2. Remove the intermediate calculations for `quant_dat` and `som_to_meta_map`. The former should not be necessary because `create_pixel_matrix` writes fresh data, so it should not crash. Since `ConsensusClusterPlus` is very fast, we don't need to add restart functionality for `som_to_meta_map`.

ark/phenotyping/pixel_consensus_cluster.R
alex-l-kong: @ngreenwald per our discussion today, we'll leave the `try`/`except` logic in.

ark/phenotyping/run_pixel_som.R
alex-l-kong: @ngreenwald same follow-up as above.

ngreenwald (changes requested):
Okay great. And you ran this version, with and without a subset of the files already being processed for each step, to check that everything works?

alex-l-kong:
> Okay great. And you ran this version, with and without a subset of the files already being processed for each step, to check that everything works?

Yes.

PR #623, opened 2022-06-27
Add `dtype` argument to channel and pixel norm functions

### LINKED ISSUE #612: Add dtype to calculate_channel_percentiles and calculate_pixel_intensity_percentile in som_utils.py
Right now, `create_pixel_matrix` is showing a warning about dtype, even when the appropriate dtype is specified in `create_pixel_matrix`. I think it is because of `calculate_channel_percentiles` and `calculate_pixel_intensity_percentile`.

The user can specify a dtype in `create_pixel_matrix`. Pass that dtype into `calculate_channel_percentiles` and `calculate_pixel_intensity_percentile` when loading images from tree. Default for those are int16 right now.

**Additional context**
image

alex-l-kong:

Closes #612. The `dtype` argument currently isn't propagated through to the channel and pixel normalization functions, causing the `load_imgs_from_tree` calls to attempt to load the images using the default `int16` type. This will throw a warning if the images were of type `float32` regardless of if the user passed `float32` to `create_pixel_matrix`.

Add `dtype` with a default argument of `int16` to both `calculate_channel_percentiles` and `calculate_pixel_intensity_percentile`. Explicitly pass the `dtype` argument received by `create_pixel_matrix` to these functions.

### DIFF
--- a/ark/phenotyping/som_utils.py
+++ b/ark/phenotyping/som_utils.py
@@ -26,7 +26,7 @@ multiprocessing.set_start_method('spawn', force=True)
 
 
 def calculate_channel_percentiles(tiff_dir, fovs, channels, img_sub_folder,
-                                  percentile, dtype="int16"):
+                                  percentile, dtype="float32"):
     """Calculates average percentile for each channel in the dataset
 
     Args:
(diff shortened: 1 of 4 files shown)


cliu72 (approved):
Looks good to me

ngreenwald (changes requested):
Did you run the notebook to confirm there aren't any other `load_imgs` functions that also need to be passed the dtype parameter?

alex-l-kong:
> Did you run the notebook to confirm there aren't any other `load_imgs` functions that also need to be passed the dtype parameter?

Good point, `smooth_channels` was the remaining function that was missing an explicit `dtype` arg.

ngreenwald:
Will you also default it to 'float32' since that's what rosetta produces

alex-l-kong:
> Will you also default it to 'float32' since that's what rosetta produces

Should we also change the default argument of `create_pixel_matrix` to `float32`?

ngreenwald:
Yeah definitely

ngreenwald (approved):

PR #625, opened 2022-06-28
Force the compression method of R feather saving `'uncompressed'`

alex-l-kong:

There have been issues with consistency in saving the `.feather` files in Python vs R. To prevent interoperability issues, we change the compression argument passed into `write_feather` in R to `'uncompressed'`


ngreenwald:
It wasn't the compression. Getting around the same number of errors, now with a different error message. 
image. Given that I used the MIBIAN a fair amount beforehand without running into these errors, it seems likely that its either the multiprocessing update or the switch to 3.7 that is causing these issues

ngreenwald:
This never popped up on my test dataset. I'm going to rerun clustering there to confirm that the issue comes up now. If so, it means it was one of the recent changes. Can you work with @srivarra to get a python 3.6 image (tagged with something besides latest) pushed to dockerhub so that I can rollback and see if that resolves it?

ngreenwald:
This was resolved in a previous PR

PR #647, opened 2022-08-09
Context tests

ackagel:
## **Purpose**

Adds basic coverage tests for context dependent spatial enrichment analysis

## **Implementation**

Uses existing enrichment testing infrastructure to test for the errorless operation of enrichment functions when provided context arguments.

## **Remaining issues**

Functionality tests should be added in a future PR.  Using tools like pytest-cases + fixtures to clean up cell table generation in`test_utils.py` would really clean up (at a minimum) `spatial_analysis_tests.py`, and would make adding functionality tests (context dependence, feature distance-hack, etc) much easier.

### DIFF
--- a/ark/analysis/spatial_analysis_test.py
+++ b/ark/analysis/spatial_analysis_test.py
@@ -197,6 +197,17 @@ def test_calculate_channel_spatial_enrichment():
     assert stats_no_enrich.loc["fov9", "p_neg", 3, 2] > .05
     assert abs(stats_no_enrich.loc["fov9", "z", 3, 2]) < 2
 
+    # run basic coverage check on context dependence code
+    all_data_context, dist_mat_context = \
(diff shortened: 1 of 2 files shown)


ackagel:
@ngreenwald I decided to just quarantine the context cell-table for its own test, rather than add to the existing complexity of the enriched cell-table generation code.  

When functionality tests are added (post-workshop PR?), I think the enriched cell-table generation in `test_utils.py` should be reworked w/ tools like pytest-cases, i.e made more similar to toffy's tests.  It'll be easier to add in context enrichment generation or any future cases, and would also help in the case of potential cell-table format changes (i.e no locally hardcoded column locations+names).

ngreenwald (approved):

ngreenwald:
Sounds good, agreed that whole area of the codebase could use some attention with out newfound knowledge

PR #653, opened 2022-08-13
Add nuclear filtering functionality to pixel clustering pipeline

### LINKED ISSUE #619: Add nuclear pixel filtering to clustering pipeline
Sometimes markers will have background staining that we want to filter out for the clustering process. One way to accomplish this is using the subcellular localization of the marker. For markers which should only be in the nucleus, we remove non-nuclear signal. Conversely, for markers that should only be the cytoplasm/membrane, we filter out nuclear signal. 

Add this function to the repo, with comments, tests, etc. 

```
from ark.utils.load_utils import load_imgs_from_tree, load_imgs_from_dir
from ark.utils.misc_utils import verify_same_elements
import skimage.io as io

def filter_with_nuclear_mask(fovs, tiff_dir, segmentation_dir, channel, img_sub_folder=None, exclude=True):
    # convert to path-compatible format
    if img_sub_folder is None:
        img_sub_folder = ''
        
    
    for fov in fovs:
        img = load_utils.load_imgs_from_tree(data_dir=tiff_dir, img_sub_folder=img_sub_folder,
                                             fovs=[fov], channels=[channel]).values[0, :, :, 0]
        seg_img = io.imread(os.path.join(segmentation_dir, fov + '_feature_1.tif'))[0, ...]
        
        if exclude:
            suffix = '_nuc_exclude.tiff'
            seg_mask = seg_img > 0
        else:
            suffix = '_nuc_include.tiff'
            seg_mask = seg_img == 0
            
        img[seg_mask] = 0
        io.imsave(os.path.join(tiff_dir, fov, img_sub_folder, channel + suffix), img, check_contrast=False)
```

alex-l-kong:

Closes #619. Prior to pixel clustering, users may need to filter out non-nuclear marker signal from the nucleus and vice-versa for nuclear marker signal.

Add a function `filter_with_nuclear_mask` to `som_utils` which runs this process. The code in #619 provides most of the functionality, however we'll need to make two modifications:

- Ensure `seg_dir is not None` prior to running. Users may run pixel clustering without segmentation labels, however they will not be able to run this function if that's the case because it relies on the `_feature_1.tiff` files.
- Ensure `channel` defaults to `None` in the notebook and function. This is the easiest guard against errors caused by running all the cells at once in the notebook.

The cell that calls this function in `example_pixel_clustering` will be placed after the smoothing process.

This PR also solidifies some of the notebook testing, as it was not using the renamed channels after smoothing. This should also apply for the nuclear filtering tests too.

### DIFF
--- a/ark/phenotyping/som_utils.py
+++ b/ark/phenotyping/som_utils.py
@@ -238,7 +238,7 @@ def smooth_channels(fovs, tiff_dir, img_sub_folder, channels, smooth_vals):
                    chan_out, check_contrast=False)
 
 
-def filter_with_nuclear_mask(fovs, tiff_dir, seg_dir, channel,
+def filter_with_nuclear_mask(fovs, tiff_dir, seg_dir, channel=None,
                              img_sub_folder=None, exclude=True):
     """Filters out background staining using subcellular marker localization.
 
@@ -260,6 +260,15 @@ def filter_with_nuclear_mask(fovs, tiff_dir, seg_dir, channel,
             Whether to filter out nuclear or membrane signal
     """
 
+    # if seg_dir is None, the user cannot run filtering
+    if seg_dir is None:
+        raise ValueError('seg_dir cannot be set to None for nuclear filtering')
+
+    # if the user doesn't specify a channel, return
+    if channel is None:
+        print("No channel specified for nuclear filtering, skipping")
+        return
+
     # convert to path-compatible format
     if img_sub_folder is None:
         img_sub_folder = ''
@@ -272,8 +281,6 @@ def filter_with_nuclear_mask(fovs, tiff_dir, seg_dir, channel,
         # load the segmented image in
         seg_img = imread(os.path.join(seg_dir, fov + '_feature_1.tif'))[0, ...]
 
-        # print(seg_img)
-
         # mask out the nucleus
         if exclude:
             suffix = '_nuc_exclude.tiff'
@@ -283,12 +290,8 @@ def filter_with_nuclear_mask(fovs, tiff_dir, seg_dir, channel,
             suffix = '_nuc_include.tiff'
             seg_mask = seg_img == 0
 
-        # print(img)
-        # print(seg_mask)
-
         # filter out the nucleus or membrane depending on exclude parameter
         img[seg_mask] = 0
-        # print(img)
 
         # save filtered image
         imsave(os.path.join(tiff_dir, fov, img_sub_folder, channel + suffix), img,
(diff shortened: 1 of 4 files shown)


templates_ark/example_pixel_clustering.ipynb
```diff
@@ -174,7 +174,7 @@
    "cell_type": "markdown",
```
ngreenwald, line 174: <p>Rather than having this as None, we should tell the user to skip the cell if they don't want to run it</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/653/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

ark/phenotyping/som_utils.py
```diff
@@ -238,6 +238,66 @@ def smooth_channels(fovs, tiff_dir, img_sub_folder, channels, smooth_vals):
...
+            Channel to apply filtering to
+        img_sub_folder (str):
+            Name of the subdirectory inside `tiff_dir` containing the tiff files.
+            Set to `None` if there isn't any.
+        exclude (bool):
+            Whether to filter out nuclear or membrane signal
+    """
+
+    # if seg_dir is None, the user cannot run filtering
+    if seg_dir is None:
```
ngreenwald, line 264: Since we don't ever set this to None, I don't see why the user would. Probably just checking that it's a valid directory makes more sense

ngreenwald (changes requested):
Looks good, just a minor tweak

ark/phenotyping/som_utils.py
alex-l-kong: Users over on Marte's side have asked for `exampel_pixel_clustering.ipynb` support without a segmentation directory (ex. `seg_dir = None`), which Candace agreed to add in and provide support for in `create_pixel_matrix`. It might be worth discussing whether to continue supporting `seg_dir = None`, especially if we plan on adding more features requiring one.

Agreed that we do need to verify if `seg_dir` is a valid path.

ngreenwald: Ah you're right, we can keep it for now. There's definitely a use case for segmentation-free pixel clustering, we can give a more informative error message rather than failing with None as an invalid path

ark/phenotyping/som_utils.py
```diff
@@ -238,6 +238,66 @@ def smooth_channels(fovs, tiff_dir, img_sub_folder, channels, smooth_vals):
                    chan_out, check_contrast=False)
 
 
+def filter_with_nuclear_mask(fovs, tiff_dir, seg_dir, channel=None,
```
ngreenwald, line 241:
```suggestion
def filter_with_nuclear_mask(fovs, tiff_dir, seg_dir, channel,
```

ngreenwald (changes requested):

ngreenwald (approved):

Issue #663, opened 2022-08-19
`create_mantis_project` errors out when specifying segmentation path


alex-l-kong:
**Describe the bug**

When running `create_mantis_project` in either the pixel or cell clustering notebooks, the line to copy the segmentation file into the corresponding `fov` directory fails. This is because the path gets set incorrectly (https://github.com/angelolab/ark-analysis/blob/master/ark/utils/plot_utils.py#L463): `seg_dir` already contains the full relative path to the segmented images, so does not need to be prepended with `img_source_dir`.

**To Reproduce**

Run the final cell in each notebook to create the Mantis directory structure as is.

**Expected behavior**

The segmentation files (suffixed with `_feature_0.tif`) should be copied over correctly.

**Screenshots**

The error traceback is as follows:

<img width="1117" alt="Screen Shot 2022-08-19 at 1 34 11 PM" src="https://user-images.githubusercontent.com/31424707/185702911-a53af157-e349-45ef-b2c0-efa113fd6f66.png">

ngreenwald:
Yeah, I found the bug as well. I fixed it in #658. I can't figure out why the tests aren't failing though. 
`img_source_dir` needs to be removed
 https://github.com/angelolab/ark-analysis/blob/master/ark/utils/plot_utils.py#L463

PR #668, opened 2022-08-24
Update kmeans neighborhood notebook

cliu72:

Update and clean-up the k-means neighborhood notebook.

- The old notebook required the cell table to include both a column that specified a numeric id for each cell phenotype, and a column that had the cell phenotypes. I simplified this to only require the column with the cell phenotypes. Specifically, in `create_neighborhood_matrix`, removed `cluster_id_col`, which was redundant with `cluster_name_col`.
- Fixed a bug where the wrong cell phenotypes were being counted due to fovs missing certain cell phenotypes.
- Set default of `self_neighbor` to False, so a cell is not counting itself as its neighbor.
- Set cluster numbers to start at 1. Previously was starting at 0, which is confusing because in cell overlays, 0 usually corresponds to empty slide.
- Add inertia as a metric.
- Subsample cells for silhouette score calculation.
- Remove cells that have no neighbors before clustering.
- Scaled (z-scored) the heatmaps, which makes them more interpretable.
- Added heatmap that shows the composition of each neighborhood (basically the average feature vector for each cluster).
- Fixed overlay generation function so correct legend is shown.
- Added Mantis function (although this currently doesn't work, breaks because of segmentation directory issue).
- Added subheadings and more descriptive text.

**Remaining issues**

- Need to fix/add tests
- Currently, the mantis function fails, I think because of #663 
- Need to talk about whether to delete the part of the spatial LDA notebook that does kmeans


cliu72:
I fixed all the tests - I think we should be able to merge this in now. For now, I just have the notebook test read in example input that Adam set-up for the enrichment notebook (I didn't bother with testbook since I think we will be using the example dataset for testing?).

I also removed the mantis function for now - pending https://github.com/angelolab/ark-analysis/issues/669.

I also reverted everything in settings.py back to what it was before since it was breaking a bajillion tests. I think rejiggering settings.py should be its own issue.

Also, sorry I royally effed up the pull request somehow where all the commits are showing up....something happened when I was rebasing but hopefully it's not confusing.

@ngreenwald let me know if you want someone else to look it over/what else should be done before merging.

ngreenwald (changes requested):
Unfortunately, it is a bit confusing. Since there are 82 files modified, it's hard to see what was actually changed. Can you see if @alex-l-kong can help you move the changes to a fresh branch so that the actual file dif shows up?

Not the end of the world if not, we'll just need the names of the files you actually changed so we can look just at those when reviewing.

cliu72:
Yeah, the files that were changed were `spatial_analysis.py`, `spatial_analysis_utils.py`, `visualize.py`, and `example_pairwise_spatial_enrichment.ipynb` (and their respective test files).

This commit has basically all the substantial changes to the code: https://github.com/angelolab/ark-analysis/pull/668/commits/35acc6a5f06b3e62e26bc5151b5d0177b27c4203.
And this commit has basically all the changes to the tests: https://github.com/angelolab/ark-analysis/pull/668/commits/806855e1ceb6bee97aef0b9c65fb1371c89b5f58
(the rest of them are basically fixing typos/reverting files back to what they were).

I looked into a bit when I messed it up. I could just close this pull request and delete the branch, then open a new one. I can do that if it's preferable. @alex-l-kong let me know if you have any other suggestions.

alex-l-kong:
@cliu72 yeah I think that's the easiest way to go.

PR #675, opened 2022-08-25
Update README.md to account for new Docker tagging system

alex-l-kong:

Although `start_docker.sh` will no longer require explicit updates to ensure the right image gets pulled, the user still needs to ensure the correct image gets pulled.

`start_docker.sh` does automatically pull the image for you, but this only applies if you don't have the image already. If you need to overwrite an existing image, you'll need to explicitly run the `docker pull` command with the correct tag.

Change the Docker section of the `README.md` to explicitly tell the user which tag to look out for when running the `docker pull` command. This change will also be applied to the Docker update section.

**Remaining issues**

If we plan on overwriting the Docker images each time the dependencies change, then we can just allow `start_docker.sh` to update the image in the background without explicit input from the user.


alex-l-kong:
Since we'll be creating new Docker images each time it needs to be rebuilt (and never overwriting tags), we can just eliminate any need for the user to run `docker pull` (removing it from the `README.md` entirely), which simplifies the pipeline.

ngreenwald (approved):

Issue #566, opened 2022-05-26
Progress updates for compute intensive functions


ngreenwald:
**Is your feature request related to a problem? Please describe.**
There are a number of functions in the clustering/spatial LDA processing that can take a long time. Some of them have progress bars, some of them print updates every 10 FOVs, and some of them have no output tracking. This makes it hard to know how long something will take

**Describe the solution you'd like**
We should standardize the way runtime info is displayed. There are two important pieces of information. 
1) How much progress has the function made
2) How much time has elapsed, and how much time is remaining

If the tqdm library can handle both of these for us, then that's great. I don't remember the exact issue with time estimate in the deepcell function, but if it's not accurate than it's not helpful to include. Instead, we can print elapsed time after set intervals.

ngreenwald:
How many of these still need a progress bar?

alex-l-kong:
@ngreenwald we now print out an update every 5 FOVs or so, for the cohort you tested on MIBIAN how long did it take to print out each update. If it's a significantly long time, perhaps we can consider adding a progress bar to each intermediate step too.

ngreenwald:
So all of the pixel/cell clustering has either a progress bar or printed out FOV numbers? Is there anything in spatial LDA, KNN, etc that still needs this?

alex-l-kong:
Spatial LDA will definitely need a second look at because of how long it takes, I don't remember if Calico updated their training function to display progress so I'll look into this over the weekend.

KNN I'm assuming refers to neighborhood analysis? It's been a hot while since I've run that notebook, I remember it not taking a very long time to run on the tests I did, but I'll look into this one as well just in case.

ackagel:
Spatial enrichment doesn't have a progress bar currently; I can add one to the batching version this weekend, since those are the ones called by users anyways.

alex-l-kong:
Per discussions with @ngreenwald and @cliu72, we'll need to add the following progress bars:

1. The EDA calculation in the spatial LDA preprocessing notebook
2. The KMeans clustering in neighborhood analysis. This point will require more discussion because silhouette score is prohibitively slow for massive cohorts. @cliu72 and I will be discussing a subsetting method.

ngreenwald:
Talk to @ackagel, he did some work on the cluster evaluation for the spatial LDA preprocessing notebook. We should standardize across both notebooks if possible, and be consistent about what metrics we use

alex-l-kong:
[...]

PR #680, opened 2022-08-26
Add progress bar for the EDA calculation in spatial LDA

alex-l-kong:

Begins addressing #566. Our repo includes a lot of intensive functions, so we'll need to ensure the user receives progress updates for these so they don't get left in the dark.

Use the `tqdm` function in the `tqdm.notebook` function.

Based on our research and discussion with @ackagel, the only function that needs this for now is `spLDA.preprocessing.compute_topic_modeling`. There are various reasons others don't need them:

- They already have a progress bar or print outs per every `n` iterations
- They are direct calls to an external package, which we can't wrap easily in `tqdm` short of making a contribution to said package
- In the case of neighborhood analysis, progress will be delayed until after the MIBI workshop


alex-l-kong:
Also needed to update the `start_docker.sh` script, since it was attempting to pull a non-existant `spatial-ark-test` image. Added the `site-packages` reference back for the time being too.

ngreenwald (changes requested):
You'll need to coordinate with Adam about the plan for the local version of `ark`

alex-l-kong:
@ngreenwald just updated `start_docker.sh` to reflect @ackagel's work in `main`.

alex-l-kong:
@ngreenwald just updated `start_docker.sh` to reflect @ackagel's work in `main`.

ark/spLDA/processing.py
```diff
@@ -278,7 +279,8 @@ def compute_topic_eda(features, featurization, topics, silhouette=False, num_boo
     stats = dict(zip(stat_names, [{} for name in stat_names]))
 
     # iterative over topic number candidates
-    for k in topics:
+    pb_format = '{l_bar}{bar}| {n_fmt}/{total_fmt} [{elapsed}<{remaining}]'
+    for k in tqdm(topics, bar_format=pb_format):
```
ngreenwald, line 283: Did you run the notebook to make sure the display looks as expected?

ngreenwald (changes requested):

ark/spLDA/processing.py
alex-l-kong: It does.

PR #681, opened 2022-08-29
Speed up ark testing

### LINKED ISSUE #336: Speed up testing
We have a couple tests that appear to be taking much longer than they should. 
image

In particular, `test_example_spatial_analysis`, `test_dimesionality_reduction`, and `test_draw_boxplot` all take a long time. 

We should look into how to speed these up. Perhaps by passing in much smaller data for the test files (i.e. 100x100 instead of 1024x1024)

alex-l-kong:

Closes #336. A lot of tests take a significant amount of time to run. This addresses the obvious ones that can be reduced by changing the data size or eliminating operations that don't need testing.

Specifically, we're looking at tests that take 10 or more seconds to run. Stuff like dimensionality reduction and pixel clustering tests.

**Remaining issues**

The notebook tests in particular are problematic. The ones that test a complete run through using the default params specified in the notebook can only be sped up by changing the underlying data.

Dimensionality reduction is also prohibitively slow, even with a small dataset, because of the bottlenecks inherent in t-SNE and UMAPs implementations. For now, I've removed the test without file saving, might need to rethink this test further.


ngreenwald:
Parallel builds soon mean this isn't really an issue, can revisit if things become slow again

PR #705, opened 2022-09-13
Allow user to set number of cores for R functions

### LINKED ISSUE #695: Number of cores specified for multithreading in R pixel clustering scripts can exceed actual number of available CPUs

**Describe the bug**

On certain computers, setting `nCores <- parallel::detectCores() - 1` can cause the call to `parallel::makeCluster` to fail, especially if several cores are occupied at the time. The issue is that `parallel::detectCores()` only tells you how many CPUs are on the machine without telling the user how many are actually available.

**Expected behavior**

The value assigned to `nCores` should not exceed the number of actual CPUs available at the moment.

Unfortunately, R does not provide a way to tell us this directly. There are a couple ways around this:

- Default to `parallel::detectCores() - 1`, but allow the user to specify number of cores if needed
- Default to `parallel::detectCores() / 2`, not a fan of this option because it's overkill in most cases

alex-l-kong:

Closes #695. The way R parallel processing runs, sometimes the number of cores needs to be adjusted to prevent failures.

Add a customizable argument `ncores` to the pixel SOM assignment and pixel consensus clustering functions, which get passed to their respective R scripts.

**Remaining issues**

Is this a function we want to add for pixel data preprocessing and pixel meta cluster renaming?

### DIFF
--- a/ark/phenotyping/pixel_consensus_cluster.R
+++ b/ark/phenotyping/pixel_consensus_cluster.R
@@ -51,8 +51,11 @@ clustToMetaPath <- args[7]
 # retrieve the batch size to determine number of threads to run in parallel
 batchSize <- strtoi(args[8])
 
+# get the number of cores
+nCores <- strtoi(args[9]) - 1
+
 # set the random seed
-seed <- strtoi(args[9])
+seed <- strtoi(args[10])
 set.seed(seed)
 
 # read cluster averaged data
--- a/ark/phenotyping/pixel_consensus_cluster.R
+++ b/ark/phenotyping/pixel_consensus_cluster.R
@@ -52,7 +52,7 @@ clustToMetaPath <- args[7]
 batchSize <- strtoi(args[8])
 
 # get the number of cores
-nCores <- strtoi(args[9]) - 1
+nCores <- strtoi(args[9])
 
 # set the random seed
 seed <- strtoi(args[10])
--- a/ark/phenotyping/pixel_consensus_cluster.R
+++ b/ark/phenotyping/pixel_consensus_cluster.R
@@ -85,7 +85,20 @@ for (batchStart in seq(1, length(fovs), batchSize)) {
     # define the parallel cluster for this batch of fovs
     # NOTE: to prevent the occassional hanging first FOV issue, we need to log to an outfile
     # to "force" a return out of the foreach loop in this case
-    parallelCluster <- parallel::makeCluster(nCores, type="FORK", outfile='log.txt')
+    parallelStatus <- tryCatch(
+        {
+            parallelCluster <- parallel::makeCluster(nCores, type="FORK", outfile='log.txt')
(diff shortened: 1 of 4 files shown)


ngreenwald:
The main issue was the lack of error messages, right? We need to make sure the user actually knows to adjust this parameter

alex-l-kong:
@ngreenwald yeah, I'm working on that right now.

ark/phenotyping/pixel_consensus_cluster.R
```diff
@@ -112,23 +128,24 @@ for (batchStart in seq(1, length(fovs), batchSize)) {
...
-        data.frame(fov=fovs[i], status=status)
+        data.frame(fov=fovs[i], status=status[1], errCond=status[2])
     }
 
     # report any erroneous feather files
     for (i in 1:nrow(fovStatuses)) {
         if (fovStatuses[i, 'status'] == 1) {
-            print(paste("The data for FOV", fovStatuses[i, 'fov'], "has been corrupted, removing"))
+            print(paste("Processing for FOV", fovStatuses[i, 'fov'], "failed, removing from pipeline. Error message:"))
+            print(fovStatuses[i, 'errCond'])
```
ngreenwald, line 148: Were you able to find an easy way to test what happens when too many cores are specified, confirming that this now shows the correct error message?

ark/phenotyping/run_pixel_som.R
```diff
@@ -97,23 +110,24 @@ for (batchStart in seq(1, length(fovs), batchSize)) {
...
-        data.frame(fov=fovs[i], status=status)
+        data.frame(fov=fovs[i], status=status[1], errCond=status[2])
     }
 
     # report any erroneous feather files
     for (i in 1:nrow(fovStatuses)) {
         if (fovStatuses[i, 'status'] == 1) {
-            print(paste("The data for FOV", fovStatuses[i, 'fov'], "has been corrupted, removing"))
+            print(paste("Processing for FOV", fovStatuses[i, 'fov'], "failed, removing from pipeline. Error message:"))
+            print(fovStatuses[i, 'errCond'])
```
ngreenwald, line 130: Same here

ngreenwald (changes requested):
Looks good, see below

ark/phenotyping/pixel_consensus_cluster.R
alex-l-kong: @ngreenwald yes, set the number of cores to an extremely high value (say, 128 on a laptop) and it now displays the error message `"Too many cores specified"`.

ark/phenotyping/run_pixel_som.R
alex-l-kong: @ngreenwald same as above.

ngreenwald (approved):
Great, looks good

PR #706, opened 2022-09-13
Handle variable size images for Deepcell input generation

### LINKED ISSUE #701: Image size issue
I am wondering do I get this error because I have images with different fov sizes? 

Should I split up my fovs by size?

image

alex-l-kong:

Closes #701. If a user specifies a cohort with variable size images, `generate_deepcell_input` will fail because it attempts to batch images of different sizes into one `xarray`. Users should be able to generate this input regardless of the image sizes specified.

Because it was observed that batching doesn't improve speed over a per-FOV iteration, we'll change it to the latter. This will be a fail-safe against loading multiple variable-sized images.

### DIFF
--- a/ark/utils/data_utils.py
+++ b/ark/utils/data_utils.py
@@ -452,23 +452,24 @@ def generate_deepcell_input(data_dir, tiff_dir, nuc_channels, mem_channels, fovs
         # load the images in the current fov batch
         if is_mibitiff:
             data_xr = load_utils.load_imgs_from_mibitiff(
-                tiff_dir, mibitiff_files=[fov], channels=channels)
+                tiff_dir, mibitiff_files=[fov], channels=channels
+            )
         else:
             data_xr = load_utils.load_imgs_from_tree(
-                tiff_dir, img_sub_folder=img_sub_folder, fovs=[fov], channels=channels)
+                tiff_dir, img_sub_folder=img_sub_folder, fovs=[fov], channels=channels
+            )
 
-        # write each fov data to data_dir
-        for fov in data_xr.fovs.values:
-            out = np.zeros((2, data_xr.shape[1], data_xr.shape[2]), dtype=data_xr.dtype)
+        fov_name = data_xr.fovs.values[0]
+        out = np.zeros((2, data_xr.shape[1], data_xr.shape[2]), dtype=data_xr.dtype)
 
-            # sum over channels and add to output
-            if nuc_channels:
-                out[0] = np.sum(data_xr.loc[fov, :, :, nuc_channels].values, axis=2)
-            if mem_channels:
-                out[1] = np.sum(data_xr.loc[fov, :, :, mem_channels].values, axis=2)
+        # sum over channels and add to output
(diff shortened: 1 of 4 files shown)


alex-l-kong:
@ngreenwald let me work on removing the extra `for` loop inside, since we're doing a per-FOV iteration it can be condensed.

alex-l-kong:
@ngreenwald issue has been resolved for Deepcell input generation, however Ben is running into the same issue trying to generate cell tables. Do we want to make a similar change to `generate_cell_table`, or is there a workaround Ben can use in the meantime?

ark/utils/data_utils.py
```diff
@@ -432,6 +432,7 @@ def generate_deepcell_input(data_dir, tiff_dir, nuc_channels, mem_channels, fovs
             the number of fovs to process at once for each batch
         dtype (str/type):
             optional specifier of image type.  Overwritten with warning for float images
+
```
ngreenwald, line 435: We should remove batch_size from docstring and function arg

ngreenwald (changes requested):
Looks good, just need to update docstring. 

Will you open an issue listing all of the functions in the`ark` pipeline that have a batch_size? I responded to his issue, a workaround in the meantime is setting batch_size equal to one. However, given that many of these batch_sizes are just a batch over an inner for loop, like this function, there's no reason to have one at all. 

What would be useful is to make a census of what functions have batch_sizes, and also which ones are actually useful, and which don't have any performance improvements. Also which ones involve a call to `load_imgs_from_tree`, and hence will have issues for dissimilarly sized images, and which don't

alex-l-kong:
@ngreenwald looks like the additional following functions use an explicit `batch_size` argument without multiprocessing:

- `data_utils.generate_and_save_pixel_cluster_masks` (and by extension, `data_utils.generate_pixel_cluster_mask`)
- `data_utils.generate_and_save_cell_cluster_masks` (and by extension, `data_utils.label_cells_by_cluster`)
- `marker_quantification.generate_cell_table` (the one @benoberlton is currently having trouble with)
- `spatial_analysis.batch_channel_spatial_enrichment` and `spatial_analysis.batch_cluster_spatial_enrichment`: talked to @ackagel about this, we can condense these to a per-FOV basis with negligible speed difference

ngreenwald:
Okay great. You should confirm that this is the case, but I think the only thing the batch size is doing for any of these is controlling the nested for loop behavior, there isn't a meaningful speedup. If that's the case, all of these can be changed to just have a single loop over one fov at a time. Can be in a separate PR though

ngreenwald:
Sounds good, won't have time to do any code reviews till this weekend, but you can go ahead and move forward on the other stuff

ngreenwald (approved):
Looks good. Will you open an issue listing those batching functions, and also add `spatial_analysis.create_neighborhood_matrix` and `spatial_analysis_utils.calc_dist_matrix' functions. These don't currently batch, I think they read everything in at once. Take a look at Candace's open PR for how they work

PR #707, opened 2022-09-14
Added Multi-Arch Docker Support

### LINKED ISSUE #611: Support multi-architecture Docker builds to accommodate M1 chip Macs

The Docker image by default is built using `amd64` architecture, which was supported by the old Mac Intel chips. However, the M1 chip Macs use `arm64` architecture. This can cause warnings (and potential notebook issues) such as:

```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```

to pop up if an M1-user pulls `ark-analysis` from Dockerhub.

To be safe, we should build a multi-arch image that supports both architectures.

This link (https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) escribes two ways we can do this:

- Using `docker manifest`
- Using `buildx`

### LINKED ISSUE #704: datasets module doesn't seem to be installed
**Describe the bug**
In the current version on main, requirements.txt definitely contains the datasets module. However, when I start Docker, it looks like python can't find the datasets module. Not sure where requirements.txt is supposed to be run, but it doesn't seem like datasets is being installed.

image

**Expected behavior**
Install datasets (not sure if it's a bigger problem with requirements.txt)

**To Reproduce**
Pull the current version on main, try "import datasets" in a jupyter notebook.

srivarra:

Closes #611. Closes #704. Added multi-arch Docker image build capabilities. Adds the dataset requirement to the docker image.

Currently the Docker Image is only built against `linux/amd64`. Added the ability to build Apple Silicon native Docker Images with `linux/arm64`. This is done with `docker buildx bake` using a `docker-bake.hcl` file. In addition adjusted the `.travis.yml` file to account for the new building process. Also split the docker build step into  a multi-stage process for easier debugging.

Steps to build multi-arch images locally include:
1. Create a multi-arch builder: `docker buildx create --name arkbuilder --use`
2. Build:
   1. With custom tag: `TAG=<your-tag> docker buildx bake build`
   2. With default tag: `docker buildx bake build`

Bumped the version from `0.4.1` to `0.4.2`.

**Remaining issues**

Will need to create a new release after this.


ngreenwald (changes requested):
So this allows the user to locally build a compatible version, but the default version will be the one we've been using? Or this also modifies what's hosted on dockerhub?

ngreenwald (changes requested):
Got it, makes sense. I'll defer to @alex-l-kong and @ackagel on the code stuff. Will you make sure the machine-specific architectures are downloaded correctly and work on an M1/Non-M1 machine?

Dockerfile
```diff
@@ -1,29 +1,37 @@
-FROM python:3.8
+# Stage 1: Start from the official Python-3.8 image: https://hub.docker.com/_/python/
+FROM python:3.8 AS base
```
alex-l-kong, line 2: Love the addition of multistage Docker builds. This will help us out with debugging in case a step goes wrong.

Dockerfile
```diff
@@ -1,29 +1,37 @@
...
 
-# install gcc and R
-RUN apt-get install -y gcc r-base
+# install R
+ARG rversion=4.0.4-1
+RUN apt-get install -y \
+    r-base-core=${rversion} \
+    r-base-dev=${rversion} \
+    r-base-html=${rversion} \
+    r-doc-html=${rversion}
```
alex-l-kong, line 26: Does the Python 3.8 image require us to explicitly specify all of these now?

alex-l-kong (commented):
Looks good, just one clarification about R installation.

Dockerfile
srivarra: Yeah, for native M1 support I had to do this. If I just did `apt-get install -y r-base` it would always fail stating that there was a version mismatch for the requirements of `r-base`.

ngreenwald:
Leaving the issue open that this is addressing, since it's unresolved, but will require starting from scratch anyway so a fresh PR makes sense

PR #723, opened 2022-09-22
Support rectangular images for pixel mask generation

### LINKED ISSUE #719: ark repo assumes square images
After successfully running all the preprocessing and clustering steps, I am running into an error in Jupyter notebook on the step of `generate_and_save_pixel_cluster_masks`. Upon further inspection, I notice that the dimensions of the TIFF files (X * Y = 447,897,600 pixels) are different from the dimensions of the fov.feather file after the `create_pixel_matrix` step (len = 297,923,080). On a smaller image sample, there is still a difference but it is smaller (TIFF X * Y = 74,207,070, and len(feather) = 74,132,387). 

This discrepancy does not happen when we use the very small example data set provided (1,048,576 pixels), but seems to occur on our larger datasets. 

Below is the error message obtained in the Jupyter notebook. Although the pipeline removes empty pixels, there seems to be issues with the dimensions later on. It seems like as the images get larger, they are truncated and various problems begin to occur that I can't quite figure out. I know I wrote previously that I did some runs in conda, but this error occurred when strictly using Jupyter notebooks in the singularity container as is

Your help with this is very much appreciated! 

image

alex-l-kong:

Closes #719. The pixel mask generation function assumes square MIBI images, however we're now seeing cohorts being run on rectangular images that will make this assumption fail. This is a problem for pixel mask generation in particular, as it's flattening and reindexing method will throw an error under.

We need to change the indexing multiplier to be the number of columns. Previously, we could assume the number of rows equalled the number of columns, so the row axis size was used for convenience sake of accessing the first available axis. 

**Remaining issues**

@jonhsussman may have additional input.

### DIFF
--- a/ark/utils/data_utils.py
+++ b/ark/utils/data_utils.py
@@ -248,7 +248,7 @@ def generate_pixel_cluster_mask(fovs, base_dir, tiff_dir, chan_file,
         y_coords = fov_data['column_index'].values
 
         # convert to 1D indexing
-        coordinates = x_coords * img_data.shape[1] + y_coords
+        coordinates = x_coords * img_data.shape[2] + y_coords
 
         # get the cooresponding cluster labels for each pixel
         cluster_labels = list(fov_data[pixel_cluster_col])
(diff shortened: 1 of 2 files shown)


alex-l-kong:
@jonhsussman while I wait for a review on this branch, you can test it out.

jonhsussman:
Excellent, thanks!

ngreenwald (approved):
Looks good

Issue #720, opened 2022-09-21
Replace batching functionality of `ark` pipeline to per-FOV looping


alex-l-kong:
*This is for internal use only; if you'd like to open an issue or request a new feature, please open a bug or enhancement issue*

### Instructions
This document should be filled out prior to embarking on any project that will take more than a couple hours to complete. The goal is to make sure that everyone is on the same page for the functionality and requirements of new features. Therefore, it's important that this is detailed enough to catch any misunderstandings beforehand. For larger projects, it can be useful to first give a high-level sketch, and then go back and fill in the details. For smaller ones, filling the entire thing out at once can be sufficient.  

**Relevant background**

Several lab members have started running cohorts containing images of varying sizes, which will cause the calls to `load_imgs_from_tree` to fail. This is because the batching functionality defaults to loading 5 images into a fixed dimensional `xarray`; this batching does not provide any significant optimization because we eventually use a per-FOV loop anyways, so it should be removed.

**Design overview**

Any function that explicitly or implicitly relies on a `batch_size` argument will need to be modified so it only processes a single FOV and removes any loops over batches. Additionally, notebooks which make calls to these functions will need to be updated.

This only applies to functions where `batch_size` doesn't remove a per-FOV loop used later on. Functions that use batching for multiprocessing are exempt.

**Code mockup**

The functions needing modification are mostly self-contained. They are sorted by appearance on the pipeline and should be completed in this order:

- [X] Cell table generation (`marker_quantification.generate_cell_table`)
- [x] Pixel/cell cluster mask generation (these two can be addressed in one PR)
    - [x] `data_utils.generate_and_save_pixel_cluster_masks` and `data_utils.generate_pixel_cluster_mask`
    - [x] `data_utils.generate_and_save_cell_cluster_masks` and `data_utils.label_cells_by_cluster`
- [x] Post-cluster cell labeling and FOV saving (these can be addressed in one PR)
    - [x] `data_utils.create_mantis_project`
    - [x] `data_utils.save_fov_images`
    - [x] `data_utils.label_cells_by_cluster`
        - This will require refactoring of the neighborhood analysis notebook too
- [x] Spatial enrichment functions (these two can be addressed in one PR)
    - [x] `spatial_analysis.batch_channel_spatial_enrichment`
    - [x] `spatial_analysis.batch_cluster_spatial_enrichment`
[...]

PR #726, opened 2022-09-24
Remove cell table batching functionality

alex-l-kong:

Addresses stage 1 of #720, removing batching of `generate_cell_table`.

The batch loop inside `generate_cell_table` has been changed to a per-FOV iteration.

The call in `1_Segment_Image_Data.ipynb` needed to be updated accordingly.

**Remaining issues**

In particular, changes were made to the way `fovs` and `filenames` were being handled in `generate_cell_table`. MIBItiff support requires us to retrieve `filenames`, however for non-MIBItiffs the loop needs to be more fully guarded against `filenames` being set to `None` (which happens when you call `list_files` on a folder structure).

We will not have to worry about this once MIBItiff support gets phased out.


alex-l-kong:
@benoberlton this PR should address the error caused by variable image sizes you were getting during `generate_cell_table`. Let me know if any issues pop up.

alex-l-kong:
@ngreenwald realized I needed to update the documentation for `create_marker_count_matrices` to indicate that the FOV dimensions should support just a single FOV, however seems I can't push the change until you review the rest of the changes (even if I remove the review request). If all looks good I'll push that change separately.

ngreenwald (changes requested):
Huh that's strange, I didn't know review status could block subsequent changes. Won't have time to look tonight, I'll submit this which will hopefully address the issue

ngreenwald (changes requested):
Looks good, just need to add back in the code that was removed. Seems like this has happened a bunch of times on your PRs, I think you said your editor does it automatically?

alex-l-kong:
> Looks good, just need to add back in the code that was removed. Seems like this has happened a bunch of times on your PRs, I think you said your editor does it automatically?

@ngreenwald yeah, looks like I'll need to officially need to retire my merge tool lol, Sublime Merge hasn't been the easiest to interface with.

alex-l-kong:
@ngreenwald this should resolve `generate_cell_table` with `main`.

ngreenwald (approved):

PR #728, opened 2022-09-27
validate_paths data prefix

### LINKED ISSUE #676: Appropriate behavior for `data_prefix=True` in `validate_paths`

**Describe the bug**
The `validate_paths` function doesn't fail appropriately when supplied a valid path which doesn't begin with '../data'.

**Expected behavior**

The original point of the `validate_paths` function was to ensure that the user not only had a correct path, but also that the path pointed to data contained within '.../data', which is the format for mounting data we use in the docker. 

The point of having the `data_prefix` argument was that the function would fail if the path didn't start with `../data`, because this would result in a docker error otherwise. So even if the path was valid, the fact that it didn't start with data_prefix made it invalid. 

However, the current version of the function with `data_prefix=True` only fails if the path is actually invalid. It will then have separate logic about `data` being missing. However, this kind of defeats the purpose  of the `data_prefix` argument. 

Given that people are often mounting external volumes, in which case the path will never begin with '.../data', perhaps we should deprecate this functionality instead of fixing it? Either way, function should be updated. 

Open to suggestions @angelolab/computational_team

camisowers:

Closes #676. 

Check for data prefix first if specified and change default to `data_prefix=False`. The instances of `validate_paths` in the notebook use the default which now doesn't require `..\data`.

**Remaining issues**

After this PR is merged I'll check throughout toffy for appropriate behavior using the False default.


ngreenwald (changes requested):
Do any of the notebooks use this function with data_prefix set to True? Or other functions in ark?

camisowers:
Nope! The notebooks just use the default and all the ark functions have `data_prefix=False`.

ngreenwald (approved):

PR #733, opened 2022-09-28
Speed up `relabel_segmentation`

### LINKED ISSUE #722: generate_and_save_cell_cluster_masks not completing
I have successfully ran through the pixel clustering notebook and the cell clustering notebook (including the heatmaps) up until the `generate_and_save_cell_cluster_masks` step. On this step, it seems to hang indefinitely for hours, not producing an error. It seems unlikely that it will take this long given the analogous step on the pixel cluster notebook completed in < 15 minutes. This is the case both when I use a conda environment as well as when I use the Jupyter notebook. 

Is this particular step known to take a very long time, or could there be some error here, for larger fov images? 

image

alex-l-kong:

Closes #722. With very large images with a high number of cells, `relabel_segmentation` will take hours to complete because of the way boolean indexing works. Additionally, very large 2-D NumPy arrays can cause cache misses due to the way they're non-contiguously stored in memory.

Using `np.vectorize` provides a significant speed up.

**Remaining issues**

@jonhsussman can you check out this branch and verify that it runs efficiently? When I timed your looped version versus the `np.vectorize` version, the latter ran about 2-2.5x faster, but that was on test data (we unfortunately don't have images of your size on hand at our lab). I can convert this back to a looped version if you find the opposite to be true.

### DIFF
--- a/ark/utils/data_utils.py
+++ b/ark/utils/data_utils.py
@@ -7,7 +7,9 @@ from typing import List, Union
 import datasets
 import feather
 import numpy as np
+import pandas as pd
 import skimage.io as io
+from skimage.measure import regionprops_table
 import xarray as xr
 from tqdm.notebook import tqdm_notebook as tqdm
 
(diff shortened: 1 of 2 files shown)


ngreenwald (changes requested):
It should be pretty easy to simulate a large image: just take a single 1024x1024 example image, and stitch it into a larger image (say 5x5 or 10x10) by duplicating the original image, incrementing the values of the segmentation labels by the previous max to generate unique vals

alex-l-kong:
@ngreenwald already tested it out a few random 20480x20480 images:

- The `np.vectorize` version (in this PR right now) took about 3 minutes for one image
- The `for` loop version (as described by @jonhsussman) takes about 7.25 minutes for one image
- The original method hangs for a very long time

With images this big, relabeling is going to be an expensive process even with the optimization. If we can safely assume that users will never generate more than a few masks for visualization purposes, then I think it should be fine. However, if a significant number of users want to visualize several FOVs, we may need to further streamline this (ex. an option to multiprocess).

ngreenwald:
Agreed, I still think it's worth testing out an image around 10k x 10k to see what the scaling behavior looks like for the two different solutions. Yours is simpler, so if it retains good performance at the larger image size then we should keep it.

alex-l-kong:
@ngreenwald just want to confirm if you meant ≈10000x10000 pixels, or stitch a sample 1024 x 1024 image into a 10000x10000 grid.

ngreenwald:
Hahaha, definitely the former, i.e. a 10x10 grid of the current image.

alex-l-kong:
@ngreenwald for a ≈10000x10000 image, it takes about 30 seconds to run with `np.vectorize`, and about 75 seconds to run with a `for` loop.

There's definitely a non-linear cost to the process as the image size increases, as a ≈20000x20000 image takes nearly 6 times longer.

ngreenwald (approved):
Great, thanks for investigating

PR #737, opened 2022-09-29
Rework mask saving and cell labeling to support just single FOV parameter

alex-l-kong:

Addresses part 3 of #720.

Main work involves condensing `save_fov_images` and `label_cells_by_cluster` to support just a single `fov` argument, as opposed to a list of `fovs` to support batching removal.

This required modification of mask generation and saving of neighborhood clusters. A new function, `generate_neighborhood_cluster_masks`, was created to streamline this.

**Remaining issues**

- [x] Test function still needs to be added for `generate_neighborhood_cluster_masks`
- [x] Verify that clustering notebooks can run seamlessly through without error


ark/utils/data_utils.py
```diff
@@ -34,16 +35,12 @@ def save_fov_images(fovs, data_dir, img_xr, sub_dir=None, name_suffix=''):
...
     if not os.path.exists(data_dir):
         raise FileNotFoundError("data_dir %s does not exist" % data_dir)
 
-    # verify that the fovs provided are valid
-    verify_in_list(
-        provided_fovs=fovs,
-        img_xr_fovs=img_xr.fovs.values
-    )
+    # convert image data to int16
+    img_data = img_data.astype('int16')
```
ngreenwald, line 43: What functions call this function? I'm surprised this hasn't caused issues with `float` image data.

ark/utils/data_utils.py
```diff
@@ -55,60 +52,82 @@ def save_fov_images(fovs, data_dir, img_xr, sub_dir=None, name_suffix=''):
...
-        # define the file name as the fov name with the name suffix appended
-        fov_file = fov + name_suffix + '.tiff'
+    img = np.copy(labeled_image)
+    unique_cell_ids = np.unique(labeled_image)
+    unique_cell_ids = unique_cell_ids[np.nonzero(unique_cell_ids)]
 
-        # save the image to data_dir
-        io.imsave(os.path.join(save_dir, fov_file), fov_img_data, check_contrast=False)
+    default_label = max(labels_dict.values()) + 1
+    for cell_id in unique_cell_ids:
```
ngreenwald, line 82: Isn't this doing the same thing as your optimization PR? Can this be combined with that newly optimized function?

ngreenwald (changes requested):
Looks good so far

ark/utils/data_utils.py
alex-l-kong: @ngreenwald yeah I didn't get the chance to merge the changes in yet.

ark/utils/data_utils.py
alex-l-kong: @ngreenwald currently just the mask generation functions as well as the mantis project creation function.

ngreenwald: Will you check to see what happens when you use this with the example dataset? My assumption is that it will ruin all of the float images, but maybe I'm missing something. 

Will you also do a search through all of `ark` to find all of the places where we change the dtype of the data, to see if there are any others?

alex-l-kong: @ngreenwald to address the first question:

Pixel masks get returned as a `'float32'`, which I think was the main motivation behind the `'int16'` conversion. Upon further inspection this happens because `np.zeros` automatically initializes to `float` in the pixel mask generation function. By specifying an explicit `dtype` during that step we should be able to fix this.

We can likewise fix this for neighborhood mask generation, which currently returns `'int64'` images, by explicitly casting the `label_map` passed into `label_cells_by_cluster` to `'int16'`.

Both of these patches should remove the need for an `'int16'` conversion directly in `save_fov_images`.

ngreenwald: I was confused, I thought this was image data generated by the MIBI, and hence floats. We should update the names to make it more clear what is being saved, but now it makes sense why this wasn't breaking anything

alex-l-kong: `save_fov_images` is only working with pixel, cell, and neighborhood masks, not the actual MIBI images. These should only store `int` types because these cluster labels can only be integers. As you mentioned, I'll be updating the name of this function to `save_fov_masks`.

alex-l-kong: @ngreenwald to address the second question, `.astype` is popping up most frequently in the `spatial_analysis_utils` module, which utilizes a bunch of cases which I think are used to save space/increase efficiency (will check in with Adam about this).

Casting is also frequently done after a call to `find_boundaries`, changing the data to unsigned 8-bit integer type.

Conversions are also happening in pixel and cell clustering utility functions for various functions, such as ensuring Gaussian blurring produces float values, cluster labels are of integer type, and renamed clusters get cast as `str` in case the user doesn't change their names.

ngreenwald:
Looks good, will wait to merge in until we confirm that the -u flag is now working.

ngreenwald (approved):

PR #741, opened 2022-10-04
Settings name update

### LINKED ISSUE #712: Change defaults in settings.py
Some of the variables in settings.py are misaligned with what we usually use in our datasets/what the notebooks are generating. For example, `FOV_ID` is currently 'SampleID' - should probably be 'fov'. `KMEANS_CLUSTER` is currently 'cell_meta_cluster_rename' which doesn't make sense.

Go through settings.py and set the defaults to match the example dataset/what we usually use.

**Additional context**
I tried to change some of these when I was updating the neighborhood notebooks but it broke a bunch of tests so I didn't bother. I don't think the tests should be checking for hard-coded variables - alas, when I reverted settings.py all the tests passed.

### LINKED ISSUE #713: Spatial scripts require a `CLUSTER_ID` column, seems redundant with `CELL_TYPE`
The spatial enrichment notebook and spatial LDA notebook require that the cell table has a `CLUSTER_ID` column, which should be integers representing unique cell types. The cell table already has a `CELL_TYPE` column, which corresponds to the cell phenotype names. These seem redundant to me. `CLUSTER_ID` and `CELL_TYPE` are referenced in `settings.py`. Right now it's confusing for users who are going from cell clustering to the spatial scripts, since there is no `CLUSTER_ID` column in the cell table.

I think we need to either add functionality to the end of the cell clustering notebook to append this `CLUSTER_ID` column to the cell table (i.e. give each phenotype a unique integer). Or edit the spatial scripts to not use the `CLUSTER_ID` column and only use the `CELL_TYPE` column (since the information is redundant anyways). Option 1 is probably easier.

camisowers:

Closes #712 and closes #713.

`FOV_ID = 'SampleID'`  ->  `FOV_ID = 'fov'`

`CELL_TYPE = 'cell_type'`  ->  `CELL_TYPE = 'cell_meta_cluster'`

`CLUSTER_ID = cell_meta_cluster`  ->  `CELL_TYPE_NUM = 'cell_num'`

`KMEANS_CLUSTER = ' cell_meta_cluster_rename'`  ->  `KMEANS_CLUSTER = 'kmeans_neighborhood'`

Changes the names to be more clear, and also removes or adjusts the outdated cell table columns. 
Fixed the spatial LDA pre-processing notebook, which was incorrectly using `KMEANS_CLUSTER` instead of `CELL_TYPE`. Individual functions now generate the `CELL_NUM` column, rather than requiring it in the cell table; I was able to completely remove the need for this column in the LDA code, but spatial enrichment still uses it. 

Note: After removing the integer cell cluster id, the LDA notebooks now show the cluster names in the visualizations rather than an integer for each cluster.

Screen Shot 2022-09-30 at 3 16 39 PMScreen Shot 2022-09-30 at 3 16 46 PM

**Remaining issues**

spatial notebooks all run without error, but need to double check the results and logic


ark/utils/spatial_analysis_utils_test.py
```diff
@@ -111,20 +111,21 @@ def test_get_pos_cell_labels_channel():
...
+    all_data[settings.CELL_TYPE_NUM] = list(all_data[settings.CELL_TYPE].
+                                            astype('category').cat.codes)
     excluded_channels = [0, 13, 22]
 
     # Subsets the expression matrix to only have channel columns
     channel_start = np.where(all_data.columns == settings.PRE_CHANNEL_COL)[0][0] + 1
     channel_end = np.where(all_data.columns == settings.POST_CHANNEL_COL)[0][0]
 
-    fov_channel_data = all_data.iloc[:, list(range(channel_start, channel_end + 1)) + [31]]
+    fov_channel_data = all_data.iloc[:, list(range(channel_start, channel_end + 1)) + [32]]
```
ngreenwald, line 122: Why do we have this magic number, and why did it change by 1?

ngreenwald (changes requested):
Looks good

templates/LDA_Preprocessing.ipynb
```diff
@@ -3,7 +3,6 @@
   {
```
ngreenwald, line 3: <pre class="ql-syntax" spellcheck="false">Line #21.    # or use manually specified cluster names
</pre><p>Will you make sure to run all of the notebooks to confirm everything is working? The spatial_LDA ones obviously, but also kmeans and enrichment analysis if those were affected at all. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/741/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

ark/utils/spatial_analysis_utils_test.py
camisowers: Great question that also confused me with a lot of failing tests. 
The test_utils function `_make_expression_mat_sa()` that generates an example expression matrix for testing expected the table to have the int cell cluster identifier and made it the 31st column in the data frame. When I deleted that from the list of expected cols, the cell cluster name became column 31 and the int identifiers that were generated and appended in the individual functions became column 32.

camisowers:
Running the kmeans and spatial enrichment notebooks gives essentially the same output as before.
The LDA notebooks have very different results now since it was using the wrong cell table column before, but it all seems to be running without issue now.

I also updated the LDA preprocessing notebook cell showing the `settings.BASE_COLS` output to have the new names.

srivarra:
Does the cell table need to be recreated to be compatible for any of the settings name changes?

camisowers:
> @camisowers Does the cell table need to be recreated to be compatible for any of the settings name changes?

Nope! The settings changes should now reflect the current cell table columns.

ngreenwald (changes requested):
Looks good, I'll let @cliu72 confirm and then it'll be good to go

templates/example_pairwise_spatial_enrichment.ipynb
```diff
@@ -224,7 +224,7 @@
    "source": [
```
cliu72, line 224: <pre class="ql-syntax" spellcheck="false">Line #3.    pheno_titles = all_data["cell_meta_cluster"].drop_duplicates()
</pre><p>I think it would be a good idea to refer to this as "settings.CELL_TYPE" instead of having the user hard-code it in. Same thing earlier in the notebook, when "segmentation_label" is hard-coded in. Probably better to have that as "settings.CELL_LABEL".</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/741/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

cliu72:
Looks good to me. Just one small comment about the spatial enrichment notebook that is a super easy fix.

templates/example_pairwise_spatial_enrichment.ipynb
camisowers: 'segmentation_label' references the existing column name in the array for the segmented tiff images so that one is necessary, but I the you're totally right that 'cell_meta_cluster' should not be hard-coded!

cliu72 (approved):

PR #742, opened 2022-10-04
Use existing Docker container when possible to prevent redundancy

### LINKED ISSUE #732: Prevent `start_docker.sh` from creating a new container each time it's run

Whenever `start_docker.sh` is run, a new Docker container is created. This because the `docker run` command is an alias for `docker create` and `docker start`. This can create a lot of "Docker clutter" where the user could unknowingly have a ton of redundant containers on their machine. It can also make things difficult for debugging in case someone needs to test the image locally with new dependencies.

We should modify `start_docker.sh` so that it uses `docker start` if the user already has the most up-to-date container on their machine. In this way, it's just going to reuse the existing Docker image as opposed to downloading a redundant one each time.

alex-l-kong:

Closes #732. Each time `start_docker.sh` is run a new container is created, which can lead to several large, redundant Docker containers on a user's computer. We should instead use an existing container with a standardized name whenever possible.

This PR requires the following:

- Each Docker container should now be tagged with a specific name, currently set to the Docker image release. The `VERSION` parameter  in `start_docker.sh` needs to be updated when a new image is released to Dockerhub. This name will help us easily identify if the Docker container exists on the user's computer.
- If the container is found on the machine, default to the original `docker run` command, but if it is, then simply use `docker start`
    - The exception is if the user attaches an external volume. We have no choice but to delete the container and rebuild it, since Docker won't allow you to change the attached volumes on an existing container.

**Remaining issues**

Haven't tried this on Windows yet, will need some external testers to assist. UPDATE: tested it on Sri's computer, output redirection works now.


start_docker.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# define the version number, this needs to be updated every new Docker release
+VERSION='v0.4.2'
```
ngreenwald, line 4: Will you add this to the developer docs, there's a list of what needs to be changed for each release. Also see if anything else is missing from there

start_docker.sh
```diff
@@ -68,4 +73,17 @@ run_params=(
...
+if [[ ! -z "$external" ]]
+  then
+    # this prevents script from failing in case the container does not already exist
+    docker rm -f $VERSION > /dev/null 2>&1 || true
+fi
+
+# if no Docker container found named $VERSION, create it, otherwise boot it up
+if [[ $(docker ps -a --format "{{.Names}}" | grep $VERSION | wc -l) -eq 0 ]]
+  then
+    docker run -it "${run_params[@]}" --name $VERSION angelolab/ark-analysis:$VERSION
```
ngreenwald, line 86: Will you test all of this logic to make sure it works?
And test it in combination with having made a change using the -u flag previously, to make sure that's carried forward?

ngreenwald (changes requested):
See below for small tweaks

alex-l-kong:
@ngreenwald on second thought, it's easiest to just remove the Docker image and recreate it each time. It doesn't cause a lot of overhead and it makes the underlying logic much easier to follow.

start_docker.sh
alex-l-kong: Updated this logic to recreate the Docker image each time, verified this works on the `-u` flag.

ngreenwald (commented):
Looks good, I'll wait for @srivarra to comment before merging

srivarra (approved):
The script does indeed delete the most recently used container, and create a new one.

ngreenwald (approved):

ngreenwald:
@alex-l-kong will you make a new release with a brief summary of what got changed

PR #752, opened 2022-10-06
Remove `batch_size` parameter from spatial analysis functions

alex-l-kong:

Addresses part 4 of #720.

Remove `batch_size` as a parameter from both `batch_channel_spatial_enrichment` and `batch_cluster_spatial_enrichment`.

Since these functions are no longer being batched, their names were also changed from `batch_channel_spatial_enrichment` to `generate_channel_spatial_enrichment_stats`, and likewise for cluster enrichment.


ngreenwald (approved):
Looks good!

PR #758, opened 2022-10-07
Add inertia and silhouette score progress bars for neighborhood analysis notebook

### LINKED ISSUE #566: Progress updates for compute intensive functions

alex-l-kong:

Closes #566. The inertia and silhouette score computations take a while to complete, so they should have progress bars.

These should be the last progress bars that need to be added to `ark`.

Use `tqdm.notebook` as implemented in other modules.


ngreenwald (changes requested):
Did you run both notebook cells to make sure formatting on TQDM displays as expected?

alex-l-kong:
> Did you run both notebook cells to make sure formatting on TQDM displays as expected?

They are displaying correctly.

PR #764, opened 2022-10-10
Process distance matrices inside the neighborhood counts function

alex-l-kong:

Addresses #720. Neighborhood analysis has an inefficiency where the distance matrices are all precomputed beforehand. This is unnecessary as we can just compute these as needed in the neighborhood analysis per-FOV loop.

We remove any precomputing of distance matrices (including preloading of label maps) inside the notebooks. In addition to the neighborhood analysis notebook, there was one that existed in the spatial analysis notebook which needed addressing.

`compute_neighborhood_counts` now takes in similar arguments as the other spatial functions to load directly from `label_dir` and computing `dist_mats` on an as-needed per-FOV basis.

### DIFF
--- a/ark/analysis/spatial_analysis.py
+++ b/ark/analysis/spatial_analysis.py
@@ -439,15 +439,23 @@ def calculate_cluster_spatial_enrichment(all_data, dist_matrices_dict, included_
     return values, stats
 
 
-def create_neighborhood_matrix(all_data, dist_matrices_dict, included_fovs=None, distlim=50,
+def create_neighborhood_matrix(label_dir, all_data, suffix='_feature_0',
+                               xr_channel_name='segmentation_label',
+                               included_fovs=None, distlim=50,
                                self_neighbor=False, fov_col=settings.FOV_ID,
                                cell_label_col=settings.CELL_LABEL,
                                cluster_name_col=settings.CELL_TYPE):
     """Calculates the number of neighbor phenotypes for each cell.
 
     Args:
+        label_dir (str):
+            Directory containing labeled tiffs.
         all_data (pandas.DataFrame):
             data for all fovs. Includes the columns for fov, label, and cell phenotype.
+        suffix (str):
+            Suffix for tif file names.
+        xr_channel_name (str):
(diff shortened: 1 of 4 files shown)


ngreenwald:
I'm not sure I follow. What does pre-computing distances matrices have to do with batching?

alex-l-kong:
> I'm not sure I follow. What does pre-computing distances matrices have to do with batching?

`create_dist_matrices` was written to make it easy to integrate with the batching functionality of the spatial analyses functions, namely the per-FOV loop. We shouldn't need to do this anymore now that the spatial functions themselves only operate on a per-FOV basis. Moving the distance matrix computations to the neighborhood matrix function (as opposed to pre-computing them all in the notebook) itself will allow us to start doing this.

I was originally intending to address both issues in this PR, however, because `create_dist_matrices` was used in all three spatial analysis functions, I think it's best to address this separately.

ngreenwald:
Okay got it. Given that all of the spatial analysis functions use the same distance matrices, I don't know if I would agree that it's inefficient to precompute them. Is there a benefit from not pre-computing? How much time does it take compared to the other tasks in the notebooks?

alex-l-kong:
@ngreenwald based on our meeting today, conducted some timing tests on the `calc_dist_matrix` function and it can be a fairly expensive function if the image contains many different cell labels and the cohort contains several FOVs.

For 1000 different cell labels, it takes about 3 seconds per FOV. For 2000, about 6-7 seconds. For 5000, about 16.5 seconds. In a cohort with many FOVs, this can easily take a very long time. This is the case for both `regionprops` and `regionprops_table` with only the subset of needed properties specified.

Based on this, I think it's probably best to leave the distance matrices pre-computed prior to running them the spatial analyses. Using the `save_path` argument in `calc_dist_matrix`, we can save a `.npz` file that can be loaded up with the required data each time.

If we move forward with this, we should close this PR as the changes made here will no longer be relevant.

ngreenwald:
Got it, sounds good

alex-l-kong:
We will address this a different way per #770.

PR #776, opened 2022-10-18
Run `pytest` as a separate job to avoid `scripts` from getting overwritten

alex-l-kong:

Fixes a bug that can pop up when a separate `scripts` gets specified in a `job`. This can overwrite the global `script`, which in effect skips the `pytest` runs in the PR.

It's also better to separate `pytest` out into its own job, as it shouldn't be included with the deployment stages. This can help us isolate the `pytest` logs from the deployment logs.

Include a separate `pytest_run` stage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,13 +22,13 @@ env:
 
 cache: pip
 
-# command to run tests
-script:
-  - python -m pip install --editable .
-  - python -m pytest --cov=ark --pycodestyle ark
-
 jobs:
   include:
+    - stage: pytest_run
+      python: 3.8
+      script:
+        - python -m pip install --editable .
+        - python -m pytest --cov=ark --pycodestyle ark
     - stage: test_pypi_deploy
       python: 3.8
       script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,15 +22,13 @@ env:
...
-  - python -m pytest --cov=ark --pycodestyle ark
-
 jobs:
   include:
+    - stage: pytest_run
+      script:
+        - python -m pip install --editable .
+        - python -m pytest --cov=ark --pycodestyle ark
     - stage: test_pypi_deploy
-      python: 3.8
```
ngreenwald, line 33: Any reason we don't need 3.8 anymore?

ngreenwald (changes requested):

.travis.yml
alex-l-kong: @ngreenwald setting `python: 3.8` at the beginning of `.travis.yml` already ensures Python 3.8 is used across all the jobs, so no need to specify it individually in each job.

ngreenwald (approved):

PR #779, opened 2022-10-19
Make non-parallel loops canon for Pixie

### LINKED ISSUE #771: Remove parallel processing from pixie
The parallel processing implemented in Pixie has in general been hard to maintain, and doesn't play nice across OS. 

Remove the parallel processing functionality in favor of slower, but simpler, code which runs each FOV in series.

alex-l-kong:

Closes #771. Adds option to run a non-parallel version of Pixie preprocessing, SOM and meta cluster assignment, and meta cluster renaming.

Add control flow statements handled by a dedicated `multiprocess` argument.

### DIFF
--- a/templates/2_Cluster_Pixels.ipynb
+++ b/templates/2_Cluster_Pixels.ipynb
@@ -127,7 +127,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Turning on multiprocessing provides a speed boost, however it is not always  cross-platform compatible."
+    "Turning on multiprocessing provides a speed boost, however it is not always cross-platform compatible. If you receive errors such as hanging cells without progress updates, `ncores` set to `NA`, or `fovStatuses` of length 0, try setting `multiprocess` back to `False`."
    ]
   },
   {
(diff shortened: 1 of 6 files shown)


ngreenwald (changes requested):
Looks good. Will you run the pixel clustering notebook with and without multiprocessing to make sure nothing slipped through? The python code is all tested, but the R code isn't

templates/2_Cluster_Pixels.ipynb
```diff
@@ -116,6 +116,37 @@
     "# fovs = [\"fov14\"]"
```
ngreenwald, line 116: <p>I would add a comment about the type of error they might get when using multiprocessing so that as a first troubleshooting step, they can try switching this back. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/779/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

alex-l-kong:
@ngreenwald this works with `multiprocess` set to both should we add a `multiprocess` and `

> Looks good. Will you run the pixel clustering notebook with and without multiprocessing to make sure nothing slipped through? The python code is all tested, but the R code isn't

Already tested.

ngreenwald (approved):
Looks good

PR #791, opened 2022-10-26
Testbook Adjustments

### LINKED ISSUE #765: Jupyter Notebook Testing Improvements

### Instructions

**Relevant background**

The Jupyter Notebook tests are currently rather inflexible and complex. It makes adjusting the notebooks themselves quite difficult as the tests also need to be updated in conjunction.

**Design overview**

Utilize a class based testing [implementation](https://github.com/angelolab/ark-analysis/discussions/738#discussioncomment-3766825) with one class per notebook. `pytest` has special `_setup()` functions which behave like `__init__()` methods in regular python classes, therefore we can setup any variables necessary for each notebook (like precomputed data, etc...). For each class we should have a function per tagged Jupyter Notebook cell, which runs that cell.

If we utilize [`pytest-order`](https://github.com/pytest-dev/pytest-order) we can order the tests the same way as the user would run the cells.

In addition we can create a temporary path utilized by the all the notebook tests, so all of their outputs can be sent there, and discarded once the tests are finished.

Here are some guidelines:

1. We should test the jupyter notebooks for completeness (i.e. the notebooks run without crashing / halting)
2. Only create a small testing dataset if necessary.
3. Use `testbook.inject` sparingly

**Code mockup**

```python
import pytest
from testbook import testbook
from typing import Iterator
import pathlib
import os

# Sets the notebook testing session directory. Saves all notebook related files
# in a temporary directory.

@pytest.fixture(scope="session")
def base_dir_generator(tmp_path_factory) -> Iterator[pathlib.Path]:
    """
    A Fixture which creates the directory where the all notebook test inputs and outputs are
    saved.
    Sets the path once for all notebook related tests.

    Args:
        tmp_path_factory (pytest.TempPathFactory): Factory for temporary directories under the
            common base temp directory.

    Yields:
        Iterator[pathlib.Path]: The iterable path containing the path of the `notebook_tests`
        inputs and outputs.
    """
    notebook_test_output_dir = tmp_path_factory.mktemp("notebook_tests")
    yield notebook_test_output_dir

SEGMENT_IMAGE_DATA_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                                       '..', '..', 'templates', '1_Segment_Image_Data.ipynb')

@testbook(SEGMENT_IMAGE_DATA_PATH, timeout=60000, execute=False)
class Test_1_Segment_Image_Data(tb, base_dir_generator):
    @pytest.fixture(autouse=True)
    def _setup(self):
        """
        Sets up necessary data and paths to run the notebooks.
        """
        # Create any simulated / fake / testing data necessary here
        self._image_data = ... # Create the tiffs
        self._paths = ... # Set any paths relative to the `setup_temp_path_factory` (a.k.a base_dir)

    @pytest.mark.order(0)
    def test_imports(self):
        self.tb.execute_cell("import")
    
    @pytest.mark.order(1)
    def test_create_dirs(self):
        self.tb.execute_cell("create_dirs")
    
    # Skip downloading the example dataset cell
    @pytest.mark.skip(reason="Either don't download the example dataset (therefore generating all data necessary), or create a small dataset used only for notebook testing.")
    @pytest.mark.order(2)
    def test_download_example_dataset(self):
        self.tb.execute_cell("download_dataset")
    
    @pytest.mark.order(3)
    def test_validate_path(self):
        self.tb.execute_cell("validate_path")

    # Here we run a cell, and then test the value of on a object in that cell.
    @pytest.mark.order(4)
    def test_load_fovs(self):
        self.tb.execute_cell("load_fovs")
        
        nb_fovs = tb.ref("fovs")
        expected_fovs = ["fov0", "fov1", "fov2"]
        assert set(nb_fovs) == set(expected_fovs)

    ...

    # Here we run a cell and then test STDOUT. This runs the `create_deepcell_output` cell
    # Test to make sure that the output is printing
    @pytest.mark.order(5)
    def test_create_output(self):
        self.tb.execute_cell("create_output")

        assert "Extracting tif files from DeepCell response" in tb.cell_output_text

```

**Required inputs**

* A *temporary directory* to get all the notebook outputs in one place.
  * Including creating any data used by the notebooks.
* Each notebook should have appropriately tagged cells.

**Output files**

* All output files should output to the *temporary directory*, there are no outputs otherwise.

**Timeline**

About 1 week.

Estimated date when a fully implemented version will be ready for review: 10/28, post example dataset PRs.

Estimated date when the finalized project will be merged in: 10/31.

srivarra:

Closes #765. 

Adjusted the notebook test suite. 1 test class for each notebook, with one test per tagged code cell in the notebooks. Split `base_dir` assignments to their own cell before the rest of the other path setup statements.

**Remaining issues**

In the future construct a small dataset for all the notebook tests. In addition add tests for neighborhood analysis, pairwise spatial enrichment, image_stitching, and for the LDA notebooks.

### DIFF
--- a/ark/utils/notebooks_test.py
+++ b/ark/utils/notebooks_test.py
@@ -3,6 +3,7 @@ from testbook import testbook
 from typing import ContextManager, Iterator
 import pathlib
 from ark.utils import notebooks_test_utils
+from pytest_cases import fixture, parametrize
 
 
 # Sets a shared notebook testing temporary directory. Saves all notebook related files in a
@@ -133,8 +134,8 @@ def nbfib_seg_context(templates_dir, base_dir_generator) -> Iterator[ContextMana
         Iterator[ContextManager]: The testbook context manager which will get cleaned up
             afterwords.
     """
-    POST_CLUSTERING: pathlib.Path = templates_dir / "example_fiber_segmentation.ipynb"
-    with testbook(POST_CLUSTERING, timeout=6000, execute=False) as nb_context_manager:
+    EXAMPLE_FIBER_SEGMENTATION: pathlib.Path = templates_dir / "example_fiber_segmentation.ipynb"
+    with testbook(EXAMPLE_FIBER_SEGMENTATION, timeout=6000, execute=False) as nb_context_manager:
         yield nb_context_manager, base_dir_generator / "efs"
 
 
--- a/ark/utils/notebooks_test.py
+++ b/ark/utils/notebooks_test.py
@@ -118,6 +118,7 @@ def nb4_context(templates_dir, base_dir_generator) -> Iterator[ContextManager]:
         yield nb_context_manager, base_dir_generator / "nb4"
 
 
+@pytest.fixture(scope="class")
 def nbfib_seg_context(templates_dir, base_dir_generator) -> Iterator[ContextManager]:
     """
     Creates a testbook context manager for the fiber segmentation notebook.
--- a/ark/utils/notebooks_test.py
+++ b/ark/utils/notebooks_test.py
@@ -528,6 +528,11 @@ class Test_4_Post_Clustering:
 
 @pytest.mark.order(4)
 class Test_Fiber_Segmentation():
+    """
+    Tests Example Fiber Segmentation for completion.
+    NOTE: When modifying the tests, make sure the test are in the
+    same order as the tagged cells in the notebook.
+    """
     @pytest.fixture(autouse=True, scope="function")
     def _setup(self, nbfib_seg_context):
         """
--- a/ark/utils/notebooks_test.py
+++ b/ark/utils/notebooks_test.py
@@ -1,125 +1,560 @@
-import os
-import subprocess
-import tempfile
-from tempfile import TemporaryDirectory as tdir
-
 import pytest
 from testbook import testbook
-
+from typing import ContextManager, Iterator
+import pathlib
 from ark.utils import notebooks_test_utils
 
-parametrize = pytest.mark.parametrize
+
+# Sets a shared notebook testing temporary directory. Saves all notebook related files in a
+# temporary directory.
+@pytest.fixture(scope="session")
+def base_dir_generator(tmp_path_factory) -> Iterator[pathlib.Path]:
+    """
+    A Fixture which creates the directory where the all notebook test inputs and outputs are
+    saved.
+    Sets the path once for all notebook related tests.
+
+    Args:
+        tmp_path_factory (pytest.TempPathFactory): Factory for temporary directories under the
+            common base temp directory.
+
+    Yields:
+        Iterator[pathlib.Path]: The iterable path containing the path of the `notebook_tests`
+        inputs and outputs.
+    """
+    notebook_test_output_dir = tmp_path_factory.mktemp("notebook_tests")
+    yield notebook_test_output_dir
 
 
-SEGMENT_IMAGE_DATA_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)),
-                                       '..', '..', 'templates', '1_Segment_Image_Data.ipynb')
-PIXEL_CLUSTER_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)),
-                                  '..', '..', 'templates', '2_Cluster_Pixels.ipynb')
-CELL_CLUSTER_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)),
-                                 '..', '..', 'templates', '3_Cluster_Cells.ipynb')
-FIB_SEG_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)),
-                            '..', '..', 'templates', 'example_fiber_segmentation.ipynb')
+@pytest.fixture(scope="session")
+def templates_dir() -> Iterator[pathlib.Path]:
+    """
+    A Fixture which gathers the `templates` directory from `ark-analysis`.
+
+    Yields:
+        Iterator[pathlib.Path]: The directory of the templates relative to this test file.
+    """
+    templates_dir: pathlib.Path = (pathlib.Path(__file__).resolve()).parents[2] / "templates"
+    yield templates_dir
 
 
-def _exec_notebook(nb_filename, base_folder):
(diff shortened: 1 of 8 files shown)


ark/utils/notebooks_test.py
```diff
@@ -1,125 +1,561 @@
...
+        base_dir_generator (pytest.Fixture): The fixture which yields the temporary directory
+            to store all notebook input / output.
+
+    Yields:
+        Iterator[ContextManager]: The testbook context manager which will get cleaned up
+            afterwords.
+    """
+    SEGMENT_IMAGE_DATA_PATH: pathlib.Path = templates_dir / "1_Segment_Image_Data.ipynb"
+    with testbook(SEGMENT_IMAGE_DATA_PATH, timeout=6000, execute=False) as nb_context_manager:
+        yield nb_context_manager, base_dir_generator / "nb1"
```
alex-l-kong, line 58: This particular block is repeated several times for each notebook. Can we pass additional arguments to this function such as:

```
def nb_context(templates_dir, base_dir_generator, notebook_name, notebook_alias, notebook_abbrev) -> Iterator[ContextManager]:
    ...
    NOTEBOOK_PATH: pathlib.Path = templates_dir / notebook_name
    with testbook(NOTEBOOK_PATH, timeout=6000, execute=False) as nb_context_manager:
        yield nb_context_manager, base_dir_generator / notebook_alias
```

We can then call it accordingly.

ark/utils/notebooks_test_utils.py
```diff
@@ -610,26 +182,18 @@ def flowsom_cell_cluster(tb, flowsom_dir, fovs, channels,
...
-        os.path.join(flowsom_dir,
-                     'cell_table_size_normalized.csv'),
-        index=False
-    )
+    cell_table.to_csv(os.path.join(base_dir, cell_table_path), index=False)
 
-    os.makedirs(os.path.join(flowsom_dir, "pixie", '%s_cell_output_dir' % cell_prefix),
-                exist_ok=True)
+    # os.makedirs(os.path.join(base_dir, "pixie", '%s_cell_output_dir' % cell_prefix),
+    #             exist_ok=True)
```
alex-l-kong, line 188: Let's remove these comments if they won't be needed.

alex-l-kong (changes requested):
Main comment regards code that (I believe) can be generalized into a single function.

ark/utils/notebooks_test.py
```diff
@@ -1,125 +1,561 @@
...
+        templates_dir (pytest.Fixture): The fixture which yields the directory of the notebook
+            templates
+        base_dir_generator (pytest.Fixture): The fixture which yields the temporary directory
+            to store all notebook input / output.
+
+    Yields:
+        Iterator[ContextManager]: The testbook context manager which will get cleaned up
+            afterwords.
+    """
+    POST_CLUSTERING: pathlib.Path = templates_dir / "example_fiber_segmentation.ipynb"
```
camisowers, line 136: should be FIBER_SEG or similar

camisowers (changes requested):
Classes look nice! Just one small name change.

ark/utils/notebooks_test.py
srivarra: I don't believe so, as fixtures don't work nicely when using request to store `notebook_name` and `notebook_alias` with inherited fixtures. I tried out some stuff in [pytest-cases](https://smarie.github.io/python-pytest-cases/), but it wasn't working out.

ark/utils/notebooks_test.py
```diff
@@ -1,125 +1,567 @@
...
+
+    def test_base_dir(self):
+        base_dir_inject = f"""
+            base_dir = "{self.base_dir}"
+        """
+        self.tb.inject(base_dir_inject, "base_dir")
+
+    def test_file_path(self):
+        self.tb.execute_cell("file_path")
+
```
ngreenwald, line 168: This looks great, so much cleaner/easier to understand what's going on

ark/utils/notebooks_test.py
```diff
@@ -1,125 +1,567 @@
...
+    def test_seg_scale_set(self):
+        self.tb.execute_cell("seg_scale_set")
+
+    def test_create_output(self):
+        # Get Deepcell Output Dir from the notebook
+        deepcell_output_dir = self.tb.ref("deepcell_output_dir")
+        fovs = self.tb.ref("fovs")
+        # Generate the sample feature_0, feature_1 tiffs
+        notebooks_test_utils.generate_sample_feature_tifs(fovs, deepcell_output_dir, (1024, 1024))
```
ngreenwald, line 198: This is one of the only substantial injections. Is this to avoid the necessity of downloading the example dataset, or sending the data to the deepcell server?

ark/utils/notebooks_test.py
```diff
@@ -1,125 +1,567 @@
...
+        # Get Pixel Clustering filenames
+
+        # Get pixel paths and fovs
+        pixel_data_dir = self.tb.ref("pixel_data_dir")
+        pixel_channel_avg_som_cluster = self.tb.ref("pc_chan_avg_som_cluster_name")
+        pixel_channel_avg_meta_cluster = self.tb.ref("pc_chan_avg_meta_cluster_name")
+        fovs = self.tb.ref("fovs")
+        channels = self.tb.ref("channels")
+        # Create fake pixel som files
+        notebooks_test_utils.create_pixel_som_files(self.base_dir,
```
ngreenwald, line 306: Same thing for this and the cell_som function. Is this so that we don't have to actually train the SOM due to save time, or for some other purpose?

ngreenwald (changes requested):
Looks great! See below for a couple comments

ark/utils/notebooks_test.py
srivarra: Yes, this is for skipping the SOM training portion.

ark/utils/notebooks_test.py
srivarra: Yeah this is for avoiding sending anything to the Deepcell service directly.

ngreenwald:
Okay, I think eventually we'll want to actually test these functions

ngreenwald (approved):

PR #801, opened 2022-10-31
Standardize usage of `.tiff` extension

### LINKED ISSUE #792: Standardize use of `.tiff` extension in `ark-analysis`

The use of `.tif` vs `.tiff` is inconsistent in `ark-analysis` right now. This can create issues with keeping track of which extension should be used when. Best to standardize `.tiff` since this is the extension the image files are saved with.

The main fix concerns the segmentation masks, which are currently saved as `.tif` files. This is what the input to Deepcell is saved as, which propagates over to the `.zip` files.

Any other places where `.tif` instead of `.tiff` is used should be changed.

Additionally, these changes will need to be propagated to HuggingFace, as the example dataset will not work otherwise.

**NOTE: the one exception should be MIBItiff file support, these should remain as `.tif` for backwards compatibility.**

alex-l-kong:

Closes #792. Replace any use of `.tif` with `.tiff`.

Fairly straightforward renaming. A couple of caveats:

- DeepCell extracts the files as `.tif` in the `.zip` files, so these need to be explicitly renamed prior to save
- A new HuggingFace [PR](https://huggingface.co/datasets/angelolab/ark_example/discussions/3) has been opened to address the renaming of the example dataset files


ark/utils/example_dataset.py
```diff
@@ -154,7 +154,7 @@ def get_example_dataset(dataset: str, save_dir: Union[str, pathlib.Path],
 
     example_dataset = ExampleDataset(dataset=dataset, overwrite_existing=overwrite_existing,
                                      cache_dir=None,
-                                     revision="main")
+                                     revision="22c044ac9a62200849e0f66a83aa2dfd30a2c56e")
```
alex-l-kong, line 157: This will be changed back to `'main'` prior to merging into the `main` branch.

ark/utils/deepcell_service_utils.py
```diff
@@ -141,12 +138,19 @@ def zip_write(zip_path):
 
         with ZipFile(zip_files[-1], "r") as zipObj:
             for name in zipObj.namelist():
+                # generate the path to save the segmentation mask
                 mask_path = os.path.join(deepcell_output_dir, name)
+
+                # DeepCell uses .tif extension, append extra f to account for .tiff standard
+                mask_path += 'f'
```
alex-l-kong, line 145: Not a huge issue to fix this, but is there a reason DeepCell uses `.tif` instead of `.tiff`?

ngreenwald: You can talk to @srivarra, some parts of this are the default output from the zip object, some we control, I don't remember which is why

ark/utils/load_utils.py
```diff
@@ -131,7 +131,7 @@ def load_imgs_from_tree(data_dir, img_sub_folder=None, fovs=None, channels=None,
     if channels is None:
         channels = iou.list_files(
             dir_name=os.path.join(data_dir, fovs[0], img_sub_folder),
-            substrs=['.tif', '.jpg', '.png']
+            substrs=['.tif', '.tiff', '.jpg', '.png']
```
ngreenwald, line 134: Should we remove `tif` as an option? Is this going to lead to issues elsewhere if the user provides single-f files? Or it won't matter and it's fine to keep it?

ark/utils/load_utils.py
```diff
@@ -246,7 +246,7 @@ def load_imgs_from_dir(data_dir, files=None, match_substring=None, trim_suffix=N
     iou.validate_paths(data_dir, data_prefix=False)
 
     if files is None:
-        imgs = iou.list_files(data_dir, substrs=['.tif', '.jpg', '.png'])
+        imgs = iou.list_files(data_dir, substrs=['.tif', '.tiff', '.jpg', '.png'])
```
ngreenwald, line 249: Same question

ngreenwald (changes requested):
Looks good. Did you do a control-f for the whole repo/notebooks to make sure none slipped through?

ark/utils/load_utils.py
alex-l-kong: I think we can remove it, `.tif` should only be kept for MIBItiff backwards compatibility and they won't be using this function (or `load_imgs_from_dir`) to load in anything.

ark/utils/deepcell_service_utils.py
alex-l-kong: @ngreenwald just talked with @srivarra, this is the default output from the `.zip` object (not something we control).

srivarra (approved):
Looks good!

ngreenwald (changes requested):
All of those images in `/example_dataset` are no longer needed because of the huggingface dataset, right? If so they can be removed. 

Some of that example data is for the spatial analysis notebooks (I think), so it needs to be kept around until we update the example dataset for those notebooks. However, anything that is only used for notebooks 1-4 can be deleted.

ark/utils/example_dataset_test.py
```diff
@@ -23,7 +23,7 @@ def dataset_download(request) -> Iterator[ExampleDataset]:
     example_dataset: ExampleDataset = ExampleDataset(
         dataset=request.param,
         cache_dir=None,
-        revision="main"
+        revision="22c044ac9a62200849e0f66a83aa2dfd30a2c56e"
```
ngreenwald, line 26: This one as well

ark/utils/example_dataset.py
ngreenwald: Looks good

ngreenwald (changes requested):
Looks good, just need to change the huggingface references back to main

ngreenwald (approved):
Looks good!

PR #803, opened 2022-11-03
Refactor distance matrix saving

### LINKED ISSUE #718: Remove unnecessary batching functionality from functions

We've found that in some cases, batching exists solely for the purpose of loading more image data in at once. We later loop over each FOV in that batch anyways, which doesn't add any significant optimization.

This additionally causes issues with cohorts with different image sizes, because loading a batch into a 1024x1024 or 2048x2048 array will cause dimension errors.

We've already removed batching from `generate_deepcell_input`, the following functions also need to be modified in this regard:

@ngreenwald looks like the additional following functions use an explicit batch_size argument without multiprocessing:

- [] `data_utils.generate_and_save_pixel_cluster_masks` (and by extension, `data_utils.generate_pixel_cluster_mask`)
- [] `data_utils.generate_and_save_cell_cluster_masks` (and by extension, `data_utils.label_cells_by_cluster`)
- [] `marker_quantification.generate_cell_table`
- [] `spatial_analysis.batch_channel_spatial_enrichment` and `spatial_analysis.batch_cluster_spatial_enrichment`: talked to @ackagel about this, we can condense these to a per-FOV basis with negligible speed difference
- [] `spatial_analysis.create_neighborhood_matrix` (and by extension, the neighborhood analysis notebook): currently, this process requires precomputing all the distance matrices prior to running then function, then an additional per-FOV loop during neighborhood analysis. We should condense all this down to one per-FOV loop in the neighborhood matrix function.
- [] `spatial_analysis_utils.calc_dist_matrix`: after the neighborhood analysis process is updated, the dist matrix function will no longer be receiving a batch of FOVs to process over, so it should be condensed to process just one FOV.

### LINKED ISSUE #720: Replace batching functionality of `ark` pipeline to per-FOV looping

### LINKED ISSUE #770: Update `calc_dist_matrix` so it preprocesses prior to running each type of spatial analysis
**Relevant background**

This was originally going to be addressed as part of #718 and #720, however, it requires enough restructuring to warrant its own design doc.

`calc_dist_matrix` is an expensive process when run on large FOV cohorts which contain many unique cell labels. The main slowdown is in the functionality of `regionprops` and `regionprops_table`. Currently, we're computing these distance matrices separately for the 3 types of spatial analyses conducted (channel spatial, cluster spatial, and neighborhood). We should ideally have all these preprocessed as a prior step.

Addressing the speed of `regionprops` should be a separate issue as this function is used throughout this repo in various capacities. In either case, it will be useful to have the distance matrices pre-computed to prevent duplicate work.

**Design overview**

We should make use of the save feature in the distance matrix calculation, which allows us to save all the matrices in a single file that can then be loaded up again.

In both spatial analysis notebooks, we should allow the user to specify the path to this distance matrix file so it can then be loaded up and passed directly into the spatial analysis computation and visualization functions with ease. Since these distance matrices are indexed by FOV, they will integrate with the now-unbatched functions with ease.

**Code mockup**

_Distance matrix creation and saving_

This will be handled by a call to `calc_dist_matrix` with the argument `save_path`. It seems unnecessary to create a separate notebook to do this, as the process is fairly trivial (a call to `calc_dist_matrix`).

We should generate these distance matrices at the beginning of each spatial analysis notebook. If the distance matrix file path doesn't exist, we'll create it, otherwise we'll load it up.

The user will explicitly specify a `dist_mat_dir` as a notebook variable when setting paths.

_Updating `calc_dist_matrix` to support label maps of different sizes_

One of the main motivations for unbatching was to allow for image cohorts of varying sizes to be processed. However, `calc_dist_matrix` currently expects an `xarray` of label maps, which makes an assumption of consistent dimensions we can no longer use.

Because `xarray` does not support images of different dimensions, we should refactor `calc_dist_matrix` so it takes a `label_dir` argument instead. In this way, we can load the label maps in one-by-one to prevent conflicting dimensions.

_Standardizing use of `.tiff` extension_

To streamline the actual propagation of changes, we'll need to standardize the use of the `.tiff` file extension throughout the repo. Currently, the inconsistent use causes problems for loading in label files vs image files, some of which may use `.tif` and others which may use `.tiff`. Standardizing the extension to `.tiff` will prevent us from having to write complicated logic to handle both when loading in these FOV-level files.

These changes will need to be made throughout the pipeline and also on Hugging Face. Because of how extensive this issue is, this will need to be addressed in a separate PR prior to finishing up the next and final step (see #792).

_Updating existing logic_

All of the spatial analysis computation functions (`generate_channel_spatial_enrichment_stats`, `generate_cluster_spatial_enrichment_stats`, and `create_neighborhood_matrix`) will need to be updated so they take in all of the pre-loaded `dist_mats`. We should specify an additional argument to these called `dist_matrices_dict` that allow us to retrieve the distance matrix by `fov` name. This `dist_matrices_dict` can then be passed down to `calculate_channel_spatial_enrichment` and `calculate_cluster_spatial_enrichment`.

**Required inputs**

The user will need to set a `save_path` needed to set for saving the distance matrices. As described in the previous section, each spatial analysis function should now take in a `dict` of distance matrices.

**Output files**

The distance matrices will be saved as `.xr` files to `dist_mat_dir` in both spatial analysis notebooks. **This directory should only contain distance matrices**.

**Timeline**
- [ ] A couple days
- [ ] A week
- [X] Multiple weeks (no more than 3 in my case). For large projects, make sure to agree on a plan that isn't just a single monster PR at the end. 

11/2

No later than 11/4

alex-l-kong:

Closes #770. In turn closes #718 and closes #720.

See design doc linked in the issue.


ark/utils/spatial_analysis_utils.py
```diff
@@ -12,64 +13,69 @@
...
 
 import ark.settings as settings
-from ark.utils import io_utils, misc_utils
+from ark.utils import io_utils, load_utils, misc_utils
 from ark.utils._bootstrapping import compute_close_num_rand
 
 
-def calc_dist_matrix(label_maps, save_path=None):
-    """Generate matrix of distances between center of pairs of cells
+def calc_dist_matrix(label_dir, save_path, prefix='_feature_0'):
```
ngreenwald, line 20: Is there logic to make sure this isn't run if the file exists already elsewhere, or should that be included here?

ngreenwald (changes requested):
Looks nice. Will you make sure you get qualitatively similar results when running this version and the previous version of the huggingface dataset? And also that it works as expected both with, and without, the distance matrices already being created?

ark/utils/spatial_analysis_utils.py
alex-l-kong: We check this in both `example_pairwise_spatial_enrichment.ipynb` and `example_neighborhood_analysis_script.ipynb`. We create `dist_mat_dir` (synonymous with `save_path` here) and populate it only if it doesn't already exist.

alex-l-kong:
> Looks nice. Will you make sure you get qualitatively similar results when running this version and the previous version of the huggingface dataset? And also that it works as expected both with, and without, the distance matrices already being created?

Spatial analyses don't rely on HuggingFace for data yet, did you mean verify against the current version in `main` or something different.

ngreenwald:
They don't rely on it, but there's a segmentation mask and a cell table in there, you should be able to use for the scripts. I think it'll be a better check to catch any potential errors since that dataset is actually large enough to get spatial relationships

cliu72 (commented):
Looks good to me. Did you go through and make sure the distance matrices are the same before and after this refactoring?

ark/utils/spatial_analysis_utils.py
```diff
@@ -12,64 +13,69 @@
...
+        fov_name = fov_file.replace(prefix + '.tiff', '')
 
-    # Extract list of fovs
-    fovs = label_maps.coords['fovs'].values
+        # load in the data
+        fov_data = load_utils.load_imgs_from_dir(
+            label_dir, [fov_file], match_substring='_feature_0',
+            trim_suffix='_feature_0', xr_channel_names=['segmentation_label']
+        )
+
```
cliu72, line 50: Is there a reason you're hard-coding in '_feature_0' in the `load_imgs_from_dir` function, even though `calc_dist_matrix` has a 'prefix' argument?

alex-l-kong: Good point, I'll replace this with `prefix`.

ark/analysis/spatial_analysis.py
```diff
@@ -169,71 +173,73 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds,
...
+                    context_mark_nums, context_pos_labels, context_dist_mat, dist_lim,
+                    bootstrap_num
+                )
 
-                close_num_rand = close_num_rand + \
-                    spatial_analysis_utils.compute_close_cell_num_random(
-                        context_mark_nums, context_pos_labels, context_dist_mat, dist_lim,
-                        bootstrap_num
-                    )
+    close_num_rand = spatial_analysis_utils.compute_close_cell_num_random(
```
cliu72, line 220: I think we're missing an `else` before this line - if `context_col is not None`, `compute_close_cell_num_random` is already called and shouldn't be called again. The corresponding `calculate_cluster_spatial_enrichment` function does have an `else`.

ark/analysis/spatial_analysis.py
```diff
@@ -169,71 +173,73 @@ def calculate_channel_spatial_enrichment(dist_matrices_dict, marker_thresholds,
...
+        close_num_rand = np.zeros((*close_num.shape, bootstrap_num), dtype=np.uint16)
 
-        if context_col is not None:
-            close_num_rand = np.zeros((*close_num.shape, bootstrap_num), dtype=np.uint16)
+        context_nums_per_id = \
+            current_fov_data.groupby(context_col)[cell_label_col].apply(list).to_dict()
 
-            context_nums_per_id = \
-                current_fov_data.groupby(context_col)[cell_label_col].apply(list).to_dict()
+        for name_i, name_j in context_pairings:
```
cliu72, line 196: I'm confused by this. This question probably doesn't belong in this PR, but why are we iterating through pairs of contexts? After Adam left, is there a go-to person for questions about this spatial enrichment script?

ark/analysis/spatial_analysis.py
alex-l-kong: Good catch, not sure why this never got added in earlier but I'll fix this up.

alex-l-kong:
@ngreenwald @cliu72 verified the outputs match between the old and new versions using the sample data on HuggingFace.

ark/analysis/spatial_analysis.py
ngreenwald: Nope, not really

PR #823, opened 2022-11-10
Subset pixel channel averaging to remove bottleneck

### LINKED ISSUE #622: compute_pixel_cluster_channel_avg is prohibitively slow
The `compute_pixel_cluster_channel_avg` function takes as long as the actual mapping of pixels. This isn't a big deal for small datasets, but for large datasets this single function takes almost 14 hours to run. 

Given that the only point of this function is to create the weights for the heatmap, I think some aggressive downsampling for the number of FOVs that this average is computed from would work.

Is there different logic we could use to create this @cliu72? The same function gets called again in the metaclustering step, which doubles the amount of time devoted to this

### LINKED ISSUE #799: Check for both FOV and summary file processing before skipping a processing step during pixel clustering.
In many of the steps for pixel clustering, the script will check to see if FOVs have been processed in the feather file (ex. 'cluster_pixels') or if a single summary file has been made (ex. ;train_pixel_som') before proceeding to run that particular step. However, if your script crashes when rendering the summary files that come out of each step these cannot be easily reproduced because the 'check' will see that the FOVs have been processed and think that step is completed. For example, I have been dealing with occasional OS-related issues when trying to do consensus clustering. My individual FOV feather files will have the metacluster information appended, but the script will crash before producing the summary csv generated by 'compute_pixel_cluster_channel_avg,' which is called from 'pixel_consensus_cluster'. If I try to re-run the consensus clustering step, it will not rerun because it 'sees' that all FOVs have been processed. In this case I would have to make a call from within the Jupyter notebook directly to 'pixel_cluster_utils.compute_pixel_cluster_channel_avg.' This is thematic across all of the steps of pixel clustering. It seems to me that there are two solutions: 

1. Check that the FOV-level AND summary file outputs have been completed from each step and enable the scrip to 'pick up where it left off' if both or either are missing.
2. For the relevant functions, separate the FOV processing from the generation of the summary files in the Jupyter notebook with two different functions. Right now both are generated in a single function call but the same function call will only check for the existence of one or another to indicate if that step has been completed.

alex-l-kong:

Closes #622. Closes #799. With several FOVs, `compute_pixel_cluster_channel_avg` can become extremely slow as multiple DataFrame summary stats have to be aggregated, concatenated, then aggregated again. This is unavoidable because we can't load the entire dataset into memory due to Docker limitations.

However, subsetting the number of pixels for each FOV that get used for averaging will allow us to generate computationally similar results and provide a massive performance boost.

We add a parameter `subset_proportion` which functionally works like how it's implemented in `create_pixel_matrix`. Prior to averaging per pixel SOM or meta cluster, we randomly sample `subset_proportion` number of pixels from the FOV data.

An additional speed boost can be obtained by storing the computed DataFrames in a list prior to concatenating.

**Remaining issues**

One potential problem exists in the subsetting not being stratified across all SOM or meta clusters. As a result, a user could end up losing clusters (especially SOM clusters) during the averaging process. This may not be a major issue but still worth noting.


ngreenwald (changes requested):
I thought part of the slowness was loading each individual FOV? How much does this speed things up? If IO is the bottleneck, then perhaps processing a subset of FOVs, rather than a subset of each FOV, would be faster.

alex-l-kong:
@ngreenwald yes, this will be more optimal. I'll update the logic.

cliu72 (commented):
I think it would be better to leave the default for the subsetting as 1 (i.e. no subsetting) in the function definitions, and have it as an option in the notebook for users to change it. I think a lot of people who haven't used Pixie before are going to first run the notebook as is with a small test set, and subsetting 10% of a small test set of FOVs is a small number - which might throw off the averaging, leading people to think that Pixie sucks.

I think the point you made in "Remaining Issues" is a good one - perhaps after subsetting and averaging, we can add a check that the number of clusters is what the users specified. And if it's not for whatever reason, have a warning that tells the user to increase the subset proportion. Or something like that, open to other suggestions.

ngreenwald:
@alex-l-kong, I thought we talked yesterday about changing the default behavior to address this issue, specifying the number of FOVs to keep instead of a percentage of the total.

cliu72:
Ah, if that's the case, I think it'd still be best if the default behavior was to use all the FOVs. I found that a lot of people who have been asking me about Pixie so far just run the entire notebook without paying much attention to the parameters they can change...

ngreenwald:
I was thinking set it to like 100 FOVs or something, where we think it wouldn't make a difference, but if you want the default to be everything to make sure there's no issues then the subsetting method doesn't really matter and we can leave it as is

cliu72:
I guess if the default is 100, we need to just add a check for datasets that have fewer than 100. I think it'd be easier to set it to everything - but I also don't have strong opinions about this.

alex-l-kong:
@ngreenwald yup we did, just hadn't had the opportunity to meet with @cliu72 about it yet. @cliu72 tomorrow I'll be in lab during the afternoon. If there's anything else to discuss, does 3 PM work?

alex-l-kong:
Personally, I like the idea of setting it by default to `min(100, len(fovs))`. I think we're at a stage where some people are just starting out with a small dataset while others will have massive cohorts. This should be a nice middle ground (so anyone with few FOVs can use them all, while those with thousands can limit their analysis to 100).

Also don't think we need an explicit notebook variable to change this unless even 100 proves problematic for some people.

cliu72:
Sure, default to `min(100, len(fovs))` sounds good to me. I don't think there's a need to meet, but I will also be around if you want to discuss anything.

alex-l-kong:
Will need to spend a bit more time on this PR because to handle the case where we lose SOM or meta clusters, a few things need to happen:

- These clusters will need to be deleted across all the FOV data files, as they will otherwise interfere with the remapping process.
- The user should have the opportunity to retry with a higher `num_fovs_subset` (which replaced `fov_subset_proportion`). Currently, however, this will not be possible because restarting only checks for the creation of the `pixel_som_cluster` or `pixel_meta_cluster` columns, which will have all been added by the time channel averaging starts. This will probably need to be a separate PR as it's related to #799.

Let me know what you think @cliu72.

ngreenwald:
What do you mean the scenario where we lose clusters? What does computing the pixel channel average have to do with metaclustering?

alex-l-kong:
> What do you mean the scenario where we lose clusters? What does computing the pixel channel average have to do with metaclustering?

Problem is the meta cluster remapping GUI does remapping based on the average pixel channel expression per SOM cluster data, which is now computed with a subset of FOVs. If the subset of FOVS taken for this computation contains just 95 SOM clusters instead of 100, we lose the ability to map the remaining 5 SOM clusters to a renamed meta cluster. This will cause integration issues with `apply_pixel_meta_cluster_remapping`, which assumes that we have this information for all the SOM clusters.

I think the easiest solution here is to, as Candace mentioned, throw a warning/error if the number of SOM clusters in `pc_chan_avg_som_cluster.csv` is less than the total number of SOM clusters generated. However, this warning/error isn't useful unless the user has the option to retry this part of the pipeline. This isn't currently possible as `cluster_pixels` doesn't currently support re-running the pipeline if only these average summary expression files need re-generation. This is why I think it may be best to address Erin's issue side-by-side.

ngreenwald:
Got it. Can we pull this part of the function out of the larger clustering function, and have it just run on its own? Then if the error is thrown, the user can re-run just the average cluster computation?

alex-l-kong:
> Got it. Can we pull this part of the function out of the larger clustering function, and have it just run on its own? Then if the error is thrown, the user can re-run just the average cluster computation?

Yeah let's do that here. I'll just address the pixel clustering side of matters with this PR. After this one is merged in I'll apply the same thing with cell clustering in a separate PR (not for subsetting in particular, just giving the user an option to restart properly if the summary file generation failed).

cliu72:
> Will need to spend a bit more time on this PR because to handle the case where we lose SOM or meta clusters, a few things need to happen:
> 
> * These clusters will need to be deleted across all the FOV data files, as they will otherwise interfere with the remapping process.
> * The user should have the opportunity to retry with a higher `num_fovs_subset` (which replaced `fov_subset_proportion`). Currently, however, this will not be possible because restarting only checks for the creation of the `pixel_som_cluster` or `pixel_meta_cluster` columns, which will have all been added by the time channel averaging starts. This will probably need to be a separate PR as it's related to [Check for both FOV and summary file processing before skipping a processing step during pixel clustering. #799](https://github.com/angelolab/ark-analysis/issues/799).
> * Related to the above point, we might need to consider supporting an `overwrite` flag to the clustering functions. The current restart logic is getting a bit unwieldy, I think this flag would be a great way to hard reset a particular part of the pipeline in case of catastrophic failure (which has popped up more than a few times with others I've helped).
> 
> Let me know what you think @cliu72 and @ngreenwald.

I don't think it's a good idea to delete unmapped clusters across FOVs.



> > What do you mean the scenario where we lose clusters? What does computing the pixel channel average have to do with metaclustering?
> 
> Problem is the meta cluster remapping GUI does remapping based on the average pixel channel expression per SOM cluster data, which is now computed with a subset of FOVs. If the subset of FOVS taken for this computation contains just 95 SOM clusters instead of 100, we lose the ability to map the remaining 5 SOM clusters to a renamed meta cluster. This will cause integration issues with `apply_pixel_meta_cluster_remapping`, which assumes that we have this information for all the SOM clusters.
> 
> I think the easiest solution here is to, as Candace mentioned, throw a warning/error if the number of SOM clusters in `pc_chan_avg_som_cluster.csv` is less than the total number of SOM clusters generated (ditto for meta clusters). However, this warning/error isn't useful unless the user has the option to retry this part of the pipeline. This isn't currently possible as `cluster_pixels`, `pixel_consensus_cluster`, and `apply_pixel_meta_cluster_remapping` don't currently support re-running the pipeline if only these average summary expression files need re-generation. This is why I think it may be best to address Erin's issue side-by-side.
> 
> UPDATE: alternatively, for each "missing" SOM cluster, we could also find a different SOM cluster that maps to an equivalent "raw" meta cluster to find its renamed meta cluster value. I'd still argue it's better for users to have the average channel expression for all SOM and meta clusters, but that's something we can discuss further.

Similarly, I don't think "finding a different SOM cluster that maps to an equivalent "raw" meta cluster" is a good idea. We would potentially be losing rare clusters if we do this.

I think it's best to just throw an error and let the users re-run just the average computation, as Noah suggested.

ark/phenotyping/pixel_cluster_utils.py
```diff
@@ -317,10 +327,29 @@ def compute_pixel_cluster_channel_avg(fovs, channels, base_dir, pixel_cluster_co
...
+    # warn the user that we can only select so many FOVs if len(fovs) < num_fovs_subset
+    if len(fovs) < num_fovs_subset:
+        warnings.warn(
+            'Provided num_fovs_subset=%d but only %d FOVs in dataset, '
+            'subsetting just the %d FOVs' %
+            (num_fovs_subset, len(fovs), len(fovs))
+        )
+
+    # subset number of FOVs based on num_fovs_subset if less than total number of fovs
+    fovs_sub = random.sample(fovs, num_fovs_subset) if num_fovs_subset < len(fovs) else fovs
```
cliu72, line 350: Do we want to add a seed here to make it reproducible?

cliu72:
Also, food for thought but doesn't necessarily need to be addressed in this PR. Getting the means of the metaclusters doesn't actually require re-reading in individual FOVs. We could just use the SOM cluster means and the counts of each SOM cluster to do that calculation (like if SOM clusters 1, 2, and 3 are in metacluster 1, just use the expression profiles and counts of those SOM clusters to calculate metacluster 1 mean). Could potentially speed those steps up, if the bottleneck is reading in individual FOVs.

alex-l-kong:
> Also, food for thought but doesn't necessarily need to be addressed in this PR. Getting the means of the metaclusters doesn't actually require re-reading in individual FOVs. We could just use the SOM cluster means and the counts of each SOM cluster to do that calculation (like if SOM clusters 1, 2, and 3 are in metacluster 1, just use the expression profiles and counts of those SOM clusters to calculate metacluster 1 mean by multiplying each cluster mean by its count to get sum, add up all the clusters, then divide by total count). Could potentially speed those steps up, if the bottleneck is reading in individual FOVs.

I think that's a good idea, we lose far less information this way. I think it's best for the next PR.

ark/phenotyping/pixel_cluster_utils.py
alex-l-kong: Yup, that's a good idea.

alex-l-kong:
Tested and verified that this new process works. I think it's going to be easier to merge this one in first prior to consensus clustering. Gonna be a less painful merge this way.

ngreenwald (commented):
Looks good, I defer to candace

cliu72 (commented):
Looks good to me. Can we change the error message to:
```
'Averaged data contains just %d clusters out of %d, average expression file not written. '
'Increase your num_fovs_subset value.'
```

"Removing those values" is confusing to me.

alex-l-kong:
@cliu72 error message has been fixed.

cliu72 (approved):
Looks good

alex-l-kong:
@ngreenwald just need your go-ahead to merge in

ngreenwald (approved):

PR #839, opened 2022-11-21
Add native Python consensus clustering process

### LINKED ISSUE #831: Port pixel and cell consensus clustering process to Python
**Relevant background**

ConsensusClusterPlus will need to be handled separately from `pyFlowSOM` in the transition to a purely Python-based pixel and cell clustering workflow. We don't need to find a perfectly equivalent Python alternative, but it does need to support the general idea of consensus clustering.

**Design overview**

[This](https://github.com/ZigaSajovic/Consensus_Clustering) implementation of consensus clustering should do the trick. Because this repo has not made a formal release, we would need to directly copy the `ConsensusCluster` class in  `consensusClustering.py` file to our repo for our own use.

This class makes it easy for us to replace the existing `ConsensusClusterPlus` logic in R; both `pixel_consensus_cluster.R` and `cell_consensus_cluster.R` can be removed after this. Note that scaling, capping, and cluster assignment will need to be moved to `pixel_consensus_cluster` and `cell_consensus_cluster` in `pixel_cluster_utils.py` and `cell_cluster_utils.py` respectively.

**Code mockup**

_Abstract Consensus Cluster Class_

We should create a separate file called `clustering_helpers.py` in `phenotyping`. This will initially contain just the Pixie consensus cluster class (`PixieConsensusCluster`) but will later also contain `PixieSOMCluster` after `pyFlowSOM` is finalized.

`PixieConsensusCluster` will have the following general implementaiton:

``` 
class PixieConsensusCluster:
    def __init__(self, input_file: pathlib.Path, columns: List[str], max_k: int = 20, cap: float = 3):
        self.input_data = pd.read_csv(input_file)
        self.columns = columns # the columns to subset on
        self.max_k = max_k
        self.cap = cap
        self.cc = # define ConsensusCluster object with parameters cluster_ = sklearn.AgglomerativeClustering (same as hierarchical clustering for CCP) L = 2 (default min), K = max_k (same as maxK arg in CCP), H = 10 (same as reps in CCP), and resample_proportion = 0.8 (same as pItem in CCP).
        self.mapping = None  # this will be set during generate_som_to_meta_map

    def scale_data(self):
        # equivalent of running scale, pmin, and pmax in R on self.input_data[columns], we'll need to find equivalent Python functions but shouldn't be hard

    def run_consensus_clustering(self):
        self.cc.fit(self.input_data[columns])

    def generate_som_to_meta_map(self):
        # use a slightly modified version of predict_data by Sajovich to map SOM clusters to meta clusters and assign to self.mapping. We have to modify as that function as it uses the k with the best metrics, but we always want the max k

    def save_som_to_meta_map(self, save_path: pathlib.Path):
        self.mapping.to_csv(save_path)  # may be adjusted to support consistency of setting base_dir

    def assign_consensus_labels(self, external_data: pandas.DataFrame) -> pandas.DataFrame:  # NOTE: for pixel clustering, this will be called for each FOV, the looping/multiprocessing support will be handled by the pixel_consensus_cluster function in pixel_cluster_utils.py
        # use self.mapping to assign equivalent meta cluster based on the SOM cluster
``` 

This functionality will replace both `pixel_consensus_cluster.R` and `cell_consensus_cluster.R`.

_Integration with `pixel_consensus_cluster`_

Prior to the call to `assign_consensus_labels`, `pixel_consensus_cluster` and `cell_consensus_cluster` must load the pixel/cell data in, and afterwards save it to the appropriate location. Note that for pixel clustering, the loop/multiprocessing currently in `pixel_consensus_cluster.R` will need to be moved to `pixel_consensus_cluster` in `pixel_cluster_utils.py`:

```
if multiprocess:
    with multiprocessing.get_context('spawn').Pool(batch_size) as fov_data_pool:
        for fov_batch in [fovs_list[i:(i + batch_size)]
                          for i in range(0, len(fovs_list), batch_size)]:
            # load in all the DataFrames and use fov_data_pool to make a multiprocess call to assign_consensus_labels
            # loop over the processed DataFrames and save
else:
    for fov in fovs_list:
        # load in the DataFrame and call assign_consensus_labels
        # save
```

_Integration with `cell_consensus_cluster`_

Because no multiprocessing is needed for cell consensus clustering, this function only needs to handle additional file loading and saving currently done in R.

_Porting over `ConsensusCluster` class_

We can copy over the `ConsensusCluster` class directly to `clustering_helpers.py`, or we can copy over the file directly. Either way, it shouldn't be necessary to write our own test functions for this.

_Testing_

We can use a small dataset (via `pytest.fixture`) to test this functionality. Since this is an abstract consensus clustering class, we can just test the general functionality in `clustering_helpers_test.py`, and leave the pixel and cell-specific tests to `notebooks_test.py` using the example datasets (the internal logic will need amending as there's no need to mock the call to `pixel_consensus_cluster`).

Additionally, `pixel_consensus_cluster` and `cell_consensus_cluster` can now be directly tested instead of being mocked. As with `clustering_helpers_test.py`, we can use simple datasets for these.

**Required inputs**

`PixieConsensusCluster` will need the respective SOM average file (generated by `cluster_pixels` and `cluster_cells`) as well as the `cap` and `max_k` arguments from the user.

**Output files**

Same as the original.

**Timeline**

- [ ] A couple days
- [ ] A week

11/27/22 (Thanksgiving break will push this one out a bit longer...)

11/29/22

### LINKED ISSUE #852: Seed invalid integer
### Discussed in https://github.com/angelolab/ark-analysis/discussions/851

<div type='discussions-op-text'>

<sup>Originally posted by **whitneyt1** December  6, 2022</sup>
Hi! 

While pixel clustering, I encountered the following error: Error in set.seed(): seed invalid integer
Screen Shot 2022-12-06 at 1 53 26 PM

When I tried manually setting seed value to 42, i received the same error. 

I corrected this by changing the pixel_consesus_cluster.R file to: 
Screen Shot 2022-12-06 at 2 03 06 PM

instead of it calling the arg[11] from the pixel_cluster_utils.py file, which reads in str(seed). 
Screen Shot 2022-12-06 at 2 04 28 PM

This might be being address in #839, but just FYI! 

thanks! 
</div>

alex-l-kong:

Closes #831. Closes #852. Consensus clustering needs to be ported over to Python as part of the efforts to remove all R code.

See #831. I decided not to use a fixture for `test_pixel_consensus_cluster` and `test_cell_consensus_cluster` since those functions already generate test datasets. When we port the entire pipeline over to OOP, then we'll think about creating test fixtures.


alex-l-kong:
Tested that this process works and produces consistent results for both pixel and cell clustering (and integrates with multiprocessing in the former). Doesn't seem to replicate the original exactly but per @cliu72 this won't be an issue.

Main questions I have are regarding:

- The current flow
- Testing, specifically of `PixieConsensusCluster`. As it stands it's fairly repetitive for the pixel and cell cluster cases. Looping in @srivarra for additional inputs.

srivarra (commented):
It looks good! Just a couple of small things.

ark/phenotyping/cluster_helpers.py
```diff
@@ -0,0 +1,261 @@
...
+    def _internal_resample(self, data: pd.DataFrame, proportion: float):
+        """Implements resampling procedure.
+
+        Args:
+            data (pandas.DataFrame):
+                The data in `(examples,attributes)` format.
+            proportion (float):
+                The percentage to sample.
+        """
+        resampled_indices = np.random.choice(
```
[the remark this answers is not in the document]
alex-l-kong: I'll double-check, but I think setting the `np.random.seed` in `pixel_consensus_cluster`/`cell_consensus_cluster` prior to invoking the `run_consensus_clustering` method does the trick here.

ark/phenotyping/cluster_helpers.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+class ConsensusCluster:
+    def __init__(self, cluster: Callable, L: int, K: int, H: int,
+                 resample_proportion: float = 0.5):
+        """Implementation of Consensus clustering, following the paper
+        https://link.springer.com/content/pdf/10.1023%2FA%3A1023949509487.pdf
+
+        Args:
+            cluster (Callable):
```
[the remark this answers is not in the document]
alex-l-kong: Not sure we'll need explicit error checks inside either constructor since we'll always be using `sklearn.cluster.AgglomerativeCluster`.

Is this something that can be done inside the documentation or the type hint?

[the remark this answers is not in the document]
srivarra: Interesting, I think what we're looking for is structural typing. I found some [documentation on it](https://peps.python.org/pep-0544/).

```python
from typing import Protocol, runtime_checkable

@runtime_checkable
class SupportsCluster(Protocol):
    def fit_predict(self) -> None: ... # Actually using the `...` 
    
    @property
    def n_clusters(self) -> int:
        return n_cluster
```

Then with typehinting you can do:
```python
class ConsensusCluster:
    def __init__(self, cluster: SupportsCluster, L: int, K: int, H: int, resample_proportion: float = 0.5):
        pass
```

ngreenwald (commented):
I'll defer to @cliu72 on this one

ark/phenotyping/cluster_helpers.py
```diff
@@ -0,0 +1,261 @@
...
+        self.columns = columns
+        self.max_k = max_k
+        self.cap = cap
+        self.cc = ConsensusCluster(
+            cluster=AgglomerativeClustering,
+            L=max_k,
+            K=max_k,
+            H=10,
+            resample_proportion=0.8
+        )
```
cliu72, line 204: Why are `H` and `resample_proportion` default to the values here?

ark/phenotyping/cluster_helpers.py
```diff
@@ -0,0 +1,261 @@
...
+
+        Returns:
+            pandas.DataFrame:
+                The data matrix prediction for `self.bestK`.
+        """
+        assert self.Mk is not None, "First run fit"
+        return self.cluster_(n_clusters=self.bestK).fit_predict(
+            data)
+
+
```
cliu72, line 161: Just to clarify, is everything above this copied 100% from ZigaSajovic, or did you change any of it?

ark/phenotyping/cluster_helpers.py
alex-l-kong: It's mostly an attempt to replicate `ConsensusClusterPlus` default args as closely as possible.

`H` is the same as `reps` in `ConsensusClusterPlus`, which is by default set to 10. `resample_proportion` is the same as `pItem` in `ConsensusClusterPlus`, which is by default set to 0.8.

ark/phenotyping/cluster_helpers.py
alex-l-kong: Yup.

ark/phenotyping/cluster_helpers.py
cliu72: Can we add a comment about that, for the future if people are confused where these values came from?

cliu72:
This looks good to me. Did you have any additional questions about testing? I didn't look at it that hard but it looks fine to me

alex-l-kong:
> This looks good to me. Did you have any additional questions about testing? I didn't look at it that hard but it looks fine to me

Nope, everything else looks straightforward.

ngreenwald (approved):
Looks good

ngreenwald:
Will wait to merge this in until #823 is resolved

alex-l-kong:
Pixel subsetting merging complete and tested.

Issue #845, opened 2022-11-30
Pixel clustering/cell clustering file cleanup


cliu72:
**Is your feature request related to a problem? Please describe.**
Just some small miscellaneous stuff I noticed when testing the pixel/cell clustering notebooks.
- [X] I think we talked before about removing the prefix from the individual filenames (and only using the prefix for the folder name). They have been removed from most of them, but they still exist on some files: *_channel_norm.feather, *_pixel_norm.feather. Related to this, I don't think the `create_pixel_matrix` function needs the `pixel_cluster_prefix` argument.
- [X] Related to the above point, currently all the individual filenames for cell clustering still include the prefix. Can we remove those?
- [x] Currently, the first 99.9 normalization file is called *_channel_norm.feather and the second 99.9 normalization is called post_rowsum_chan_norm.feather. Can we make these more similar and intuitive - maybe channel_norm.feather and channel_norm_post_preprocessing.feather - or something like that?
- [x] Related to the point above, the two 99.9 normalization files have different formats. The first one (*_channel_norm.feather) has one column for channels, and a second column for norm values, while the second one (post_rowsum_chan_norm.feather) has a different column for each marker - can we make these two files consistent?
- [x] The threshold we use to filter out low expressing pixels is currently called *_pixel_norm.feather. I think this is slightly confusing because we're not really normalizing using this value. We're using it as a threshold. Can we call it 'pixel_threshold.feather' or something like that? Similarly, in this file, the name of the column is "pixel_norm_val" - can we change this to "pixel_threshold_val"?
- [x] Can we change the name of the weights file from "pixel_weights.feather" to "pixel_som_weights.feather"? Same for the cells - can we change to "cell_som_weights.feather"?
- [x] I don't think it's necessary to save the "pixel_som_to_meta.feather" file. That information is already captured in "pixel_channel_avg_som_cluster.csv." Same for *_cell_clust_to_meta.feather.
- [x] Can we change *_cluster_counts_norm.feather to *_cluster_counts_size_normalized.feather?
- [x] Can we save *_weighted_cell_channel.csv as a feather file, just to be consistent? *_cluster_counts.feather and *_cluster_counts_norm.feather have basically the same dimensions.
- [x] It seems to me that *_cell_mat.feather and *_cluster_counts_norm.feather have basically the same info, but *_cell_mat.feather has an additional columns for the clusters, I don't think we need the count info in multiple places - maybe just append cluster info to "cluster_counts_norm.feather"?
[...]

PR #863, opened 2022-12-14
Integrate `pyFlowSOM` with Pixie

### LINKED ISSUE #688: Supposed file corruption during Pixel Clustering step 2.2: Assign pixel SOM clusters
Dear ark team,

I was trying out the pixel clustering notebook out today and ran into an issue I don't know how to resolve.

After successfully training the pixel SOM (step 2.1), trying to run the next cell first gives the output "The data for FOV <fovname> has been corrupted, removing" for all FOVs in the fovs list, then proceeds to throw the key error "pixel_som_cluster", as there are no FOVs left to process.

I tried reinstalling the codebase and updating docker, rerunning the notebook, unfortunately to no avail.
I also checked all involved paths and couldn't find any issues there.

Do you guys have an idea where the issue may lie?
Thanks already for your help!

Best wishes,
Sven

grafik
grafik

### LINKED ISSUE #811: Integration of pyFlowSOM with current `ark` codebase
Zak is almost done updating pyFlowSOM to be fully native in python. Once he does, it will be up to us to integrate it with the existing Pixie codebase. 

We'll need to decide the format of the new python-only functions which call `map_data_to_codes` for both the pixel and cell notebooks, what the tests will look like, and which code (including much of the mocking for R) can be deleted. You should talk with @cliu72 to make a plan so that we're ready to go once Zak is done.

### LINKED ISSUE #833: Port `FlowSOM` functionality over to `pyFlowSOM`
**Relevant background**

`pyFlowSOM` will replace the `FlowSOM` library which requires external `R` scripts to call. This will allow us to remove any R dependencies and files and transfer their functionality over to the respective Python functions. Additionally, this will also allow for full-fledged testing of the process.

**Design overview**

Because `pyFlowSOM` is essentially a trimmed-down native Python-implementation of the existing `FlowSOM` library, we can take a plug-and-play approach with the call to both train the SOM and map cluster values to the data.

There are a few extra elements that will require porting over to the Python:

- 99.9% normalization of the training data prior to training and the full data prior to cluster assignment
- Multiprocessing support for pixel clustering
- File loading and saving

**Code mockup**

_Abstract and derived SOM cluster classes_

As with Consensus Clustering, we define a SOM cluster class (`PixieSOMCluster`) in `clustering_helpers.py` as functionality will be generally shared between pixel and cell clustering. However, because of differences in normalization and data loading processes, it will be best to explicitly define `PixieSOMCluster` as an abstract class and have two other classes, `PixelSOMCluster` and `CellSOMCluster`, implement certain abstract methods in it.

The implementation will follow the general process:

```
from abc import ABC, abstractmethod
class PixieSOMCluster(ABC):
    @abstractmethod
    def __init__(self, columns: List[str], weights_path: pathlib.Path, num_passes: int = 1, xdim: int = 10, ydim: int = 10, lr_start: float = 0.05, lr_end: float=0.01):
        self.columns = columns
        self.weights = None if not weights_path.exists() else feather.read_dataframe(weights_path)  # NOTE: user can pass in an older weights path if desired to skip the training process
        self.num_passes = num_passes
        self.xdim = xdim
        self.ydim = ydim
        self.lr_start = lr_start
        self.lr_end = lr_end
        self.weights_path = weights_path

    @abstractmethod
    def normalize_data(self):
        pass

    def train_som(data: pd.DataFrame):
        self.weights = som(data=data, xdim=self.xdim, ydim=self.ydim, alpha_range=(self.lr_start, self.lr_end))  # imported from pyFlowSOM)
        feather.write_dataframe(self.weights, self.weights_path, compression='uncompressed'

    def assign_som_clusters(self, external_data: pandas.DataFrame) -> pandas.DataFrame:
        # make sure the columns of self.weights and external_data match up per Zak's request
        # call map_data_to_codes (imported from pyFlowSOM) and assign to external_data
        

class PixelSOMCluster(PixieSOMCluster):
    def __init__(self, pixel_data_folder: pathlib.Path, pixel_subset_folder: pathlib.Path, norm_vals_path: pathlib.Path, columns: List[str], weights_path: pathlib.Path, num_passes: int = 1, xdim: int = 10, ydim: int = 10, lr_start: float = 0.05, lr_end: float=0.01):
        super().__init__(columns, weights_path, num_pases, xdim, ydim, lr_start, lr_end, weights_path)
        self.norm_data = feather.read_dataframe(norm_vals_path)
        self.pixel_data_folder = pixel_data_folder  # the unsubsetted data folder
        self.train_data = # load and combine all the .feather files in pixel_subset_folder here

    def normalize_data(self, data: pandas.DataFrame):  # NOTE: this function will also be called per full FOV pixel dataset during cluster assignment
        # use the values in norm_data to normalize train_data

    def train_som(self):
        super().train_som(self.normalize_data(self.train_data))

    def assign_som_clusters(self, external_data: pandas.DataFrame) -> pandas.DataFrame:  # NOTE: for pixel clustering, this will be called for each FOV, the looping/multiprocessing support will be handled by the pixel_consensus_cluster function in pixel_cluster_utils.py
        super().assign_som_clusters(self.normalize_data(external_data))
        

class CellSOMCluster(PixieSOMCluster):
    def __init__(self, cell_data_path: pathlib.Path, columns: List[str], , weights_path: pathlib.Path, num_passes: int = 1, xdim: int = 10, ydim: int = 10, lr_start: float = 0.05, lr_end: float=0.01):
        super().__init__(columns, weights_path, num_pases, xdim, ydim, lr_start, lr_end, weights_path)
        self.cell_data = feather.read_dataframe(cell_data_path)

    def normalize_data(self, data: pandas.DataFrame):
        # normalize cell_data by the 99.9% values

    def train_som(self):
        super().train_som(self.normalize_data(self.cell_data))

    def assign_som_clusters(self) -> pandas.DataFrame:
        super().assign_som_clusters(self.normalize_data(self.cell_data))
```

This functionality will replace `create_pixel_som.R`, `run_pixel_som.R`, `create_cell_som.R`, and `run_cell_som.R`.

_Integration with `train_pixel_som` and `cluster_pixels`_

We can remove `pixel_weights_path` as the class instance directly stores the weights.

Because SOM cluster assignment can be an expensive process for massive pixel cohorts, we have optional multiprocessing support turned on. However, this will need to be ported over to Python. A general framework in `cluster_pixels` will look like:

```
if multiprocess:
    with multiprocessing.get_context('spawn').Pool(batch_size) as fov_data_pool:
        for fov_batch in [fovs_list[i:(i + batch_size)]
                          for i in range(0, len(fovs_list), batch_size)]:
            # load in all the DataFrames and use fov_data_pool to make a multiprocess call to assign_som_labels
            # loop over the processed DataFrames and save
else:
    for fov in fovs_list:
        # load in the DataFrame and call assign_som_labels
        # save
```

_Integration with `train_cell_som` and `cluster_cells`_

We can remove `cell_weights_path` as the class instance directly stores the weights.

Because no multiprocessing is needed for cell SOM clustering, this will be much more plug-and-play for us. We only need to handle additional file loading and saving in these functions.

_Testing_

We can use a small dataset (via `pytest.fixture`) to test this functionality. We don't need to directly test `PixieSOMCluster` because it is meant to be derived from. We can limit tests to `PixelSOMCluster` and `CellSOMCluster` functions with this small sample dataset.

`notebooks_test.py` will handle integration testing on more extensive datasets. Note that we can remove any mocks that exist since functionality will be fully ported over to Python.

Because all R helper scripts are being removed, we can now fully test `train_pixel_som`, `cluster_pixels`, `train_cell_som`, and `cluster_cells`. As with `PixelSOMCluster` and `CellSOMCluster`, we can use a simple fixture (perhaps the same one) to test.

**Required inputs**

No additional inputs will be required. However, instead of the R scripts, inputs will be shipped over to the `PixelSOMCluster` or `CellSOMCluster` instance.

**Output files**

Same as before

**Timeline**
- [ ] A couple days
- [ ] A week

TBD. Depends on Zak's progress. Definitely aiming for before winter closure.

TBD. Depends on Zak's progress. Definitely aiming for before winter closure.

alex-l-kong:

Closes #811. Closes #833. Closes #688. This PR addresses integration of `pyFlowSOM`. See the issues for more details.

See #833.


alex-l-kong:
@ngreenwald tested this on both pixel and cell clustering, verified that both runs seamlessly through.

ngreenwald (changes requested):
Looks good! Will you see if the issue here with large image files is still an issue with our python port? I'm guessing it is, since Zak didn't change the C code, which I believe is where the memory error came from. 

If so, I think it should be straightforward to update the logic to process large chunks at a time rather than the entire image

https://github.com/angelolab/ark-analysis/issues/688

alex-l-kong:
Tested the new Dockerfile locally, confirmed that it `ark` notebooks can run without R dependencies. Building an image and sending it to Dockerhub should be much faster now as the only bottleneck is installing the Python requirements.

alex-l-kong:
> Looks good! Will you see if the issue here with large image files is still an issue with our python port? I'm guessing it is, since Zak didn't change the C code, which I believe is where the memory error came from.
> 
> If so, I think it should be straightforward to update the logic to process large chunks at a time rather than the entire image
> 
> #688

That was an issue with the `.C` module in R and the way the data was being formatted prior to the call to `map_data_to_codes` (now `map_data_to_nodes`), with Cython this shouldn't be an issue. However, I do think it's worth implementing the block-based version @jonhsussman suggested since that will be far more memory friendly if the input data is large enough.

alex-l-kong:
@ngreenwald we'll probably need to spend another PR to optimize the `block_size` logic, though. Will need to do some experimentation to see what the timing tests on that one will look like and at what data sizes will using blocks (as opposed to the whole thing) end up being more optimal.

alex-l-kong:
@ngreenwald just ran a few basic tests in the Docker and it seems that even on a smaller dataset, using a block-based approach by default provides a more optimal experience. It becomes far more apparent as you increase the dataset size.

So I think we should use `block_size` by default. We might still need some calculation to determine the optimal `block_size` value but per @jonhsussman 100 generally works fine.

alex-l-kong:
@ngreenwald apologies, went through a bit of a back-and-forth regarding the `block_size` experiments.

Upon further testing, the block-based optimization isn't as significant as I had thought, we're looking on the order of just a few milliseconds at best, and it performs worse on smaller datasets.

To reiterate, the issue of long vectors is caused by an R interoperability with C issue using the `.C` library, which doesn't pop up in Python. Since this is the case, I think we should close out #688 and leave the current implementation as is.

ngreenwald:
So this issue: 
> due to a hard limit of passing a vector of more than 2^31 elements> 
is specific to R? You confirmed that the python version will work with super large arrays?

alex-l-kong:
> So this issue:
> 
> > due to a hard limit of passing a vector of more than 2^31 elements
> 
> is specific to R? You confirmed that the python version will work with super large arrays?

I did some more stress testing on different OS, let's revert back to our original plan and implement the block-based method. You were right to be skeptical, we should not assume that each OS will have the same definition of max int size for the number of rows.

alex-l-kong:
SOM cluster assignment has been updated to reflect the block-based method per @jonhsussman implementation. That should about wrap up the final request.

ngreenwald:
Okay, and the issues about performance? How much of a difference does this make for predicting a single image worth of pixels?

alex-l-kong:
> Okay, and the issues about performance? How much of a difference does this make for predicting a single image worth of pixels?

Block-based seems slightly better on average.

ngreenwald (commented):
Okay great, once Candace is happy this should be good to go. I assume you'll want to merge in the cleanup PR first?

We can make a new minor release afterwards to deal with the docker issues.

alex-l-kong:
> Okay great, once Candace is happy this should be good to go. I assume you'll want to merge in the cleanup PR first?
> 
> We can make a new minor release afterwards to deal with the docker issues.

Sounds good. I'll merge the cleanup PR in first, then this one after approval.

ark/phenotyping/cluster_helpers.py
```diff
@@ -1,16 +1,310 @@
...
+
+    def train_som(self, data: pd.DataFrame):
+        """Trains the SOM on the data provided and saves the weights generated
+
+        Args:
+            data (pandas.DataFrame):
+                The input data to train the SOM on.
+        """
+        # make sure to run a deterministic SOM for reproducibility purposes
+        som_weights = som(
```
cliu72, line 70: In the `som` function (and correspondingly in init), I think it would be good to have the `rlen` parameter (the number of passes through the dataset - that is something that might be changed).

ark/phenotyping/cluster_helpers.py
```diff
@@ -1,16 +1,310 @@
...
+            data=data.values, xdim=self.xdim, ydim=self.ydim,
+            alpha_range=(self.lr_start, self.lr_end),
+            deterministic=True
```
cliu72, line 73: Hm, so this `deterministic=True` thing is bothering me slightly. In Zak's implementation, he has these lines: https://github.com/angelolab/pyFlowSOM/blob/c7989bdaaaa2f3d23c936df4c0b64c2769198353/pyFlowSOM/cyFlowSOM.pyx#L106-L107

So if `deterministic=True`, that means the nodes are not randomly initialized (it's always the first 100 data points). Perhaps with a lot of iterations, it doesn't matter how the nodes are initialized, but we never tested this. In the og implementation of FlowSOM, it's always random (unless the user specifies the initial nodes): https://github.com/SofieVG/FlowSOM/blob/d5fc07fd81ff1840a4f8791e0e7526d86864cf1a/R/2_buildSOM.R#L190.

I think it would be better to always have a true random initialization of nodes. I think Zak probably made this `deterministic=True` feature for testing. I think in practice, it would still be better to set a seed, and have that be something the users can change if they wish. This would require changing the pyFlowSOM code. I don't think we need this `deterministic` parameter - instead, we should replace it with a `seed` parameter, and just set the seed before initializing the nodes (before this line: https://github.com/angelolab/pyFlowSOM/blob/c7989bdaaaa2f3d23c936df4c0b64c2769198353/pyFlowSOM/cyFlowSOM.pyx#L109). 

Here, if `deterministic=True`, Zak sets a random seed: https://github.com/angelolab/pyFlowSOM/blob/c7989bdaaaa2f3d23c936df4c0b64c2769198353/pyFlowSOM/cyFlowSOM.pyx#L146. We could probably just replace that with whatever seed the user chooses.

I don't think this is that complex (I think @alex-l-kong or @srivarra could do it?). @ngreenwald thoughts?

cliu72: Or alternatively, set the seed before the call to pyFlowSOM and then set `deterministic=False` in the function call. We'd have to confirm that this actually does make pyFlowSOM reproducible though.

srivarra: @cliu72 There are a couple of ways we can do this:

1. In [Python / Cython](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng)
2. Directly in C with the [Numpy C - API](https://numpy.org/devdocs/reference/random/c-api.html)

alex-l-kong: @cliu72 let me test this out. However, if the actual subsetting process is set to a random seed (which it is in `preprocess_fov` prior to calling `create_fov_pixel_data`), we always should be getting the same first 100 rows right?

alex-l-kong: @cliu72 I just tested it out. Setting `np.random.seed` in Python prior to the call to `pyFlowSOM.som` does not randomly initialize the weights. I think this random seed needs to be explicitly set in C.

Using `deterministic=True` does ensure the weights are randomly initialized because the random seed setting is passed off to C. Also it looks like these nodes are initialized to random N data points of `data`:

```
    if nodes is None:
        # If we don't supply nodes, we randomly sample them from the data
        if deterministic == True:
            nodes = data[0: xdim * ydim, :]
        else:
            nodes = data[np.random.choice(data.shape[0], xdim * ydim, replace=False), :]  # this will randomly select `xdim * ydim points from data
```

ngreenwald: I think it'll be worthwhile for Sri and/or Alex to see what it takes to make an updated release for pyFlowSOM, to make sure we can handle it when things come up in the future. Changing the logic would be quite straightforward. I'm not that familiar with C, so I can't quite figure out where this random seed gets translated to the C_seed, and how that actually impacts the nodes.

cliu72: Okay so I just did this test:
```
np.random.seed(100)
pyFlowSOM.som(dat.values, deterministic=False)

np.random.seed(100)
pyFlowSOM.som(dat.values, deterministic=False)
```

And those were returning a different output - meaning that it does indeed randomly initialize the weighs, but setting a random seed isn't doing anything (not what we want). What we want is for those two functions to return the same output.

For this, I think the easiest thing to try is to change cyFlowSOM.pyx itself and not touch the c code (although I will open another issue in the pyFlowSOM repository in a sec that would need to change c code...unrelated to this though). The og version of FlowSOM did not set a random seed in the c code (the random seed was set in R before calling the FlowSOM function). Here, I think we can try adding `np.random.seed(seed)` to the beginning of the som function in cyFlowSOM.pyx (and add the seed as a parameter to the som function). If after testing that with `deterministic=False` and confirming that setting the same seed returns the same output and changing the seed returns a different output, I think we can completely remove having the `deterministic` parameter.

@alex-l-kong Regarding your point that the subsetting process is using a random seed - yes that is true, but I think we should deal with the randomization here independent of the input data fed into the SOM. There could be the case with small datasets (like mine), where I am not subsetting at all. In that case, taking the first N rows of the data might be like the first row of pixels in the image (which are not very diverse). I think we want to keep the functionality of being able to feed the SOM the same exact input data with a different seed and get different results. Overall, I think it would be better to try to recapitulate as much of the original FlowSOM code as possible - and that includes randomizing the nodes (independent of what the input data is).

cliu72: Ugh actually, looking at this a bit more, Zak changed the c code from the original package. In the original c code, https://github.com/SofieVG/FlowSOM/blob/master/src/som.c, there are these lines:

https://github.com/SofieVG/FlowSOM/blob/d5fc07fd81ff1840a4f8791e0e7526d86864cf1a/src/som.c#L14-L15
https://github.com/SofieVG/FlowSOM/blob/d5fc07fd81ff1840a4f8791e0e7526d86864cf1a/src/som.c#L93

It seems like these were removed by Zak. I think we would need to add those lines back in, and adding these back in would allow us to use np.random.seed.

alex-l-kong: From my initial analysis I think the only reason `GetRNGstate` and `PutRNGstate` are called in SofieVG's implementation is because they make an external call to `R_CheckUserInterrupt` every 1024 updates, as well as R's internal `unif_rand` function for random selections. These functions are needed to preserve the original default seed since C is actually calling an R-based package (see https://www.rdocumentation.org/packages/httpuv/versions/1.6.0/topics/getRNGState). However, Zak implemented this logic using just C (https://github.com/angelolab/pyFlowSOM/blob/c7989bdaaaa2f3d23c936df4c0b64c2769198353/pyFlowSOM/flowsom.c#L6-L8), which shouldn't cause any such issues to my knowledge.

It also seems like the `deterministic` param is a recreation of `init` in the `SOM` function of `FlowSOM`, however the deterministic creation of the weights in `FlowSOM` is a lot more sophisticated (https://github.com/SofieVG/FlowSOM/blob/d5fc07fd81ff1840a4f8791e0e7526d86864cf1a/R/2_buildSOM.R#L279-L296), meaning we avoid the issue you brought up of sampling non-diverse pixels. I'm not sure why Zak didn't choose to implement this version and defaulted to a basic top-N sampling for `deterministic=True`.

In either case, I think the main requirement here is ensuring that `pyFlowSOM.som` can take an explicit random seed to initialize the weights in all cases. I'd argue for dropping the `deterministic` parameter and instead pass a `seed` parameter that is then set towards the beginning. In this way, we should be able to use something like:

```
if nodes is None:
    np.random.seed(seed)  # Sri does make a good point about using np.random.default_rng instead per https://albertcthomas.github.io/good-practices-random-number-generators/, this can be a PR on its own
    nodes = data[np.random.choice(data.shape[0], xdim * ydim, replace=False), :]
```

to generate the same starting nodes each time on the same dataset.

cliu72: I told Zak that we never use the `init` parameter and to leave it out of his implementation. In the original version of FlowSOM, the default is `init=FALSE` and we never change that, so the complex initialization function you linked isn't relevant for us (we never use that function).

What we do use is a random initialization of nodes (https://github.com/SofieVG/FlowSOM/blob/d5fc07fd81ff1840a4f8791e0e7526d86864cf1a/R/2_buildSOM.R#L190-191), which is what is currently implemented in Zak's version when `deterministic=False`. I think Zak only implemented `deterministic=True` for testing.

And yes as you outlined and as I talk about above, I think setting a random seed before nodes is the way to go and removing `deterministic`. Your code outline looks good to me.

ngreenwald: Okay great. Is there anything else besides this issue that still needs to be addressed from this PR from you @cliu72 ? We discussed today that it will be easier to merge this one in using the existing release of pyFlowSOM, and then have a separate PR that fixes the randomization issue in a different pyFlowSOM release.

ark/phenotyping/cluster_helpers.py
```diff
@@ -1,16 +1,309 @@
...
+        self.cell_data_path = cell_data_path
+
+        # load the cell data in
+        self.cell_data = feather.read_dataframe(cell_data_path)
+
+        # since cell_data is the only dataset, we can just normalize it immediately
+        self.cell_data = self.normalize_data()
+
+    def normalize_data(self) -> pd.DataFrame:
+        """Uses `norm_data` to normalize `cell_data`
```
cliu72, line 261: Just a small comment error here - `norm_data` isn't actually used in this function (99.9% values are calculated in the function).

ark/phenotyping/cluster_helpers.py
```diff
@@ -1,16 +1,309 @@
...
+        Returns:
+            pandas.DataFrame:
+                `cell_data` with `columns` normalized by the values in `norm_data`
+        """
+        # only 99.9% normalize on the columns provided
+        cell_data_sub = self.cell_data[self.columns].copy()
+        cell_data_sub = cell_data_sub.div(cell_data_sub.quantile(0.999))
+
+        # assign back to cell_data
+        self.cell_data[self.columns] = cell_data_sub
```
cliu72, line 272: My python oos might just be a little rusty - is it necessary to assign `self.cell_data` here, and then return it, and then assign `self.cell_data` in the init function above (line 258)? To mirror the `normalize_data` function in the pixel version, maybe we can just return cell_data_sub here and assign it to `self.cell_data` in the init function.

ark/phenotyping/pixel_cluster_utils.py
```diff
@@ -923,36 +920,60 @@ def train_pixel_som(fovs, channels, base_dir,
...
+    fov_path = os.path.join(pixel_data_path, fov + '.feather')
+
+    # read in the fov data with SOM labels
+    try:
+        fov_data = feather.read_dataframe(fov_path)
+    # this indicates this fov file is corrupted
+    except (ArrowInvalid, OSError, IOError):
+        return fov, 1
+
+    # assign the consensus labels to fov_data
```
cliu72, line 963: Comment error here - these aren't consensus labels (just SOM labels)

ark/phenotyping/pixel_cluster_utils.py
```diff
@@ -923,36 +920,60 @@ def train_pixel_som(fovs, channels, base_dir,
...
+    fov_data = pixel_pysom_obj.assign_som_clusters(fov_data)
+
+    # resave the data with the meta cluster labels assigned
```
cliu72, line 966: Same thing here, comment error - these aren't meta cluster labels (just SOM labels)

ark/phenotyping/cell_cluster_utils.py
```diff
@@ -394,8 +394,8 @@ def train_cell_som(fovs, channels, base_dir, pixel_data_dir, cell_table_path,
                    pixel_cluster_col='pixel_meta_cluster_rename',
                    pc_chan_avg_name='pc_chan_avg.csv',
                    som_weights_name='cell_som_weights.feather',
-                   weighted_cell_channel_name='weighted_cell_channel.feather',
-                   xdim=10, ydim=10, lr_start=0.05, lr_end=0.01, num_passes=1, seed=42):
+                   weighted_cell_channel_name='weighted_cell_channel.csv',
```
cliu72, line 397: Why was `weighted_cell_channel_name` changed back to a CSV here? It was addressed in this issue https://github.com/angelolab/ark-analysis/issues/845 to have it be a feather file (for consistency). The code assumes that it is a feather file (ex. lines 507-509, line 115 in this file).

ark/phenotyping/cluster_helpers.py
cliu72: Added a few more small comments

ark/phenotyping/cell_cluster_utils.py
alex-l-kong: Might have been an artifact of resolving a merge conflict, that was a real messy one 😅.

ark/phenotyping/cluster_helpers.py
alex-l-kong: Good catch. I was hoping to avoid having ugly type hints for the checker but think I found a way to easily resolve that.

alex-l-kong:
Everything except the randomization issue (which will be handled in a separate PR) has been addressed and tested.

Dockerfile
```diff
@@ -3,62 +3,13 @@
...
-# install flowsom
-# RUN R -e "library(BiocManager); BiocManager::install('FlowSOM')"
-RUN R -e "library(devtools); devtools::install_github('angelolab/FlowSOM', upgrade = FALSE, upgrade_dependencies = FALSE)"
+# install gcc
+RUN apt-get install -y gcc
 
-# terminate Docker build if FlowSOM fails to import
-RUN R -e "library(FlowSOM)"
+# # install cmake (needed for nloptr)
+# RUN apt-get install -y cmake
```
ngreenwald, line 12: Is this commented stuff no longer needed?

ngreenwald (changes requested):
Looks good, just a question about the docker

Dockerfile
alex-l-kong: Good catch, seems like this one's a straggler.

cliu72 (approved):
Looks good to me

PR #878, opened 2023-01-10
Load Utils - Expanding Acceptable File Types

### LINKED ISSUE #866: Have `load_imgs_from_tree` also look for .tif files
Could we make the `load_imgs_from_tree` function also read in ".tif"? I think it's fine if all image files that are being written by ark are ".tiff", but I think it would be helpful to be able to read in images that are ".tif." Especially for my paper, we are marketing Pixie as being very generalizable. I think this would be (hopefully) an easy thing to add back in. I'm only requesting that the 2 Pixie notebooks be able to read in ".tif" files, not the whole pipeline.

To me, it doesn't seem like it would be that bad, since the function is already reading in jpeg and png. Seems like we would just need to change these lines:
https://github.com/angelolab/ark-analysis/blob/a98648c8f3bd11fd713be3b41eb885bf5058d93f/ark/utils/load_utils.py#L134 and
https://github.com/angelolab/ark-analysis/blob/a98648c8f3bd11fd713be3b41eb885bf5058d93f/ark/utils/load_utils.py#L140
to look for ".tif" files as well.

Similarly to how `load_imgs_from_tree` takes in jpeg/pngs, have it take in tif files.

Alex suggested adding something in the beginning to turn everything into "tiffs", but I don't think it's necessary since we already have the ability to read in jpegs/pngs.

srivarra:

Closes #866. Users will now be able to read in files ending in "tif" (one 'f').

Used the `EXTENSION_TYPES` constant to generalize over all major image formats ("tif", "tiff", "png", etc...)

**Remaining issues**

N/A.

### DIFF
--- a/ark/utils/load_utils.py
+++ b/ark/utils/load_utils.py
@@ -1,12 +1,13 @@
 import os
-import warnings
 import re
+import warnings
 
+import natsort as ns
 import numpy as np
 import skimage.io as io
 import xarray as xr
-import natsort as ns
 
+from ark.settings import EXTENSION_TYPES
 from ark.utils import io_utils as iou
 from ark.utils import misc_utils
 from ark.utils.tiff_utils import read_mibitiff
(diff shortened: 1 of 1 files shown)


cliu72 (approved):
Looks good to me

PR #884, opened 2023-01-18
Remove zero column pixel cluster columns from cell clustering pipeline

### LINKED ISSUE #882: 3_pixie_cluster_cells notebook: Pixel meta cluster that doesnt map into any cell
**Describe the bug**
Pixel clustering shows a meta cluster with ~3000 pixels (image attached). When proceeding to the pixie cluster cell notebook, cluster_counts_size_norm.feather file generated has NAs causing cell consensus clustering to output error (input X contains NaN). Going back to pixel clustering notebook and merging this cluster with a different cluster and coming back to cell clustering resolves the issue.

**Expected behavior**
ignore or delete any NaN columns from relevant files when going into cell consensus clustering. 

**To Reproduce**
Can provide test dataset and parameters if required. 

Please either copy/paste or screenshot all of the code you ran which produced the error. In addition, copy the full error message. 
Screenshot 2023-01-17 at 10 32 29 AM

Screenshot 2023-01-17 at 12 41 29 PM

Screen Shot 2023-01-13 at 5 24 25 PM
Screen Shot 2023-01-13 at 5 24 42 PM

alex-l-kong:

Closes #882. Occasionally, there will be pixel cluster expression columns consisting of all zeros for cell clustering. This will happen when certain pixel clusters are not expressed in any cell of a cohort. During column normalization, this will set all the values to NA and subsequently throw errors during SOM training.

Instead, training should proceed as normal with these columns removed.

In `create_c2pc_data`, which generates the input data to the cell SOM, we add a check at the end that removes all-zero expression columns prior to returning. This will apply to both the unnormalized version (which is used for weighted cell channel calculation) and normalized version (the SOM input data).


alex-l-kong:
@cliu72 tested that this works. Tagged @HPiyadasa for additional testing purposes too.

cliu72 (commented):
This looks good to me. @HPiyadasa can you check if this fix works?

cliu72 (changes requested):
I'm not sure if @HPiyadasa had a chance to test yet, but I just did a small test, and now I'm getting this error. I manipulated the data so the CD31 pixel cluster was outside of all cells - seems like the appropriate column was removed from the cell count table, but it's now running into an issue in `compute_p2c_weighted_channel_avg`, since that function checks that the cell count table and the pixel cluster average table have the same columns. Before this check, we can probably subset both tables to only the overlapping columns to avoid this error.

image

alex-l-kong:
@cliu72 updated to filter `cell_counts_cluster_ids` on the existing pixel cluster values in `compute_p2c_weighted_cell_channel_avg`. Let me know if this new version works.

alex-l-kong:
@cliu72 just wanted to follow up on this one, let me know if you have any additional comments!

cliu72 (approved):
Looks good!

PR #885, opened 2023-01-19
Integrate existing Pixie cell clustering process with generic cell clustering process

### LINKED ISSUE #850: Abstract Pixie to work with different inputs
The current Pixie workflow assumes that input data was created by the pixel clustering notebook. This makes it difficult to use with other types of clustering preprocessing

Abstract the loading logic and remove any hard-coded parameters that are specific to pixel clustering. Instead, the cell clustering notebook should be able to flexibly handle a cell table with cell-level annotations from a variety of sources.

### LINKED ISSUE #881: Ensure cell clustering process can be run with a variety of cluster types
**For future reference**

Use draw.io for flowcharts lol.

**Relevant background**

The cell clustering pipeline currently only supports pixel clusters generated from `2_Pixie_Cluster_Pixels` (`pixel_som_cluster` or `pixel_meta_cluster_rename`). This has worked fine because `3_Pixie_Cluster_Cells` is designed to run immediately afterwards.

However, increasingly users have needed to run cell clustering on inputs outside of pixel clustering. The cell clustering pipeline will need to be generalized in order to support this.

**Design overview**

Generalizing cell clustering will require changes to multiple components.

We're currently presetting the following variables using `cell_clustering_params` saved from `2_Pixie_Cluster_Pixels`:

- fovs
- channels
- segmentation_dir
- seg_suffix
- pixel_data_dir
- pc_chan_avg_som_cluster_name
- pc_chan_avg_meta_cluster_name

We should still provide support for this, however we should also give the user flexibility to use custom inputs.

_Pixel SOM input data (pixel_data_dir)_

Currently, the input data to train the SOM is generated using a helper function. However, the user will need to be specifying this data on their own.

The pixel clustering SOM input data is formatted as such:

| cell_size | fov | segmentation_label | pixel_som_cluster_0 | pixel_som_cluster_1 | ... |
| ---------| --- | -------------------- | --------------------- | ------------------- | --- |
| ... | fov0 | 0 | ... | ... | ... |
| ... | fov0 | 1 | ... | ... | ... |
| ... | ... | ... | ... | ... | ... |

For the purposes of generalizability, we can replace columns `pixel_som_cluster_0`, `pixel_som_cluster_1`, etc. with `user_defined_col_0`, `user_defined_col_1`, etc. Users will need to provide data that contains, at the very least, the `fov`, `segmentation_label`, and these user defined column data.

In the case of custom inputs for cell clustering, the user will need to specify these column names on their own, as they will not have consistent a prefix naming scheme.

This change will mean that if the full Pixie pipeline is run, the cell SOM inputs should be created as an explicit preprocessing step to training the actual cell SOM (as opposed to being implicitly built into the training function). This will allow the training function to use a consistent form of input data for both the full Pixie and the generalized pipeline.

_Cluster channel average expression file_

If the user is providing generalized inputs, they should not be required to provide this file, as this is based on pixel clusters only and has no reasonable interpretation outside of this context.

_Weighted cell channel average file_

If the user is running this as part of the full Pixie pipeline, this file should also be generated as an explicit preprocessing step before training the cell SOM (and not inherently built into that function, as it is now).

Users running this with generalized inputs will not need to generate this file.

_Deprecating cell clustering params (for those starting with cell clustering)_

Users starting directly from cell clustering should not be required to make their own cell clustering params JSON file.

Users should instead provide the path to their custom data, formatted like the table above. From this file, we can infer their list of FOVs and the columns to subset on (throwing an error if the data doesn't follow a consistent column naming scheme). Users will still need to provide a segmentation directory, segmentation suffix, and cell table path on their own.

**Code mockup**

For this process, we should make a new notebook called `3b_Generic_Cluster_Cells`. This is a tentative name subject to change.

_Data preprocessing (`3a_Pixie_Cluster_Cells`, tentative name)_

Users running the full Pixie pipeline process can use the outputs directly from `2_Pixie_Cluster_Pixels`, however the generation of inputs into the cell SOM pipeline will be different.

To support both pipelines (full Pixie and starting from cell clustering), `crete_c2pc_cell_data` should directly take the input data as opposed to generating it itself. In `3a_Pixie_Cluster_Cells`, this can happen with:

```
# user will set variables cell_table_path, pixel_cluster_col, and cell_som_input_name prior

cluster_counts, cluster_counts_size_norm = create_c2pc_data(
    fovs, pixel_data_dir, cell_table_path, pixel_cluster_col
)

# define the count columns found in cluster_counts_norm
cell_cluster_columns = cluster_counts_size_norm.filter(
    regex=f'{pixel_cluster_col}.*'
).columns.values

feather.write_dataframe(
    cluster_counts_size_norm,
    os.path.join(base_dir, cell_som_input_name),
    compression='uncompressed'
)
```

The variable `cell_cluster_columns` is needed to support a generic list of columns a user should now be able to specify to `train_cell_som` identifying the features needed for training (required for generalized cell SOM inputs). This will no longer be consistent with the `pixle_cluster_col_prefix` used in the past (which ran a regex find of all matching columns), so these arguments will need to be removed and replaced with the explicit set of columns passed.

Additionally, the weighted cell channel average file should also be precomputed. This can be done by using the unnormalized `cluster_counts`:

```
# user will be setting variables pixel_cluster_col and weighted_cell_channel_name prior,

# depending on which pixel_cluster_col is selected, choose the pixel channel average table accordingly
if pixel_cluster_col == 'pixel_som_cluster':
    pc_chan_avg_name = pc_chan_avg_som_cluster_name
elif pixel_cluster_col == 'pixel_meta_cluster_rename':
    pc_chan_avg_name = pc_chan_avg_meta_cluster_name

weighted_cell_channel = compute_p2c_weighted_channel_avg(
   pixel_cluster_channel_avg, channels, cluster_counts,
   fovs=fovs, pixel_cluster_col=pixel_cluster_col
)

feather.write_dataframe(
    weighted_cell_channel,
    os.path.join(base_dir, weighted_cell_channel_name),
    compression='uncompressed'
)
```

Users can then pass `weighted_cell_channel_name` as normal.

_Parameter setting (`3b_Generic_Cluster_Cells` tentative name)_

The new notebook for the generic cell clustering process will be structured similarly, however, one major difference will be in the way the inputs are set.

Users will need to provide an explicit path to their input data:

```
cell_som_input_data = '/path/to/input/data'
```

After this is loaded in, we can then determine the list of FOVs using:

```
cell_som_input_data['fov'].unique()
```

Users will need to provide their own list of columns to use:

```
cell_som_cluster_columns = [...]
```

They will also need to specify their own `segmentation_dir`, `seg_suffix`, and `cell_table_path`.

_Changes to the cell clustering process_

- `train_cell_som`: should now take `cell_som_input_name` as an argument to specify the input file. It should no longer take `pixel_data_dir`, `pc_chan-avg_name`, `cluster_counts_name`, `cluster_counts_size_norm_name`, or `pixel_cluster_col` as explicit arguments. `channels` should be renamed to `cell_som_cluster_columns`.
- Any function that takes `pixel_cluster_col` or `pixel_cluster_col_prefix` arguments to determine columns to subset on should now just take an explicit list of `cell_som_cluster_columns`.
- Any function that takes `weighted_cell_channel_name` should support a `None` value for this param. Generic cell cluster inputs will not have a good interpretation for average weighted channel expression which is specific to pixel clusters. When these functions are called from notebook `3b`, they should set this param to `None`, causing the average channel expression computations to be skipped
- `3b_Generic_Cluster_Cells`: the weighted cell channel heatmaps should not be generated per the previous point.

_Example dataset changes_

We'll need to add sample input data for notebook `3b`. This should consist of a custom `input_data` file, a `cell_table`, and `segmentation_data`.

_Notebook test changes_

For `3a`, ensure that the preprocessing cells get run. Changes should not need to be significant. The test class for this notebook should be renamed accordingly.

For `3b`, ensure that each cell runs with the example dataset for this notebook. We'll need a new test class for this.

_Updated function testing_

Most significant change here will be ensuring we define the full set of columns to pass in place of `pixel_cluster_col_prefix`. Additionally, conditionals where `weighted_cell_channel_avg` needs to handle the `None` case will also need additional testing.

**Required inputs**

_3a_Pixie_Cluster_Cells_

The existing inputs should integrate smoothly.

_3b_Generic_Cluster_Cells_

The user will need to explicitly specify their cell SOM input data, cell table, segmentation dir, and segmentation suffix.

**Output files**

_2_Pixie_Cluster_Pixels_

This will now generate the cell SOM input data and weighted cell channel average file on top of the cell clustering parameters file.

_3a_Pixie_Cluster_Cells_

Same as before, except where they will be generated may be different.

_3b_Generic_Cluster_Cells_

Mostly the same, with the exception of heatmaps, which will not be displayed.

**Timeline**
- [ ] A couple days
- [ ] A week

January 31

(I'll can have the code without tests done earlier if Lorenz needs it before then)

Early February

alex-l-kong:

Closes #850. Closes #881. The existing Pixie pipeline will need some modifications to support generic cell clustering.

Main change is the addition of a preprocessing step outside of `train_cell_som`. This will help generate the cell clustering specific files (SOM input data based on pixel clusters and weighted cell channel expression) independent of the eventual notebook for generic cell clustering.

Additionally, it's better to pass the full list of expression columns to use for cell clustering as opposed to inferring them from a pixel cluster prefix. This will also make integration with generic cell clustering easier.


alex-l-kong:
Tested that the updated pipeline works. The following changes have been made:

1. Preprocess `cluster_counts_size_norm` and `weighted_cell_channel` prior to `train_cell_som`
2. Update the parameters accordingly
3. Pass the full list of expression columns to use for SOM training (as opposed to inferring from `pixel_cluster_col_prefix`)

ngreenwald (commented):
Looks good to me, will wait for Candace to weigh in. 

After she's happy with this, I think opening a separate branch off of this one makes sense to begin constructing the notebook 3b. It will probably be the case that as you put that notebook together you realize that additional things may need to change

templates/3_Pixie_Cluster_Cells.ipynb
```diff
@@ -189,16 +189,7 @@
    "cell_type": "markdown",
```
cliu72, line 189: <p>I think it would still be helpful to save the version of this table that is not size normalized (i.e. cluster_counts).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/885/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

cliu72 (approved):
Looks good!

alex-l-kong:
We will definitely need to open a separate branch off of this to address the case where the user wants to re-run generic cell clustering with new columns specified. This is a similar issue that #903 addressed

alex-l-kong:
Per @cliu72 and @ngreenwald I've renamed the notebook so it doesn't interfere with any reviewers. We'll merge this in and open a PR for any issues that do pop up in the future.

ngreenwald (changes requested):
@JLrumberger will you try out the new notebook with the network output from a subset of the TONIC data to see how it looks? And also see if any issues come up? You can set up a time to meet with @alex-l-kong to go over it and make sure everything looks good.

templates/3_Pixie_Cluster_Cells.ipynb
cliu72, line 189: <p>Second and third sentence here are redundant.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/885/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

cliu72 (commented):
Looks good to me. Just one small comment thing.

alex-l-kong:
@ngreenwald per our meeting today, we'll get this merged in for the time being, then have @JLrumberger open up any subsequent issues that come up with his runs.

src/ark/settings.py
```diff
@@ -60,4 +60,4 @@
 
 # Switch it from `main` to the commit ID on HuggingFace to test a
 # specific version of the Example Dataset
-EXAMPLE_DATASET_REVISION: str = "main"
+EXAMPLE_DATASET_REVISION: str = "43041013a7ae4abf64c7b8b6c6a8bc798f3c7318"
```
ngreenwald, line 63: Just need to swap this

ngreenwald (approved):

PR #892, opened 2023-01-24
Add notebook to run generic cell clustering

### LINKED ISSUE #853: Pull out cell clustering summary file generation into separate functions

Per #799, we've separated out the summary file generation for pixel clustering as this was an issue for users. While not as crucial for cell clustering, aesthetically it would look more consistent if we apply a similar procedure here too.

As with pixel clustering, define in `cell_cluster_helpers.py` functions to `generate_som_avg_files` and `generate_meta_avg_files`.

alex-l-kong:

Addresses #881. Closes #853.

This adds the generic cell clustering notebook to the pipeline, which allows users to start at step 3 without needing to run the full Pixie pipeline prior.

Only additional functionality needed internally in `cell_cluster_utils.py` is the ability to skip weighted cell channel processing steps.


alex-l-kong:
@ngreenwald @srivarra only thing left before review is to check in on the example flowchart. I'm not the most artistically strong, so would appreciate any inputs on how to best organize it after adding this option in!

Alternatively, we can choose to leave it out and allow the `README.md` text to do the talking.

ngreenwald:
We can wait on the flow chart, it's not yet clear how this fits into the overall workflow, will depend on how we end up running the model. I would leave it out of the readme for now, and just have a brief explanation at the top of the notebook. Once we decide how to integrate it, we'll update all the documentation.

ark/phenotyping/cell_cluster_utils.py
```diff
@@ -477,17 +475,23 @@ def cluster_cells(base_dir, cell_pysom, cell_som_cluster_cols,
     if cell_pysom.weights is None:
         raise ValueError("Using untrained cell_pysom object, please invoke train_cell_som first")
 
+    # non-pixel cluster inputs won't be cell size normalized
```
ngreenwald, line 478: Where does size normalization happen? The generic data shouldn't automatically be size normalized

ark/phenotyping/cell_cluster_utils.py
```diff
@@ -477,17 +475,23 @@ def cluster_cells(base_dir, cell_pysom, cell_som_cluster_cols,
...
 
-    misc_utils.verify_same_elements(
-        enforce_order=True,
-        cluster_counts_size_norm_columns=cluster_counts_size_norm.columns.values,
-        cell_weights_columns=cell_pysom.weights.columns.values
+    # handles the case if user specifies a subset of columns for generic cell clustering
+    # NOTE: CellSOMCluster ensures column ordering by using the preset self.columns as an index
+    misc_utils.verify_in_list(
+        cell_weights_columns=cell_pysom.weights.columns.values,
+        cluster_counts_size_norm_columns=cluster_counts_size_norm.columns.values
```
ngreenwald, line 494: For lines 506-onwards, since this isn't necessary for the generic clustering, we should pull this out into its own function or add a flag

ark/phenotyping/cell_cluster_utils.py
```diff
@@ -649,47 +678,49 @@ def cell_consensus_cluster(fovs, channels, base_dir, cell_som_cluster_cols, max_
...
-    cell_meta_cluster_channel_avg = compute_cell_cluster_channel_avg(
-        fovs,
-        channels,
-        base_dir,
-        weighted_cell_channel_name,
-        cluster_counts_size_norm_name,
-        'cell_meta_cluster'
-    )
+        # compute the weighted channel average expression per cell meta cluster
+        print("Compute average weighted channel expression across cell meta clusters")
```
ngreenwald, line 709: I think part of the reason the code complexity is high comes from the clustering functions also including the pre- and post-processing. If you think it'll create more headaches, let me know, but to me it seems like things would be simpler if these steps were pulled out into separate functions. For example, computing the averages across cells.

ngreenwald (changes requested):
See below, overall looking quite good. The preprocessing and postprocessing for generic clustering won't be the same as for Pixie, and won't always be the same depending on the inputs

templates/3b_Generic_Cluster_Cells.ipynb
```diff
@@ -0,0 +1,665 @@
+{
```
ngreenwald, line 1: <p>We should move segmentation_dir and tiff_dir definitions to the end of the notebook, since they're only needed for visualization</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/892/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

ngreenwald, line 1: <pre class="ql-syntax" spellcheck="false">Line #8.    cell_som_input_data = pd.read_csv(cell_som_input_name)
</pre><p>We should either put the file I/O in a separate step where the user can specify the extension, or abstract it within the train_som function so that it can support CSV files directly</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/892/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

ngreenwald, line 1: <pre class="ql-syntax" spellcheck="false">Line #8.    cell_som_cluster_count_avg_name = os.path.join("pixie", cell_output_dir, 'cell_som_cluster_count_avg.csv')
</pre><p>Number of SOM clusters isn't relevant for generic clustering, we should get rid of these. </p><p> </p><p>The useful visualization we'll want for this is the average value of each channel for the cells in each cluster</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/892/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

ngreenwald, line 1: <p>Anything that says pixel SOM/meta cluster in this notebook is not relevant anymore, and should be updated to state what's actually being computed if its relevant, or removed if its not. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/892/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

ngreenwald, line 1: <p>Same here. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/892/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

ark/phenotyping/cell_cluster_utils.py
alex-l-kong: Yeah I have an issue in #853 to address this. We can address that here.

ark/phenotyping/cell_cluster_utils.py
alex-l-kong: This happens in `create_c2pc_data`, a step that is only run in `3_Pixie_Cluster_Cell`. This step is bypassed in `3b_Generic_Cluster_Cells` and thus the SOM inputs won't be size normalized for this.

ark/phenotyping/cell_cluster_utils.py
alex-l-kong: Agreed we can pull this out into its own function. However, we still need to compute average number of each cell SOM cluster per cell since these are the inputs for ConsensusClustering.

templates/3b_Generic_Cluster_Cells.ipynb
alex-l-kong: <p>I'm a little confused since <code>cell_som_cluster_count_avg_name</code> is the input to consensus clustering and the metaclustering step.</p><p> </p><p>Also, when you say average value of each channel for the cells in each cluster I'm assuming you mean the average values per cell of each expression column the user specifies? Does this need to be done per cell SOM/meta cluster as well or just per cell?</p>

alex-l-kong: <p>Were we thinking about using the average value of each expression column per cell or cell SOM cluster as the inputs instead to ConsensusClustering?</p>

ark/phenotyping/cell_cluster_utils.py
ngreenwald: Can you explain what the inputs to consensus clustering are in the current notebook 3, not this version? What's the shape, columns, etc?

alex-l-kong: The inputs to consensus clustering is a table in the following format:

| cell_som_cluster | pixel_meta_cluster_1 | pixel_meta_cluster_2 | ... | pixle_meta_cluster_n | count |
| --| -- | -- | -- | -- | -- |
| 1  | ... | ... | ... | ... | ... |
| 2 | ... | ... | ... | ... | ... | 
| 3 | ... | ... | ... | ... | ... |
| ... | ... | ... | ... | ... | ... |
| 100 | ... | ... | ... | ... | ... |

Each entry is the average number of pixel meta clusters found per cell SOM cluster. Training is done on these average count columns, the `count` column is needed to visualize the barchart at the top of the remapping step.

The general idea is we need, for each cell SOM cluster, the average expression of the columns used for training. I'm starting to understand the concern now, if the original expression columns are prediction values instead of counts, the concept of "average predicted probability per cell SOM cluster" might not make as much sense.

Let me know what you think. Maybe @cliu72 and/or @JLrumberger can comment on this further.

ngreenwald: Yes exactly. The equivalent here would be the average value of each column in each cell cluster, rather than the average count of each column.

alex-l-kong: @ngreenwald the existing functionality already computes this. We just need to change the nomenclature used in functions and variable names to avoid confusion.

ngreenwald: Sounds good

templates/3b_Generic_Cluster_Cells.ipynb
alex-l-kong: We can definitely include support for both CSV and feather files, not a huge fan though because it becomes quite bit messy to do so (we have several places where we read and write to the cell SOM input file, currently assumed to always be `.feather`).

I like option 1 the most, where the user specifies the extension in the notebook. I still think we should do the `.csv` to `.feather` conversion in the notebook, because then we can also reset the extension of the cell SOM input file across all functions to `.feather` as well (rather than have to handle it on a case-by-case basis in each function).

All this will be much easier to handle we convert the entire process to a class-based system.

alex-l-kong: Addressed by ensuring naming conventions are clear.

ark/phenotyping/cell_cluster_utils.py
```diff
@@ -12,17 +13,17 @@
 from ark.phenotyping import cluster_helpers
 
 
-def compute_cell_cluster_count_avg(cell_cluster_path, cell_som_cluster_cols,
-                                   cell_cluster_col, keep_count=False):
+def compute_cell_cluster_expr_avg(cell_cluster_path, cell_som_cluster_cols,
+                                  cell_cluster_col, keep_count=False):
     """For each cell SOM cluster, compute the average number of associated pixel SOM/meta clusters
```
cliu72, line 18: Comment needs to be updated here as well (remove reference to pixel SOM/meta clusters)

ark/phenotyping/cell_cluster_utils.py
```diff
@@ -426,24 +421,24 @@ def train_cell_som(fovs, base_dir, pixel_data_dir, cell_table_path, cell_som_clu
...
 
     # verify the cell_som_cluster_cols columns provided are valid
     misc_utils.verify_in_list(
         provided_cluster_cols=cell_som_cluster_cols,
-        som_input_cluster_cols=cluster_counts_size_norm_data.columns.values
+        som_input_cluster_cols=cell_som_input_data.columns.values
     )
 
     # define the cell SOM cluster object
     cell_pysom = cluster_helpers.CellSOMCluster(
```
cliu72, line 440: Question for @ngreenwald. In the constructor for `CellSOMCluster`, `normalize_data` is called - which divides each input feature by its 99.9th percentile. Is that something that we always want to do?

ark/phenotyping/cell_cluster_utils.py
```diff
@@ -469,25 +462,29 @@ def cluster_cells(base_dir, cell_pysom, cell_som_cluster_cols,
...
 
+    # non-pixel cluster inputs won't be cell size normalized
+    cols_to_drop = ['fov', 'segmentation_label']
+    if 'cell_size' in cell_pysom.cell_data.columns.values:
+        cols_to_drop.append('cell_size')
+
     # ensure the weights columns are valid indexes, do so by ensuring
     # the cluster_counts_norm and weights columns are the same
     # minus the metadata columns that appear in cluster_counts_norm
     cluster_counts_size_norm = cell_pysom.cell_data.drop(
```
cliu72, line 479: Sorry, I think I'm being redundant with what Noah already said, but I only skimmed through his comments and thought I'd put this here anyways. Probably want to change the variable name of `cluster_counts_size_norm` to avoid the "cluster_counts" vocabulary.

ark/phenotyping/cell_cluster_utils.py
```diff
@@ -604,58 +604,139 @@ def cell_consensus_cluster(fovs, channels, base_dir, cell_som_cluster_cols, max_
...
     )
 
-    # compute the weighted channel average expression per cell SOM cluster
+
+def generate_wc_avg_files(fovs, channels, base_dir, cell_cc,
+                          cell_som_input_data_name='cluster_counts_size_norm.feather',
+                          weighted_cell_channel_name='weighted_cell_channel.feather',
+                          cell_som_cluster_channel_avg_name='cell_som_cluster_channel_avg.csv',
+                          cell_meta_cluster_channel_avg_name='cell_meta_cluster_channel_avg.csv'):
+    """Generate the weighted channel average files per cell SOM and meta clusters.
```
cliu72, line 695: I think we can add a bit more text here, explaining when "weighted channel" makes sense. Something like "When using Pixie pixel clusters to do cell clustering, the number of each pixel cluster in each cell is tabulated. Here, we are multiplying these counts by the average expression profile of each pixel cluster to determine the weighted channel expression per cell."

ark/phenotyping/cell_cluster_utils.py
```diff
@@ -604,58 +604,139 @@ def cell_consensus_cluster(fovs, channels, base_dir, cell_som_cluster_cols, max_
...
+    io_utils.validate_paths([cell_som_input_data_path, weighted_channel_path])
+
+    # if the weighted channel average files exist, skip
+    if os.path.exists(som_cluster_channel_avg_path) and \
+       os.path.exists(meta_cluster_channel_avg_path):
+        print("Already generated average weighted channel expression files, skipping")
+        return
+
     print("Compute average weighted channel expression across cell SOM clusters")
     cell_som_cluster_channel_avg = compute_cell_cluster_channel_avg(
```
cliu72, line 734: It seems to be that `compute_cell_cluster_channel_avg` and `compute_cell_cluster_expr_avg` are doing very similar things (and have very similar names) - any way to just combine those 2?

cliu72 (commented):
Looks pretty good to me, just a few small things

ark/phenotyping/cell_cluster_utils.py
alex-l-kong: We can spin up a combination solution, however there are three reasons I've left them separated:

1. Running `compute_cell_cluster_channel_avg` involves an additional file. This leads to enough components being different between the two (validation, subsetting, etc.) that I'd argue it's better to separate the two to avoid code bloat.
2. Code maintainability: if a future bug fix/patch/etc. involves the averaging step, it'd be (just a bit) easier to isolate the problem to one of these functions, as opposed to a single function that handles multiple cases.
3. Documentation: the way the functions are, it's clear what the column expression and channel averaging functions do and what arguments they take. I fear it'd be more difficult to understand if we combined all of this in one function.

@ngreenwald let me know what you think.

templates/3b_Generic_Cluster_Cells.ipynb
ngreenwald: What about avoiding any writes until the end of the pipeline? Will that require rewriting a ton of stuff? It seems like loading a df, appending to it, and then at the end saving would be simpler, given that the cell table is never going to be so big that we need to read/write parts of it at a time

ark/phenotyping/cell_cluster_utils.py
ngreenwald: I'm not sure. We'll need to wait and see how the probabilities behave for clustering. It's probably a good idea, but we may decide it makes thing worse and want to change this a user-defined step

ark/phenotyping/cell_cluster_utils.py
ngreenwald: I could see it going either way honestly, up to you guys

templates/3b_Generic_Cluster_Cells.ipynb
alex-l-kong: It shouldn't be too bad to implement, especially since we pre-compute `cluster_counts_size_norm` in notebook 3 (the inputs for SOM training). I'll experiment with this a bit today.

alex-l-kong: I'm going to just do this for the SOM input training for now (`cluster_counts_size_norm` in notebook 3, and `cell_som_input_data` in notebook 3b), since this is the only data that will be significantly changed for both SOM clustering, consensus clustering, and remapping.

ark/phenotyping/cell_cluster_utils.py
cliu72: I feel like in my head, 99.9 normalization is meant to help with markers that are always waaaay brighter (or dimmer) than other markers. If the input features are probabilities, they should all be in the same range anyways (like a marker shouldn't always have suuuuper low probabilities - if so, that's probably just a crappy marker and should've been thrown out anyways). I think it might make more intuitive sense (at least to me) to first try it without normalization, and then add it back in later if the clustering looks bad. But anyways, I agree that it could probably wait until we see how the data is behaving.

ark/phenotyping/cell_cluster_utils.py
cliu72: I'm not going to die on this hill, but I think the functionality is so similar that it makes it more confusing for the user that they are two functions (especially with such similar names). Like when I was reviewing the code, I spent quite a bit of time trying to understand why it's 2 separate functions. In my view, code maintainability would actually be easier if it's one function since the functionality is basically the same. Basically all the function does is average over cell som/meta clusters for a set of input columns. The reason why `compute_cell_cluster_channel_avg` takes an additional input is because the clusters and the input features are in two different files - it's only a few lines to first combine those into one table, those lines could easily be moved to before wherever the function is being called.

But like I said, if you truly think it'll be annoying to combine them, we can leave it. But at the very lease, I think the function names should be changed so they're not so similar.

ark/phenotyping/cell_cluster_utils.py
alex-l-kong: Let's leave this on the back burner then. I'll open an issue for this if and when the time comes.

ark/phenotyping/cell_cluster_utils.py
alex-l-kong: @cliu72 how about we make a separate issue for this and revisit it later this month? I do agree it's worth exploring options here, but I think this is extending beyond the scope of this particular PR.

cliu72: At the very least, can we change the name of the functions as a part of this one? `compute_cell_cluster_channel_avg` and `compute_cell_cluster_expr_avg` have basically the same meaning.

cliu72:
Also can we wait to merge this until my paper is accepted? Or remove the "3" from the notebook name until after my paper is accepted? Lorenz or whoever wants to use it can probably just go to this branch or tell them to use the appropriate notebook. I don't want to confuse my reviewers about why we're offering 2 notebooks for cell clustering.

ngreenwald:
Yes, we'll definitely pick a different name in the meantime, we don't want people getting confused

ark/phenotyping/cell_cluster_utils.py
alex-l-kong: Done!

alex-l-kong:
@cliu72 @ngreenwald ok this PR should have addressed most everything so far. Let me know if there are any additional issues to resolve before merging this into #885 (not main). We can create additional PRs off #885 if any more issues come up (ex. adding option to turn off 99.9% normalization).

ngreenwald (approved):
LGTM

PR #894, opened 2023-01-26
Correct 99.9% normalization to 100% normalization for "many-zero" columns

### LINKED ISSUE #890: In cell clustering, 99th percentile normalization is returning NaNs

This is actually Candace on Hadeesha's account lol. @cliu72 

**Describe the bug**
During cell clustering, the 99.9th percentile is being applied across all values (even 0's) - which causes the 99.9th percentile to be 0. So dividing by 0 results in NaN.

In contrast, during pixel clustering, we are first subsetting for only pixels that are >0 - we want to mirror that here.

**Expected behavior**
Exclude 0's when calculating 99.9th percentile when doing cell clustering preprocessing (so there are no NaNs in the cell table).

Relevant line:
https://github.com/angelolab/ark-analysis/blob/main/ark/phenotyping/cluster_helpers.py#L268

**To Reproduce**
Example from Hadeesha's data. Before normalization (ChymaseTryptase for some random cell):
Screenshot 2023-01-24 at 12 22 29 PM

After normalization:
Screenshot 2023-01-24 at 12 23 17 PM

alex-l-kong:

Closes #890. When an expression column for cell Pixie clustering is mostly zero, the 99.9% quantile value may equal 0, which will cause NaNs to appear for normalization. To prevent this, 100% quantile values should be set for those columns.

Add correction in the way the normalization values are set in `CellSOMCluster.normalize_data`. Define the normalization values prior to applying them to `cell_data_sub`.


cliu72 (changes requested):
This definitely solves the error that arises, but actually the correct behavior is for the function to ignore 0's when calculating the 99.9th percentile. For example, this is the mirroring code in `create_pixel_matrix`: https://github.com/angelolab/ark-analysis/blob/e6050fca79daeb9fb9e771f1df4e383ad54afaa6/ark/phenotyping/pixel_cluster_utils.py#L780-L782.

In that function, you are setting all 0's to nan, so the quantile function ignores those numbers. That's what we want here too.

cliu72 (approved):
Looks good

alex-l-kong:
@HPiyadasa can you please approve this if it looks good? Need it so I can officially merge the changes in.

HPiyadasa (approved):

PR #897, opened 2023-02-01
Allow user to set an explicit skip flag for custom regionprops and marker extraction functionality in notebook 1

### LINKED ISSUE #875: Optimize cell table generation code
The current code for generating a cell table is quite slow. Initially, since we only performed this action once, it wasn't a big deal. However, I have now written a couple helper functions that rely on this code, and the runtime is becoming more of an issue

It's not clear to me which part of the code is the slowest part. My suspicion is that it's computing the per-cell [marker counts](https://github.com/angelolab/ark-analysis/blob/main/ark/segmentation/signal_extraction.py#L59), but I'm not sure. 

We should first perform some profiling to see where the problem is. If it's in that section of the code, it's possible that the function @JLrumberger wrote to compute the average model confidence per cell could be used here instead.

alex-l-kong:

Closes #875. Since PyTorch/Tensorflow optimizations to running these extractions will require more changes to the Docker than anticipated, for now we add the functionality to skip these computations.

A flag, `skip_extraction`, can be set for `generate_cell_table`, which gets propagated to the internal extractions, allowing them to avoid running these expensive processes if so desired by the user.

**Remaining issues**

We do need to maintain the `'cell_size'`, `'label'`, `'coords'`, and `'centroid'` features from `regionprops`. The first 3 are needed for indexing and `'centroid-0'` and `'centroid-1'` (generated from `'centroid'`) is needed for spatial LDA. This means there will still be the bottleneck of running `regionprops_table`, since I believe the way it works is it generates all the regionprops features _and then_ subsets on them.

Please comment if there are additional base features we should keep regardless of if `skip_extraction` is set.


alex-l-kong:
Tested with the `skip_extraction` flag and it runs about 70-80% times faster without. The number of cells to assign is an unavoidable bottleneck, but without signal extraction on the example dataset it takes on our worst FOVs 40s (it's about 2-3x slower without `skip_extraction`).

ngreenwald (changes requested):
Skipping the extraction step makes sense for testing purposes to see where the bottleneck is, but for in terms of actual user setting, isn't relevant. The only reason to generate the cell table is to have the extracted intensity values. 

The parameter that would be helpful for the user is the option to skip the custom regionprops functions, which are quite slow. 

My understanding is that all the features in `REGIONPROPS_BASE ` don't contribute materially to the runtime. Is that correct? If so, then having a `skip` flag which sets `REGIONPROPS_SINGLE_COMP ` and `REGIONPROPS_MULTI_COMP` to None would be sufficient. If it's the case that certain features in `REGIONPROPS_BASE` are also slow to compute, then the `skip` flag should also remove any of those slow ones.

alex-l-kong:
@ngreenwald fixed up the pipeline to only remove `regionprops_single_comp` and `regionprops_multi_comp` and leave marker extraction intact.

By removing certain regionprops base calculations (all but `label`, `centroid`, and `coord`), we can trim the single compartment props retrieval time by several orders (up to 10x for our more expensive FOVs). So I'd say the `skip` step should also set `regionprops_base` to just these features in addition to setting `regionprops_single_comp` and `regionprops_multi_comp` to `[]`.

ngreenwald:
Got it. And what if you remove centroid and coord as coord as well?

alex-l-kong:
> Got it. And what if you remove centroid and coord as coord as well?

No significant difference, a few tenths of a second difference at most.

ngreenwald:
great, sounds good

ark/segmentation/marker_quantification.py
```diff
@@ -197,7 +202,10 @@ def compute_marker_counts(input_images, segmentation_labels, nuclear_counts=Fals
         split_large_nuclei (bool):
             controls whether nuclei which have portions outside of the cell will get relabeled
         extraction (str):
-            extraction function used to compute marker counts.
+            extraction function used to compute marker counts
+        skip_extraction (bool):
+            if set, skips both the regionprops and the signal extraction steps regardless of
```
ngreenwald, line 207: This isn't true anymore. We should change the documentation, and also the name of the flag, to reflect the behavior. Maybe `fast_extraction`

ngreenwald (changes requested):
The documentation needs to be updated to reflect the behavior of skip extraction. Will you also add a flag in notebook1 with a description where the user can toggle this on or off?

ngreenwald (approved):
Looks good!

PR #903, opened 2023-02-07
Ensure pixel SOM training reruns if new markers specified

### LINKED ISSUE #887: 2_pixie cluster pixel does not retrain SOM when new channel added
**Describe the bug**
When new channel is added, pixel preprocessing generates "New channels provided: overwriting whole cohort" and reprocesses the data. However going into - create the pixel-level SOM weights section does not regenerate the pixel SOM with new channel and instead get this msg - "Pixel SOM already trained". This results in an error when using pixel SOM weights to assign pixel cluster step. 

"1 value(s) provided for list pixel som weights columns and list pixel data columns are not found in both lists."

**Expected behavior**
Ability to add new channels and recluster
Screenshot 2023-01-23 at 11 32 34 AM
Screenshot 2023-01-23 at 11 32 45 AM
Screenshot 2023-01-23 at 11 32 57 AM

alex-l-kong:

Closes #887. It's currently not possible to re-run `2_Pixie_Cluster_Pixels` in case the user needs to specify a new subset of markers on the same cohort. This is a common occurrence and should be addressed.

In `PixelSOMCluster.train_som`, add a check to verify the existing weights columns (if applicable) and the provided set of columns match. If they do not, this means a new set of columns was specified and `PixieSOMCluster.train_som` should re-run.

### DIFF
--- a/ark/phenotyping/cluster_helpers.py
+++ b/ark/phenotyping/cluster_helpers.py
@@ -190,10 +190,15 @@ class PixelSOMCluster(PixieSOMCluster):
     def train_som(self):
         """Trains the SOM using `train_data`
         """
-        # do not train SOM if weights already exist
+
         if self.weights is not None:
-            warnings.warn('Pixel SOM already trained')
-            return
+            # do not train SOM if weights already exist and the same markers used to train
+            if set(self.weights.columns.values) == set(self.columns):
+                warnings.warn('Pixel SOM already trained on specified markers')
+                return
+
+            # notify the user that different markers specified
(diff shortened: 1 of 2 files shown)


cliu72 (changes requested):
I think it would be a good idea to add this functionality to `train_som` in the cell version as well. I think it could come up in the general cell clustering case (for example, in Lorenz's case if he wants to re-run the training and leave out some markers)

cliu72 (approved):
Looks good!

PR #904, opened 2023-02-07
Ensure z-score slider adjusts colorbar range when changed

### LINKED ISSUE #898: Fix z-score slider for metacluster remapping process
**Describe the bug**

The z-score slider currently won't adjust the min and max z-score bounds.

**Expected behavior**

The min and max values should change. Additionally, the scaling should remain consistent.

**To Reproduce**

Run either pixel or cell clustering notebooks as is.

alex-l-kong:

Closes #898. The z-score slider needs to be able to adjust the colorbar on change.

Re-add the `calibrate` function and call it on `self.normalizer` when `update_gui` invoked. Adjust range of `normalizer` back to `(-1, 0, 1)`.

**Remaining issues**

The normalizer still fails to add fractional ticks correctly (anything with .5 added to it). Additionally, for high z-score values, the tick range may not match exactly and the end ticks get drawn at weird spots.

### DIFF
--- a/ark/phenotyping/cluster_helpers.py
+++ b/ark/phenotyping/cluster_helpers.py
@@ -190,10 +190,15 @@ class PixelSOMCluster(PixieSOMCluster):
     def train_som(self):
         """Trains the SOM using `train_data`
         """
-        # do not train SOM if weights already exist
+
         if self.weights is not None:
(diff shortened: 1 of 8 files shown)


cliu72 (approved):
Looks good to me!

PR #944, opened 2023-03-14
Next Release - v0.5.3

alex-l-kong:

Bumps to v0.5.3. This is specifically for Candace's reviewers.

**Remaining issues**

Some things will need to be added back after Candace's review cycle completes (specifically, the channel and pixel normalization for Pixie preprocessing).


ngreenwald (changes requested):
We should make this a minor release, not a patch, i.e. 0.6.0

PR #950, opened 2023-03-16
Deprecate use of cluster, metacluster, and mc_name outside of remapping

### LINKED ISSUE #846: Standardize column names used in metacluster remapping

Currently we do a lot of column renaming to ensure the data can be processed by `metacluster_remap_gui` properly. This is largely a function of a design choice by Zak when creating it. To simplify the process, we should standardize the column names in `metacluster_remap_gui` so they reflect those of `pixel_cluster_utils.py` and `cell_cluster_utils.py` on completion.

It will still be necessary to internally rename these columns as the metaclustering GUI runs generically for pixels and cells. However, when the remapping file gets generated, they should be renamed back.

Instead of `'cluster'`, `'metacluster'`, and `'mc_name'`, the remapped file should have `'pixel_som_cluster'`, `'pixel_meta_cluster'`, and `'pixel_meta_cluster_rename'` and similarly for cell clusters.

One potential issue would be ensuring that the correct cluster type prefixes these column names. We might be able to accomplish this by using an additional property to keep track of this.

alex-l-kong:

Closes #846. It's much easier for the user to read `pixel_som_cluster` as opposed to `cluster`, `pixel_meta_cluster` as opposed to `metacluster`, and `pixel_meta_cluster_rename` as opposed to `mc_name`. We had used these as a result of standards Zak used, but these names should be reverted outside of the remapping process.

Any function outside of `metacluster_remap_gui` needs to use the new standard of column naming. When the mapping is saved after a click event of `New metacluster` happens, the columns get renamed prior to the save.


alex-l-kong:
The example dataset version will be changed to main just before merging in.

srivarra (approved):
LGTM!

camisowers:
Agreed, adding the `cluster_type` arg is easy to follow.

src/ark/settings.py
```diff
@@ -60,4 +60,4 @@
 
 # Switch it from `main` to the commit ID on HuggingFace to test a
 # specific version of the Example Dataset
-EXAMPLE_DATASET_REVISION: str = "main"
+EXAMPLE_DATASET_REVISION: str = "fe6c548d66125d862273f2e435554d9f6b8730f2"
```
ngreenwald, line 63: Merge when you're ready

ngreenwald (approved):
Looks good! As long as you made sure both notebooks run without any issues should be good to go.

PR #951, opened 2023-03-21
Change os.rename to shutil.move

cliu72:

Apparently Windows doesn't like it when you try to `os.rename` something that is an "open process" (causing pixel clustering functions to fail). After much experimenting, seems like the fix is easy. `shutil.move` seems to do the trick.

Change `os.rename` to `shutil.move`. Tested on my computer, Ben's, and MIBIAN.

**Remaining issues**

None


alex-l-kong (approved):
Looks good

Issue #967, opened 2023-03-30
plot_neighborhood_cluster_result has multiple colors per cell


ngreenwald:
*Please refer to our [FAQ](https://ark-analysis.readthedocs.io/en/latest/_rtd/faq.html) and look at our [known issues](https://github.com/angelolab/ark-analysis/issues) before opening a bug report.*

**Describe the bug**
This function should assign each cell a unique color. However, upon closer inspection, it's clear that many cells are getting assigned multiple colors, but only at the borders. From a distance it's not clear, but a zoom in highlights the problem. 

<img width="994" alt="image" src="https://user-images.githubusercontent.com/13770365/228930651-d30e790c-2730-49a6-95ec-9a7310e34543.png">
<img width="1155" alt="image" src="https://user-images.githubusercontent.com/13770365/228930680-1297964c-9a89-42e1-aa0f-7bec64a12801.png">

This is after running the neighborhood analysis notebook, specifically this cell:
```
# load the masks in and plot them
for fov in subset_neighborhood_fovs:
    neighborhood_mask = load_utils.load_imgs_from_dir(
        data_dir=overlay_out_dir,
        files=[fov + "_neighborhood_mask.tiff"],
        trim_suffix="_neighborhood_mask",
        match_substring="_neighborhood_mask",
        xr_dim_name="neighborhood_mask",
        xr_channel_names=None,
    )

    plot_utils.plot_neighborhood_cluster_result(neighborhood_mask, [fov], k=k, save_dir=overlay_out_dir)
```

camisowers:
It was an issue with plt.imshow() using a custom defined ListedColormap for int values instead of a colormap along a range. By default, imshow uses `interpolation='antialiased'` for boundary points. We can fix this by setting that arg to 'none'.

<img width="712" alt="Screenshot 2023-04-25 at 1 27 04 PM" src="https://user-images.githubusercontent.com/38049893/234395960-afad36f9-7674-4bae-81fc-81d29339b09e.png">
<img width="600" alt="Screenshot 2023-04-25 at 1 27 04 PM" src="https://user-images.githubusercontent.com/38049893/234396144-6ce29389-06f2-4b54-90ff-88f690daaa62.png">

PR #977, opened 2023-04-24
Set Kmeans parameter to prior default

### LINKED ISSUE #968: Potential bug in neighborhood assignments

**Describe the bug**
I'm running into some weird behavior with the neighborhood analysis script. Specifically, it seems like cells with very similar neighborhoods are being assigned to different clusters. 

For example, in the upper right hand corner, all of the blue cancer cells seem to have almost exactly the same neighbors
image

However, they are assigned to different neighborhoods in the output. 
image

I'm not sure if this is related to #967. It could be that the visualization isn't working correctly. However, the heatmap of the clusters roughly lines up with the visual, so I think that's less likely. Not sure exactly what's going on. I think a good first step once #967 is resolved will be to re-run on some previous data and confirm that we still get the qualitatively same clustering results, making sure to re-generate the neighbor_counts, rather than using the previously extracted ones.

camisowers:

Maybe fixes #968. 

Changes `n_init` from 'auto' to 10, to keep the algorithm params consistent with the prior default value.

**Note:** I've pinned alpineer temporarily so that the new multiple tiled stitching adjustments made won't apply until a formal PR is opened. A few of the ark stitching functions will need to be tweaked.

**Remaining issues**

Unrelated to #967. I'm still not quite sure what's going on there.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 dependencies = [
-    "alpineer>=0.1.5",
+    "alpineer==0.1.5",
     "Cython>=0.29,<1",
     "datasets>=2.6,<3.0",
     "dill>=0.3.5,<0.4",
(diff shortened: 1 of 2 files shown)


ngreenwald (approved):
Changed the description to avoid closing it until we've tested it to confirm. 

I agree about pinning alpineer, the goal of having things in different repos is that nothing will update until we want it to.

PR #981, opened 2023-05-01
Mutiple tiled stitching tweaks

### LINKED ISSUE #975: Alpineer updates causing tiled stitching errors

**Describe the bug**
Adjust stitching to reflect alpineer helper function adjustments made in angelolab/alpineer#27.

**Expected behavior**
For `stitch_images_by_shape` to not fail.
Currently, can be solved by pinning `"alpineer==0.1.5"` in pyproject.toml for our current PRs, but we will need to actually adjust the tiled stitching scripts in a new PR.

camisowers:

Closes #975. Adds ability to account for multiple tiles in ark stitching. 

Adjust the output of `alpineer.load_utils.load_tiled_img_data` to be a list of tuples. Loop through the various tuples, since each should indicate an individual tiled grid.

Also angelolab/alpineer#25 changed the error message for empty lists in `verify_in_list`, so the `spatial_lda_utils_test` was adjusted.

**Remaining issues**

- [x] Add multiple tile testing


camisowers:
@srivarra Could use your input on how `alpineer.misc_utils.verify_in_list` now handles empty lists.

srivarra:
@camisowers If either list (or both) are none, it immediately returns `True`.

src/ark/utils/data_utils.py
```diff
@@ -534,16 +534,22 @@ def stitch_images_by_shape(data_dir, stitched_dir, img_sub_folder=None, channels
...
     for chan in channels:
-        image_data = load_utils.load_tiled_img_data(data_dir, fovs, expected_fovs, chan,
-                                                    single_dir=any([segmentation, clustering]),
-                                                    file_ext=file_ext,
-                                                    img_sub_folder=img_sub_folder)
-        stitched_data = data_utils.stitch_images(image_data, num_cols)
-        current_img = stitched_data.loc['stitched_image', :, :, chan].values
-        image_utils.save_image(os.path.join(stitched_dir, chan + '_stitched.' + file_ext),
-                               current_img)
+        for tile in expected_tiles:
```
srivarra, line 541: We can remove the nested loop with [`itertools.product`](https://docs.python.org/3/library/itertools.html#itertools.product).

tests/utils/spatial_lda_utils_test.py
```diff
@@ -51,11 +51,11 @@ def test_check_format_cell_table_args():
         spu.check_format_cell_table_args(valid_df, invalid_markers1, None)
     with pytest.raises(ValueError):
         spu.check_format_cell_table_args(valid_df, invalid_markers2, None)
-    with pytest.raises(ValueError, match=r"List arguments cannot be empty"):
+    with pytest.raises(ValueError, match=r"empty"):
```
srivarra, line 54: `alpineer.misc_utils.verify_in_list` will return `True` if the lists are empty, do you think it would be better to do something else? Here is the PR for it: https://github.com/angelolab/alpineer/pull/25

camisowers:
> @camisowers If either list (or both) are none, it immediately returns `True`.

What was the reasoning there? Just curious because `spatial_lda_utils_test.test_check_format_cell_table_args` specifically tests empty lists to trigger an error.

src/ark/utils/data_utils.py
```diff
@@ -524,7 +524,7 @@ def stitch_images_by_shape(data_dir, stitched_dir, img_sub_folder=None, channels
             dir_name=os.path.join(data_dir, fovs[0], img_sub_folder),
             substrs=EXTENSION_TYPES["IMAGE"])
     else:
-        channel_imgs = io_utils.list_files(data_dir, substrs=fovs[0])
+        channel_imgs = io_utils.list_files(data_dir, substrs=fovs[0]+'_')
```
srivarra, line 527: Are you adding the `'_'` because there would be a conflict between `run_1` and `run_10`? If so, I've  also ran into that issue. Do you think it would be useful to implement a fix where numbers are treated properly?

srivarra (changes requested):
Looks good, just a couple of small comments.

srivarra:
> > @camisowers If either list (or both) are none, it immediately returns `True`.
> 
> What was the reasoning there? 

I think the reasoning had something to do with the mathematics of sets, since that's what we are effectively doing. The empty Set is a member of the empty Set.

srivarra:
Looks like docs aren't building, there is a thread here with everyone experiencing the same issue: https://github.com/urllib3/urllib3/issues/2168. One solution is to cap `urllib3` with `urllib3<2` in the `rtd-requirements.txt`.

alex-l-kong:
> Looks like docs aren't building, there is a thread here with everyone experiencing the same issue: [urllib3/urllib3#2168](https://github.com/urllib3/urllib3/issues/2168). One solution is to cap `urllib3` with `urllib3<2` in the `rtd-requirements.txt`.

Pinning `requests==2.28.0` also worked for me.

tests/utils/spatial_lda_utils_test.py
camisowers: We can discuss at pipeline today. I think it really just depends on what we're using `verify_in_lists` for and if empty lists are considered something we would want to know about.

src/ark/utils/data_utils.py
```diff
@@ -534,16 +534,22 @@ def stitch_images_by_shape(data_dir, stitched_dir, img_sub_folder=None, channels
...
+            stitched_subdir = os.path.join(stitched_dir, prefix)
+            if not os.path.exists(stitched_subdir):
+                os.makedirs(stitched_subdir)
+            image_data = load_utils.load_tiled_img_data(data_dir, fovs, expected_fovs, chan,
+                                                        single_dir=any([segmentation, clustering]),
+                                                        file_ext=file_ext,
+                                                        img_sub_folder=img_sub_folder)
+            stitched_data = data_utils.stitch_images(image_data, num_cols)
+            current_img = stitched_data.loc['stitched_image', :, :, chan].values
+            image_utils.save_image(os.path.join(stitched_subdir, chan + '_stitched.' + file_ext),
```
alex-l-kong, line 554: Minor QOL fix: `file_ext` often refers to the extension plus the leading `'.'` in our repo (ex. `.bin`). We generally use `os.path.splitext` to extract this, noticed you used `.split('.')[1]` which also works.

src/ark/utils/data_utils.py
camisowers: Yeah that's exactly why, but regarding the full FOV name. It's only coming up here because for segment/clustering mask directories there's no fov level folders, so I have to extract the "channel" names (e.g "whole_cell.tiff", "nuclear.tiff", etc.) and it was resulting in "0_whole_cell.tiff" for R1C1 if R1C10 exists.

tldr: I think it's fine for now, just something we have to be mindful of.

tests/utils/spatial_lda_utils_test.py
alex-l-kong: We could handle this by setting a flag to override the default set behavior of empty lists. Something like `fail_on_empty` which defaults to `True`. 

Since I'd imagine empty lists aren't the desired behavior for many users, we could also completely change `verify_in_list` and `verify_same_elements` to fail immediately on discovery of an empty list.

alex-l-kong (commented):
One more comment on top of @srivarra.

tests/utils/spatial_lda_utils_test.py
camisowers: @srivarra  @alex-l-kong For now, I'm going to check within the function for an empty list. We can make long term changes to alpineer later if needed.

srivarra (approved):
Looks good!

alex-l-kong (approved):
👍

ngreenwald (approved):

srivarra:
@ngreenwald Looks like the Merge Queue issue relates to coveralls being able to report back in merge queues. As such I've disabled it for now.

PR #984, opened 2023-05-02
Skip spatial LDA visualization tests due to Numpy error

alex-l-kong:

The spatial LDA visualization tests currently imports `spatial_lda.visualization` and `spatial_lda.processing`. There's a bug at https://github.com/calico/spatial_lda/blob/primary/spatial_lda/online_lda.py#L31 where an incompatible `np.float` call is made. This is outdated with the most up-to-date version of `numpy` and throws an error on import.

We'll have to check in with `calico` to get this resolved, but in the meantime, pin numpy back at 1.23, the last version that doesn't throw an error.

**Remaining issues**

Not sure how long `calico` will take to respond. They don't seem particularly invested. Who was Brian's point of contact on this?


ngreenwald (approved):
They'll never respond, it's fine to just pin for now. Once this version gets deprecated, we'll know if we're still using spatialLDA or not.

PR #993, opened 2023-05-16
Prevent duplicate renamed metaclusters in Pixie remapping

### LINKED ISSUE #972: 2 identicle pixel cluster names causes error when running train_cell_som
**Describe the bug**
When 2 pixel meta clusters are assigned the same name they are considered as 2 separate metaclusters. However in cell clustering notebook, these two pixel metaclusters are considered to be the same. This results in an error when running train_cell_som function 

image

**Expected behavior**
Generate a warning message when assigning a duplicate pixel metacluster name in the GUI or a warning when running the apply_pixel_meta_cluster_remapping after renaming the metaclusters in the GUI. 

@alex-l-kong @cliu72

alex-l-kong:

Closes #972. Duplicate metacluster renamed values causes issues with grouping in later steps, so this should not be allowed.

Verify that the provided renamed metacluster has not already been taken. If it has, then terminate the renaming process. This will cause that particular metacluster name to be truncated for the time being and is noted in the documentation.


start_docker.sh
```diff
@@ -44,7 +44,7 @@ fi
 # find lowest open port available
 PORT=8888
 
-until [[ $(docker container ls | grep 0.0.0.0:$PORT | wc -l) -eq 0 ]]
+until [[ $(netstat -an | grep LISTEN | grep $PORT | wc -l) -eq 0 ]]
```
srivarra, line 47: What does this do?

src/ark/phenotyping/pixel_meta_clustering.py
```diff
@@ -371,12 +371,22 @@ def apply_pixel_meta_cluster_remapping(fovs, channels, base_dir,
...
+        ['pixel_meta_cluster', 'pixel_meta_cluster_rename']
+    ].drop_duplicates()
+    pixel_meta_dups = pixel_meta_pairs[
+        pixel_meta_pairs.duplicated('pixel_meta_cluster_rename', keep=False)
+    ]['pixel_meta_cluster_rename'].unique().tolist()
+    if len(pixel_meta_dups) > 0:
+        raise ValueError(
+            "Duplicate renamed pixel meta cluster values found: %s, "
+            "please re-run remapping GUI to resolve naming conflicts" % str(pixel_meta_dups)
+        )
```
srivarra, line 386: Since you've used this block of code for both `cell` and `pixel` meta clustering, do you think it would be valuable to refactor it into a utility function?

srivarra (changes requested):
Looks good, just a question and a refactoring suggestion.

camisowers (commented):
Nice. Nothing more to add onto @srivarra's comments.

start_docker.sh
alex-l-kong: If people are running local Jupyter servers using `docker container ls` doesn't always find them. `netstat -an | grep LISTEN` does the trick more thoroughly.

src/ark/phenotyping/pixel_meta_clustering.py
alex-l-kong: On second thought, might want to think a bit more about this. We have separate `pixel_cluster_utils` and `cell_cluster_utils`, but no module that includes functionality shared by both pixel and cell.

I'm a bit hesitant to use `cluster_helpers`, since that was intended for just the `Pixie` class definitions. However, since SOM and meta clustering processes are shared for both pixel and cell, we could potentially just plop shared functions there anyways.

Open to discussion.

srivarra (approved):
Looks good!

cliu72 (commented):
Looks good to me. Just a tinnyyy notebook wording change.

templates/2_Pixie_Cluster_Pixels.ipynb
```diff
@@ -1,7 +1,6 @@
 {
```
cliu72, line 1: <p>Can we change this to: "You cannot use the same name for different meta clusters; doing so will cause the next step to fail."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/993/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='42'/>

templates/3_Pixie_Cluster_Cells.ipynb
```diff
@@ -574,7 +574,7 @@
     "### Other features and notes\n",
```
cliu72, line 574: <p>Can we change this to: "You cannot use the same name for different meta clusters; doing so will cause the next step to fail."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/993/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

src/ark/phenotyping/cluster_helpers.py
```diff
@@ -16,6 +16,38 @@
 from alpineer.misc_utils import verify_in_list
 
 
+def verify_unique_meta_clusters(pixie_remapped_data: pd.DataFrame, meta_cluster_type: str):
```
cliu72, line 19: I think it would be helpful to have `meta_cluster_type` specified like Sri has `cluster_type` in another pull request (so it's clear that it should be either "pixel" or "cell"):

```
def save_colored_masks(
        fovs: List[str],
        mask_dir: Union[str, pathlib.Path],
        save_dir: Union[str, pathlib.Path],
        cluster_id_to_name_path: Union[str, pathlib.Path],
        metacluster_colors: Dict,
        cluster_type: Literal["cell", "pixel"],
) 
```

cliu72 (approved):
Looks good!

PR #1015, opened 2023-06-14
Mixing score default thresholds

camisowers:

After testing the mixing score on 100 images and across various populations, we determined that total cell count was more useful to restrict rather than counts for the individual populations. As a result, we need to be more restrictive with the ratio of cells that we allow. The defaults that seem to work best are `ratio_threshold=5` and `cell_count_thresh=200`.

For plotting the ratios in the mixing notebook, we should drop any bigger than 15. Large ratio values just mess with the x axis and makes the plots hard to interpret. 
Image

Image

Change the threshold default values in both the notebook and the code. For the output file, only return one `cell_ratio` value per image, instead of both the pop1/pop2 ratio and pop2/pop1 ratio.

Adjust the notebook to 
1. Check for and automatically generate the specified neighbor matrix, rather than rely on the user to run the correct cell. 
2. Tweak the ratio plots to have x-limit 15 and the mixing histogram plot to always start at 0.

**Remaining issues**

N/A


src/ark/analysis/neighborhood_analysis.py
```diff
@@ -393,8 +393,8 @@ def compute_cell_ratios(neighbors_mat, target_cells, reference_cells, fov_list,
             ref_targ_ratio.append(reference_total / target_total)
 
     # remove nan values for plotting
-    targ_ref_remove_nan = [x for x in targ_ref_ratio if str(x) != 'nan']
-    ref_targ_remove_nan = [x for x in ref_targ_ratio if str(x) != 'nan']
+    targ_ref_remove_nan = [x for x in targ_ref_ratio if x < 15 and str(x) != 'nan']
```
srivarra, line 396:
```python
targ_ref_ratio = np.array(targ_ref_ratio)
targ_ref_remove_nan= targ_ref_ratio[~np.isnan(targ_ref_ratio)]
```

I'd recommend using casting these two lists to numpy arrays and then using the builtin `np.isnan` function to filter out all `np.nan` values.

srivarra (changes requested):
Looking goood, just a `np.nan` filtering suggestion.

srivarra (approved):

ngreenwald (approved):
Nice stuff

PR #1020, opened 2023-06-20
Pixie percentile saving

### LINKED ISSUE #900: Data frame related warning generated in pixie cluster pixels notebook
**Describe the bug**
When creating pixel cluster matrix, after processing 100 fovs the following warning (see image) was generated for every FOV thereafter. 

Screenshot 2023-02-03 at 4 06 02 PM

### LINKED ISSUE #1016: In pixel clustering, if restarting `create_pixel_matrix`, not all FOVs are used to find the 99.9th percentile values
**Describe the bug**
If the `create_pixel_matrix` function in the pixel clustering notebook crashes or is stopped before it finishes for whatever reason, there is code in place to ignore any FOVs that have already been processed (https://github.com/angelolab/ark-analysis/blob/0266b96c689e9118f3b77355fe08df8f9c17b52c/src/ark/phenotyping/pixie_preprocessing.py#LL245-L258). However, this causes a problem with generating the 99.9th percentile values because only the FOVs that were processed will be used to generate the 99.9th percentile values (ignoring all FOVs that have already been processed). The correct behavior is to use all FOVs to calculate the 99.9th percentile values.

**Expected behavior**
Either re-calculate the 99.9th percentile values for FOVs that have already been processed at the start of the function, or save the 99.9th percentile values for each FOV as we go. Open to other suggestions as well.

camisowers:

1. Closes #1016. Implement intermediate saving of per image 99.9th percent values. 

In addition, two small corrections were made to the pixie code.

2. adds a new `channel_percentile_postnorm` argument to `create_pixel_matrix()`, rather than hard coding 0.999.
3. add fix for the edge case where an image has no positive pixels (previously would error out)
(@HPiyadasa ran into this issue when trying to pixels cluster only SMA+ pixels within a mask)
4. Should also close #900.

1. Add image values to quantile_data.csv as they are processed. Also check the FOVs included in the file upon a restart, on the off chance a FOV had a feather file written, but the quantile data failed to be saved. Such images will be re-processed.
2. Set default `channel_percentile_postnorm=0.999`.
3. If there are no pixel cluster labels in an image, simply return an empty array rather than trying to concatenate an empty list (this causes an error).

**Remaining issues**
Long term, we could probably prevent zero positive pixel images from attempting to be processed at all, but that likely requires a more in depth fix and the above solution works fine for now.

### DIFF
--- a/src/ark/phenotyping/pixie_preprocessing.py
+++ b/src/ark/phenotyping/pixie_preprocessing.py
@@ -244,7 +244,7 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
         os.mkdir(os.path.join(base_dir, subset_dir))
 
     # create variable for storing 99.9% values
-    quantile_path = os.path.join(base_dir, pixel_output_dir, "quantile_data.csv")
+    quantile_path = os.path.join(base_dir, data_dir, "quantile_data.csv")
 
     # find all the FOV files in the full data and subsetted directories
     # NOTE: this handles the case where the data file was written, but not the subset file
--- a/src/ark/phenotyping/pixie_preprocessing.py
+++ b/src/ark/phenotyping/pixie_preprocessing.py
@@ -357,3 +357,6 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
     feather.write_dataframe(mean_quant.T,
                             os.path.join(base_dir, norm_vals_name),
                             compression='uncompressed')
+
+    # delete quantile data file
+    os.remove(quantile_path)
--- a/src/ark/phenotyping/pixie_preprocessing.py
+++ b/src/ark/phenotyping/pixie_preprocessing.py
@@ -262,8 +262,9 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
     fovs_list = list(set(fovs).difference(set(fovs_full)))
 
     # check for missing quant data and add to the list of FOVs for processing
-    quant_fov_list = pd.read_csv(quantile_path).columns if os.path.exists(quantile_path) else []
-    quant_missing = list(set(quant_fov_list).difference(set(fovs_full)))
+    quant_fov_list = pd.read_csv(quantile_path, index_col="channel").columns \
+        if os.path.exists(quantile_path) else []
+    quant_missing = list(set(quant_fov_list).difference(set(fovs)))
     fovs_list = list(set(fovs_list).union(set(quant_missing)))
 
     # if there are no FOVs left to preprocess don't run function
@@ -317,15 +318,16 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
 
                     # drop the metadata columns and generate the 99.9% quantile values for the FOV
                     fov_full_pixel_data = pixel_mat_data.drop(columns=cols_to_drop)
-                    quant_dat_fov = fov_full_pixel_data.replace(
-                        0, np.nan
-                    ).quantile(q=channel_percentile_postnorm, axis=0)
+                    quant_dat_fov = fov_full_pixel_data.replace(0, np.nan).quantile(
+                        q=channel_percentile_postnorm, axis=0).rename(fov)
+                    quant_dat_fov.index.name = "channel"
 
                     # read in previously processed fov quantile values or initialize new df
-                    quant_dat_all = pd.read_csv(quantile_path) if os.path.exists(quantile_path) \
-                        else pd.DataFrame()
+                    quant_dat_all = pd.read_csv(quantile_path, index_col="channel") \
+                        if os.path.exists(quantile_path) else pd.DataFrame()
                     # update the file with the newly processed fov quantile value
-                    quant_dat_all[fov] = quant_dat_fov
+                    quant_dat_all = quant_dat_all.merge(quant_dat_fov, how="outer",
+                                                        left_index=True, right_index=True)
                     quant_dat_all.to_csv(quantile_path)
 
                 # update number of fovs processed
@@ -338,13 +340,15 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
             # drop the metadata columns and generate the 99.9% quantile values for the FOV
             fov_full_pixel_data = pixel_mat_data.drop(columns=cols_to_drop)
             quant_dat_fov = fov_full_pixel_data.replace(0, np.nan).quantile(
-                q=channel_percentile_postnorm, axis=0)
+                q=channel_percentile_postnorm, axis=0).rename(fov)
+            quant_dat_fov.index.name = "channel"
 
             # read in previously processed fov quantile values or initialize new df
-            quant_dat_all = pd.read_csv(quantile_path) if os.path.exists(quantile_path) \
-                else pd.DataFrame()
+            quant_dat_all = pd.read_csv(quantile_path, index_col="channel") \
+                if os.path.exists(quantile_path) else pd.DataFrame()
             # update the file with the newly processed fov quantile values
-            quant_dat_all[fov] = quant_dat_fov
+            quant_dat_all = quant_dat_all.merge(quant_dat_fov, how="outer",
+                                                left_index=True, right_index=True)
             quant_dat_all.to_csv(quantile_path)
 
             # update number of fovs processed
@@ -354,8 +358,8 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
             if fovs_processed % 10 == 0 or fovs_processed == len(fovs_list):
                 print("Processed %d fovs" % fovs_processed)
 
-    # get mean 99.9% across all fovs for all markers, check that none are missing
-    quant_dat = pd.read_csv(quantile_path)
+        # get mean 99.9% across all fovs for all markers, check that none are missing
+    quant_dat = pd.read_csv(quantile_path, index_col="channel")
 
     mean_quant = pd.DataFrame(quant_dat.mean(axis=1))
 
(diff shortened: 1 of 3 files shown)


src/ark/phenotyping/pixie_preprocessing.py
```diff
@@ -308,9 +318,17 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
...
                     fov_full_pixel_data = pixel_mat_data.drop(columns=cols_to_drop)
-                    quant_dat[fov] = fov_full_pixel_data.replace(
-                        0, np.nan
-                    ).quantile(q=0.999, axis=0)
+                    quant_dat_fov = fov_full_pixel_data.replace(0, np.nan).quantile(
+                        q=channel_percentile_postnorm, axis=0).rename(fov)
+                    quant_dat_fov.index.name = "channel"
+
+                    # read in previously processed fov quantile values or initialize new df
+                    quant_dat_all = pd.read_csv(quantile_path, index_col="channel") \
```
alex-l-kong, line 326: See comment above, we shouldn't need to keep loading in `quantile_path` on each iteration. Have it pre-loaded before the loop, then incrementally merge in new data and save while keeping it in memory.

src/ark/phenotyping/pixie_preprocessing.py
```diff
@@ -308,9 +318,17 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
...
+                        if os.path.exists(quantile_path) else pd.DataFrame()
+                    # update the file with the newly processed fov quantile value
+                    quant_dat_all = quant_dat_all.merge(quant_dat_fov, how="outer",
+                                                        left_index=True, right_index=True)
+                    quant_dat_all.to_csv(quantile_path)
```
alex-l-kong, line 331: Make sure to set `index=False` to prevent an additional indexing column from appearing.

src/ark/phenotyping/pixie_preprocessing.py
```diff
@@ -321,7 +339,17 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
 
             # drop the metadata columns and generate the 99.9% quantile values for the FOV
             fov_full_pixel_data = pixel_mat_data.drop(columns=cols_to_drop)
-            quant_dat[fov] = fov_full_pixel_data.replace(0, np.nan).quantile(q=0.999, axis=0)
+            quant_dat_fov = fov_full_pixel_data.replace(0, np.nan).quantile(
+                q=channel_percentile_postnorm, axis=0).rename(fov)
+            quant_dat_fov.index.name = "channel"
+
+            # read in previously processed fov quantile values or initialize new df
+            quant_dat_all = pd.read_csv(quantile_path, index_col="channel") \
```
alex-l-kong, line 347: Same comments as above:

- Have `quantile_path` preloaded
- Use the `index=False` param setting on `to_csv`

src/ark/phenotyping/pixie_preprocessing.py
```diff
@@ -330,7 +358,9 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
             if fovs_processed % 10 == 0 or fovs_processed == len(fovs_list):
                 print("Processed %d fovs" % fovs_processed)
 
-    # get mean 99.9% across all fovs for all markers
+        # get mean 99.9% across all fovs for all markers, check that none are missing
```
alex-l-kong, line 361: Alignment

src/ark/phenotyping/pixie_preprocessing.py
```diff
@@ -330,7 +358,9 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
...
+    quant_dat = pd.read_csv(quantile_path, index_col="channel")
+
     mean_quant = pd.DataFrame(quant_dat.mean(axis=1))
```
alex-l-kong, line 364: It's a tossup for me whether or not we should delete `quantile_data.csv` after this function is done, personally I would but up to you.

alex-l-kong (changes requested):
Comments about the loading process

src/ark/phenotyping/pixie_preprocessing.py
```diff
@@ -174,7 +175,8 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
                         subset_dir='pixel_mat_subsetted',
                         norm_vals_name='channel_norm_post_rowsum.feather', is_mibitiff=False,
                         blur_factor=2, subset_proportion=0.1, seed=42,
-                        channel_percentile=0.99, multiprocess=False, batch_size=5):
+                        channel_percentile=0.99, channel_percentile_postnorm=0.999,
```
cliu72, line 178: Can we change `channel_percentile` to `channel_percentile_prenorm` to make it a smidge more clear?

cliu72, line 178: Sorry to be super annoying about this, but could we remove `channel_percentile` for now (and add it back in when we add back in the first normalization step)? Since `channel_percentile` isn't being used now, it's a little confusing to have it still in the function definition.

src/ark/phenotyping/pixie_preprocessing.py
```diff
@@ -218,6 +220,8 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
             The random seed to set for subsetting
         channel_percentile (float):
             Percentile used to normalize channels to same range
+        channel_percentile_postnorm (float):
+            Percentile used for second round of normalization
```
cliu72, line 224: This is very inconsequential, but to make it a tad more clear, can we change this description to say "Percentile used to normalize channels after pixel normalization"?

src/ark/phenotyping/pixie_preprocessing.py
```diff
@@ -240,7 +244,7 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
         os.mkdir(os.path.join(base_dir, subset_dir))
 
     # create variable for storing 99.9% values
-    quant_dat = pd.DataFrame()
+    quantile_path = os.path.join(base_dir, data_dir, "quantile_data.csv")
```
cliu72, line 247: Again pretty inconsequential, but can we change the filename to "channel_norm_post_rowsum_perfov.csv", just so the name is a little more descriptive (and matches the feather file name)?

src/ark/phenotyping/pixie_preprocessing.py
```diff
@@ -257,6 +261,12 @@ def create_pixel_matrix(fovs, channels, base_dir, tiff_dir, seg_dir,
     # NOTE: if an existing FOV is already corrupted, future steps will discard it
     fovs_list = list(set(fovs).difference(set(fovs_full)))
 
+    # check for missing quant data and add to the list of FOVs for processing
+    quant_fov_list = pd.read_csv(quantile_path, index_col="channel").columns \
```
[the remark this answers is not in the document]
cliu72: Agree

src/ark/phenotyping/pixie_preprocessing.py
cliu72: I think we should delete this file as well. I think having this file stick around would just confuse people.

cliu72 (commented):
Agree with Alex's comments, and added just a few inconsequential naming changes.

src/ark/phenotyping/pixie_preprocessing.py
camisowers: I need the index to set to true since that's where the channel names are stored. Rather than included it as a column which would just later need to be removed when taking the column averages.

src/ark/phenotyping/pixie_preprocessing.py
camisowers: It only is there until `pixel_som_clustering.generate_som_avg_files()` is ran 2 cells later in the notebook, then it's deleted. Not quite sure why? Is the pixel_mat_data subdir emptied and the feather files regenerated?? @alex-l-kong @cliu72

cliu72: Hmm I think I'm a little confused - it seems like you're creating a file called `quantile_data.csv` in `data_dir` to store the quantiles for each FOV right? I think we meant that at the end of the `create_pixel_matrix` function after the mean quantiles has been calculated and stored in the feather file, we can delete `quantile_data.csv` since it's no longer needed. Should be independent of the `cluster_pixels` function.

camisowers: Oh okay, I can delete it at the end of the function! I was just curious why it gets deleted on it's own later in the notebook anyways.

cliu72: Ohh I see what you mean. Yeah so the reason it's implemented like that is to deal with when someone restarts the notebook during cluster assignment. So like all the FOV feather files are created in `pixel_mat_dir` during preprocessing, and then during cluster assignment, the feather files that have the cluster assignments are written to `pixel_mat_dir_temp` (because if they were written to `pixel_mat_dir` directly, then it would overwrite the originals we wouldn't be able to tell which FOVs were already clustered without iterating through all the FOVs to check and that would take longer). So we write to the temp directory and then after all the clustering is done, then we delete the original `pixel_mat_dir` and replace it with the temp one. Hopefully that made sense.

I think we can delete quantile_data separately from all that though, just to keep it a bit cleaner.

camisowers:
Consulted with @alex-l-kong and removed a failing test that was resulting from the missing channel normalization code.

src/ark/phenotyping/pixie_preprocessing.py
```diff
@@ -6,6 +6,7 @@
 import numpy as np
 import pandas as pd
 import scipy.ndimage as ndimage
+from natsort import natsorted
```
cliu72, line 9: Just curious - what is this for?

cliu72 (approved):
Looks good to me!

src/ark/phenotyping/pixie_preprocessing.py
camisowers: Oh I was planning on using it to sort the fov names, but I didn't end up needing it. I'll remove it!

alex-l-kong (approved):
Looks good.

ngreenwald (approved):

Issue #857, opened 2022-12-10
Refactor mask generation


ngreenwald:
**Is your feature request related to a problem? Please describe.**
We have a fair amount of duplicate code in these two plotting functions:
https://github.com/angelolab/ark-analysis/blob/main/ark/utils/plot_utils.py#L24

https://github.com/angelolab/ark-analysis/blob/main/ark/utils/plot_utils.py#L91

In addition, the shared code is not modular enough to enable it to be reused for other plotting. It's a fairly common task to have a segmentation mask and want to color the cells given a provided set of labels. We should refactor these two functions as necessary to create a core conserved function

**Describe the solution you'd like**
I wrote something for my own project that does something similar. The proposed solution doesn't need to be exactly this, but similar in spirit. 

def create_cell_overlay(cell_table, seg_folder, fovs, cluster_col, plot_dir, save_names):
    cell_subset = cell_table[cell_table['fov'].isin(fovs)]
    cell_subset['unique_ids'] = pd.factorize(cell_subset[cluster_col])[0] + 1

    # import viridis colormap from mpl
    num_categories = np.max(cell_subset.unique_ids)
    cm_values = cm.get_cmap('Paired', num_categories - 1)

    # get RGB values from cm_values
    rgb_values = cm_values(np.arange(num_categories - 1))

    # combine with all black for background
    rgb_values = np.vstack((np.array([0, 0, 0, 1]), rgb_values))

    new_cmap = colors.ListedColormap(rgb_values)

    for idx, image in enumerate(fovs):
        seg_mask = io.imread(os.path.join(seg_folder, image + '_feature_0.tif'))[0]

        edges = find_boundaries(seg_mask, mode='inner')
        seg_mask = np.where(edges == 0, seg_mask, 0)

        # convert string entries in pandas df to unique integers
        cell_subset_plot = cell_subset[cell_subset['fov'] == image]
        labels_dict = dict(zip(cell_subset_plot['label'], cell_subset_plot['unique_ids']))

        # relabel the array
        relabeled_img_array = data_utils.relabel_segmentation(seg_mask, labels_dict)

        output = new_cmap(relabeled_img_array / np.max(relabeled_img_array))
        io.imsave(os.path.join(plot_dir, save_names[idx]), output)

ngreenwald:
This would actually be quite useful to have as a general function. Some additional useful points:
1. User-provided or default color mask
2. Consistent coloring across multiple images, assigning the same cell type the same color each time
[...]

PR #1021, opened 2023-06-30
Allow FOVs to contain various maximum clusters

srivarra:

Keeps the segmentation consistent across all FOVs in a cohort, even if a FOV does not have a particular phenotype / cluster when compared to others in its cohort.
Adds the fix for cell clustering, neighborhood clustering, and for Mantis ingestion.

Created a new dataclass `CellClusterMaskData` which contains all the cell data, cluster column, segmentation column for a particular cohort (or all FOVs used for the cell table).

**Remaining issues**

Partially addresses #857.


src/ark/utils/plot_utils.py
```diff
@@ -293,17 +312,16 @@ def plot_pixel_cell_cluster_overlay(img_xr, fovs, cluster_id_to_name_path, metac
             cax=cax,
             orientation='vertical'
         )
-        cbar.ax.set_yticklabels(['Empty'] + list(
+        cbar.ax.set_yticklabels(
             mcc.metacluster_id_to_name[f'{cluster_type}_meta_cluster_rename'].values)
```
camisowers, line 316: Is "Empty" now included in `metacluster_id_to_name`?

camisowers (commented):
Looks really nice, just a couple clarifying questions.

src/ark/segmentation/marker_quantification.py
```diff
@@ -426,6 +426,9 @@ def create_marker_count_matrices(segmentation_labels, image_data, nuclear_counts
     arcsinh = pd.DataFrame(data=marker_counts_arcsinh.values[0, :, :],
                            columns=marker_counts_arcsinh.features)
 
+    normalized[settings.CELL_LABEL] = normalized[settings.CELL_LABEL].astype(np.int64)
```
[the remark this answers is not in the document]
srivarra: Makes sense, I'll convert it to `np.int32`.

src/ark/utils/plot_utils.py
srivarra: Yup

alex-l-kong (approved):
Looks good, nothing further to add to Cami's comments.

src/ark/utils/data_utils.py
```diff
@@ -54,119 +57,199 @@ def save_fov_mask(fov, data_dir, mask_data, sub_dir=None, name_suffix=''):
...
+
+        self.data[self.label_column] = self.data[self.label_column].astype(np.int64)
+        self.data[self.cluster_column] = self.data[self.cluster_column].astype(np.int64)
+
+        self.unique_fovs = ns.natsorted(self.data[self.fov_col].unique().tolist())
+
+        self.unassigned_id: int = self.data[self.cluster_column].max() + 1
+
+        self.mapping: DataFrameGroupBy = (
+            pd.concat(
```
ngreenwald, line 106: This is confusing to read, especially without comments. I would think about splitting this up into multiple actions, rather than a single "one liner" that spans multiple lines

src/ark/utils/data_utils.py
```diff
@@ -54,119 +57,199 @@ def save_fov_mask(fov, data_dir, mask_data, sub_dir=None, name_suffix=''):
...
+                            self.cluster_column: np.repeat(
+                                0, repeats=len(self.unique_fovs)
+                            ),
+                        }
+                    ),
+                ]
+            )
+            # sort by FOV first, then by segmentation label
+            # Then group by FOV
+            .sort_values(by=[self.fov_col, self.label_column])
```
ngreenwald, line 125: So in the end, this dataclass stores every FOV, and for each FOV, the mapping between each individual segmentation ID and the corresponding cluster ID, is that correct?

Have you tried running this function, and using this dataclass, on a very large cell table? My concern is that doing all the mapping ahead of time and storing the results, rather than in real time when needed, is going to cause this be slow

ngreenwald (changes requested):
Looks good, just some questions about implementation

src/ark/utils/data_utils.py
srivarra: Optimized this and ran ran some performance metrics with [scalene](https://github.com/plasma-umass/scalene):
1. Switched from a `dataclass` to a regular `class`, therefore we will only be extracting the `fov_column` `label_column` (`segmentation_label`), and the `cluster_column` from the cell table, whereas the`dataclass` would contain both the original DataFrame and the 
2. Benchmarked the new process with a Cell Table consisting of 10 Million cells, 2000 FOVs, and 40 MetaClusters totaling 7.9Gb. Creating the mapping ahead of time requires additional 1.4 GB of memory usage for the whole table. This step took less than 3 seconds.

ngreenwald (approved):
Looks good, once @cliu72 is happy can merge it in.

src/ark/utils/data_utils.py
```diff
@@ -54,119 +56,193 @@ def save_fov_mask(fov, data_dir, mask_data, sub_dir=None, name_suffix=''):
...
+        self.fov_column: str = fov_col
+        self.label_column: str = label_col
+        self.cluster_column: str = cluster_col
+
+        # Extract only the necessary columns: fov ID, segmentation label, cluster label
+        mapping_data: pd.DataFrame = data[[self.fov_column,
+                                           self.label_column, self.cluster_column]].copy()
+
+        mapping_data[self.cluster_column] = mapping_data[self.cluster_column].astype(np.int64)
+        mapping_data[self.label_column] = mapping_data[self.label_column].astype(np.int64)
```
cliu72, line 97: I guess not super important, but related to Cami's comment, np.int32 would probably be okay here?

src/ark/utils/data_utils.py
```diff
@@ -54,119 +56,193 @@ def save_fov_mask(fov, data_dir, mask_data, sub_dir=None, name_suffix=''):
...
+
+        self.unique_fovs: List[str] = ns.natsorted(mapping_data[self.fov_column].unique().tolist())
+
+        self.unassigned_id: np.int64 = mapping_data[self.cluster_column].max() + 1
```
cliu72, line 101: Again, I don't think this needs to be np.int64?

src/ark/phenotyping/post_cluster_utils.py
```diff
@@ -97,6 +98,13 @@ def create_mantis_project(cell_table, fovs, seg_dir, pop_col,
     # generate unique numeric value for each population
     small_table['pop_vals'] = pd.factorize(small_table[pop_col].tolist())[0] + 1
 
+    ccmd_pop = data_utils.CellClusterMaskData(
+        data=small_table,
+        fov_col=settings.FOV_ID,
+        label_col=settings.CELL_LABEL,
+        cluster_col="pop_vals",
+    )
+
```
cliu72, line 107: Can we have `fov_col` and `label_col` be arguments to the `create_mantis_project` function and have the default values be `settings.FOV_ID` and `settings.CELL_LABEL`? Not a huge change but I found before when testing that it was helpful to be able to change these values directly from the notebook (for example, if I'm using a different cell table or something to not have to change the settings file)

src/ark/utils/data_utils.py
```diff
@@ -54,119 +56,193 @@ def save_fov_mask(fov, data_dir, mask_data, sub_dir=None, name_suffix=''):
...
-    # NOTE: label_map fov validation happens in loading function
-    misc_utils.verify_in_list(fov_name=[fov], all_data_fovs=all_data[fov_col].unique())
+    misc_utils.verify_in_list(
+        fov_name=[fov],
+        all_data_fovs=ccmd.unique_fovs)
 
-    # subset all_data on the FOV
-    df = all_data[all_data[fov_col] == fov]
+    # condense extraneous axes
+    labeled_image: np.ndarray = label_map.squeeze().values.astype(np.int64)
```
cliu72, line 167: Again, I don't think this needs to be np.int64?

src/ark/utils/data_utils.py
```diff
@@ -54,119 +56,193 @@ def save_fov_mask(fov, data_dir, mask_data, sub_dir=None, name_suffix=''):
...
+    labeled_image: np.ndarray,
+) -> np.ndarray:
+    """
+    Relabels a labled segmentation image according to the provided cluster labels.
+
+    Args:
+        mapping (nb.typed.typeddict): A Numba typed dictionary mapping segmentation labels to
+            cluster labels.
+        unassigned_id (np.int64): The cluster labels for each cell, where the index
+            is the segmentation label.
```
cliu72, line 201: I don't think this is the right description for `unassigned_id`?

src/ark/utils/data_utils.py
```diff
@@ -220,21 +294,27 @@ def generate_and_save_cell_cluster_masks(fovs: List[str],
             Specify what to append at the end of every cell mask. Defaults to `''`.
     """
 
+    ccmd = CellClusterMaskData(
+        data=cell_data,
+        fov_col=settings.FOV_ID,
```
cliu72, line 299: Same comment as above, can we have `fov_col` be an argument to `generate_and_save_cell_cluster_masks` and make the default `settings.FOV_ID`?

src/ark/utils/data_utils.py
```diff
@@ -220,21 +294,27 @@ def generate_and_save_cell_cluster_masks(fovs: List[str],
...
+        label_col="segmentation_label",
```
cliu72, line 300: Can we also make `label_col` an argument to `generate_and_save_cell_cluster_masks`?

src/ark/utils/data_utils.py
```diff
@@ -400,6 +480,11 @@ def generate_and_save_neighborhood_cluster_masks(fovs: List[str],
             Specify what to append at the end of every pixel mask. Defaults to `''`.
     """
 
+    ccmd = CellClusterMaskData(data=neighborhood_data,
+                               fov_col=settings.FOV_ID,
+                               label_col=settings.CELL_LABEL,
+                               cluster_col=settings.KMEANS_CLUSTER)
+
```
cliu72, line 487: Same as above - can we make all of these that have `settings` as function arguments and make these as the defaults?

src/ark/utils/plot_utils.py
```diff
@@ -64,29 +74,48 @@ def _metacluster_cmap_generator(self) -> None:
...
+        # Extract unique pairs of (metacluster-ID,  name)
+        # Set the unassigned cluster ID to be the max ID + 1
+        # Set 0 as the No Cluster value
+        metacluster_id_to_name: pd.DataFrame = pd.concat(
+            [
+                metacluster_id_to_name.drop_duplicates(),
+                pd.DataFrame(
+                    data={
+                        f"{self.cluster_type}_meta_cluster": [unassigned_id, 0],
+                        f"{self.cluster_type}_meta_cluster_rename": ["Unassigned", "No Cluster"]
```
cliu72, line 105: Can we change "No Cluster" to "Empty"?

cliu72 (changes requested):
Just a few small comments

src/ark/settings.py
```diff
@@ -4,6 +4,7 @@
 # default cell table column names
 CELL_SIZE = 'cell_size'  # cell size (number of pixels in the cell)
 CELL_LABEL = 'label'  # cell label number (regionprops)
+CELL_SEGMENTATION_LABEL = "segmentation_label"  # cell label number (segmentation)
```
cliu72, line 7: Sorry if you're not done making changes yet - but why is there both `CELL_LABEL` and `CELL_SEGMENTATION_LABEL` in the settings file? I assume those are the same thing?

srivarra: @cliu72 

Looks like `"label"` is converted to `"segmentation_label"` in Pixie https://github.com/angelolab/ark-analysis/blob/44b46c44fd9e4e632fc9bdb4bc9a2f491b33df38/src/ark/phenotyping/cell_cluster_utils.py#L107-L109


It's used throughout pixie, here is an example. Not really sure why we use both.
https://github.com/angelolab/ark-analysis/blob/44b46c44fd9e4e632fc9bdb4bc9a2f491b33df38/src/ark/phenotyping/weighted_channel_comp.py#L127-L129
@alex-l-kong could probably give some more insight.

cliu72: Ah I see. I think it'd be better to make it consistent. @alex-l-kong in the comments you say it makes for a simpler merge, but why is it necessary? I don't think there is anything else in the table with the "label" column name

If it's going to be too annoying to change `segmentation_label` to `label`, I think we should remove `CELL_SEGMENTATION_LABEL` from the settings file and just manually specify `label_col = "segmentation_label"` instead (wherever necessary). I think having both `CELL_LABEL` and `CELL_SEGMENTATION_LABEL` in the settings file is confusing.

alex-l-kong: @cliu72 yeah agree that this `'label'` and `'segmentation_label'` discrepancy is annoying. The reason it's done this way is `'label'` is the name of seg label column in the cell table, but `'segmentation_label'` is the name of the equivalent column in the pixel cluster data `.feather` files.

I'd say change the `segmentation_label` to `label` (or vice versa) is best practice, the only annoying thing would be to update the `.merge` logic that happens inside the `c2pc` data function. If you're not comfortable with that @srivarra then I can contribute that change to this PR.

Alternatively, to expedite the review on this PR, we could implement Candace's hotfix, then open a separate issue for standardizing the segmentation label column names.

src/ark/phenotyping/cell_cluster_utils.py
```diff
@@ -209,9 +209,9 @@ def add_consensus_labels_cell_table(base_dir, cell_table_path, cell_som_input_da
     # read in the data, ensure sorted by FOV column just in case
     cell_table = pd.read_csv(cell_table_path)
 
-    # for a simpler merge, rename segmentation_label to label in consensus_data
+    # for a simpler merge, rename label to label in consensus_data
     cell_som_results = cell_som_input_data.rename(
-        {'segmentation_label': 'label'}, axis=1
+        {'label': 'label'}, axis=1
```
cliu72, line 214: Is this necessary? Seems like you're renaming something to itself?

cliu72 (commented):
Thanks Sri! Looks good to me, just one tiny comment and I think we're good to merge.

src/ark/phenotyping/cell_cluster_utils.py
srivarra: spittin facts, I'll fix that.

PR #1023, opened 2023-07-11
Deepcell upload loop cleaning

### LINKED ISSUE #628: Retry missing `fov_group` zip files in case of DeepCell response error

The currently-open PR to address retrying in case of a bad DeepCell response assumes that this operation can be completed in `n` tries, printing an error if it doesn't. In the long term, what we want happening is to identify the `fov_groups` that don't have an associated `.zip` file, then re-run the call to Deepcell on those.

This requires a few steps to complete:

1. Standardized naming conventions for `.zip` files. Currently, it's hard to associate which FOVs were placed in which `.zip` files without unzipping them. A more indicative name would be the easier solution.
2. Enclose the call to the Deepcell runner function in a loop, with a check to see if all the `.zip` files have been created (number will depend on the batch size). If not, then identify which `fov_groups` need to be resent, and remain in the loop. Otherwise, break.

camisowers:

Closes #628. Changes how we attempt to re-upload failed fov batches to the deepcell server.
Also refactors `create_deepcell_output()` with cleanlier helper functions. The code will now also check for previously processed batches, to avoid re-zipping and re-uploading already segmented batches to deepcell. 

Currently, we attempt to re-upload a failed fov batches until success or 5 retry attempts are reached. A cleaner, more long term fix is to attempt to process each batch once, check if any batch .zip files are missing, and re-run the failed batches.
The re-running continues until either a successful output or the function times out. 

* Re-run friendly additions. For each batch of fovs:
1. Check if `fovs_batch_1.zip` exists, if not generate the input zip file
2. Check if  `deepcell_output_fovs_batch_1.zip` exists, if not upload input file to deepcell and extract segmentation tiffs

**Remaining issues**

- [x] Set a time limit for reprocessing attempts.
- [x] Tweak the code so it does not re-upload processed fov batches when the cell is re-run


src/ark/utils/deepcell_service_utils.py
```diff
@@ -99,85 +176,41 @@ def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
...
+        input_zip_path = zip_input_files(deepcell_input_dir, fov_group, batch_num)
 
-        # extract the .tif output
-        print("Extracting tif files from DeepCell response.")
-        zip_names = io_utils.list_files(deepcell_output_dir, substrs=[".zip"])
+        # add timeout limit
+        batch_filename = Path(input_zip_path).name
+        output_zip_path = os.path.join(deepcell_output_dir, f"deepcell_response_" + batch_filename)
+        while not os.path.exists(output_zip_path):
+            # upload to deepcell
```
camisowers, line 187: I need to add some kind of timeout here to avoid an endless loop if the deepcell upload continuously fails. I haven't run segmentation on enough data to know a reasonable time limit. Any suggestions? (Keep in mind a single successful iteration will consist of uploading the files to deepcell, running the prediction script, and extracting the tiffs from the deepcell output zip file.) @alex-l-kong @srivarra

alex-l-kong: Because cohorts may be very large, we don't want a single batch of 5 FOVs to be holding up the line. Some leeway is fine just in case the network is slow, but since we'll be including the FOV names in these batches, it should not be difficult for the user to figure out which FOVs failed, if any, and which ones to re-run.

From my experience a FOV batch should not take exceptionally long to process. I'd say no more than 2-3 mins, and depending how long 2048x2048 images take, I might even drop down to 1-1.5 mins.

src/ark/utils/deepcell_service_utils.py
```diff
@@ -99,85 +176,41 @@ def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
...
-        print('Uploading files to DeepCell server.')
-        status = run_deepcell_direct(
-            zip_path, deepcell_output_dir, host, job_type, scale, timeout
-        )
-
-        # ensure execution is halted if run_deepcell_direct returned non-zero exit code
-        if status != 0:
-            print("The following FOVs could not be processed: %s" % ','.join(fov_group))
-            return
+    for batch_num, fov_group in enumerate(fov_groups):
```
srivarra, line 179: In `zip_input_files` and `extract_deep_cell_response` we add 1 for the `batch_num`. We can start the enumeration at `1` with `enumerate(Iterable, start=1)` and remove that op to make it easier to read overall.

srivarra (changes requested):
Jus a couple of small things. 

For the `Retry` call at line 245 `num_retries` is not defined.

src/ark/utils/deepcell_service_utils.py
camisowers: oh this is v helpful thanks!

camisowers:
> Jus a couple of small things.
> 
> For the `Retry` call at line 245 `num_retries` is not defined.

This PR was actually meant to scrap the retry upload a certain amount of times and instead implement recall `run_deepcell_direct()` in a loop until success or timeout. Looks like I missed some code that needed to be removed.

camisowers:
> Jus a couple of small things.
> 
> For the `Retry` call at line 245 `num_retries` is not defined.

This PR was actually meant to scrap the retry a certain amount of times logic and instead implement recall run_deepcell_direct() in a loop until success or timeout. Looks like I missed some code that needed to be removed.

src/ark/utils/deepcell_service_utils.py
```diff
@@ -18,10 +17,91 @@
...
+
+    # create zip files, skip any existing
+    if not os.path.exists(zip_path):
+        print('Zipping preprocessed tiff files.')
+        with ZipFile(zip_path, 'w', compression=ZIP_DEFLATED) as zipObj:
+            for fov in fov_group:
+                # file has .tiff extension
+                basename = fov + '.tiff'
+                filename = os.path.join(deepcell_input_dir, basename)
+                zipObj.write(filename, basename)
```
alex-l-kong, line 42: I'd leave the option to overwrite `.zip` files here. Or at the very least, warn the user that since the `.zip` file exists, it won't be written to.

src/ark/utils/deepcell_service_utils.py
```diff
@@ -99,85 +176,41 @@ def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
...
+    deepcell_outputs = io_utils.remove_file_extensions(
+        io_utils.list_files(deepcell_output_dir, substrs=[".zip"]))
+    misc_utils.verify_same_elements(
+        output_files=deepcell_outputs,
+        expected_files=[f"deepcell_response_fovs_batch_{i + 1}" for i in range(len(fov_groups))])
 
 
 def run_deepcell_direct(input_dir, output_dir, host='https://deepcell.org',
-                        job_type='mesmer', scale=1.0, timeout=3600, num_retries=5):
+                        job_type='mesmer', scale=1.0, timeout=3600):
```
alex-l-kong, line 213: I'd say let's default the timeout to something smaller, 1 hour is a long time to wait for a single batch to upload. 90 or 120 is more reasonable.

alex-l-kong (changes requested):
A few additional comments.

src/ark/utils/deepcell_service_utils.py
```diff
@@ -222,34 +253,22 @@ def run_deepcell_direct(input_dir, output_dir, host='https://deepcell.org',
     http.mount('https://', adapter)
     http.mount('http://', adapter)
```
camisowers, line 254: @alex-l-kong Do you think I should scrap the retry_strategy stuff and use requests.post() like before? Or would just setting num_retries=1 would be appropriate and keep the upload method the same.

alex-l-kong: I think we should just convert to `requests.post()` for consistency purposes.

src/ark/utils/deepcell_service_utils.py
```diff
@@ -99,85 +175,49 @@ def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
...
+            # pass the zip file to deepcell.org
+            status = run_deepcell_direct(
+                input_zip_path, deepcell_output_dir, host, job_type, scale, timeout
+            )
 
-        zip_write(zip_path)
+            # successful deepcell response
+            if status == 0:
+                # extract segmentation masks from deepcell output
+                extract_deepcell_response(deepcell_output_dir, fov_group, batch_num, wc_suffix,
```
alex-l-kong, line 202: Let's `break` out of this `while` loop if `extract_deepcell_response` runs. Otherwise, `total_time` will still get updated, which could also trigger the `exceeded time limit` print statement in certain edge cases.

src/ark/utils/deepcell_service_utils.py
```diff
@@ -99,85 +175,49 @@ def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
...
-                    warnings.warn(f'Deep Cell nuclear output file was not found for {fov}.')
-
-    # make calls in parallel
-    if parallel:
-        with ThreadPoolExecutor() as executor:
-            executor.map(_zip_run_extract, fov_groups, range(len(fov_groups)))
-            executor.shutdown(wait=True)
-    else:
-        list(map(_zip_run_extract, fov_groups, range(len(fov_groups))))
+    if unprocessed_fovs:
```
alex-l-kong, line 213: In the case that a batch fails the first time but passes the next, would this not still print the warning?

There should be a check under `if status == 0` to account for this:

```
if status == 0:
    # extract segmentation masks from deepcell output
    extract_deepcell_response(deepcell_output_dir, fov_group, batch_num, wc_suffix, nuc_suffix)
    
    # since the batch_num is now processed, it should be removed from the dict
    if batch_num in unprocessed_fovs:
        unprocessed_fovs.pop(batch_num)

    # addresses the previous comment
    break

**Review by alex-l-kong — CHANGES_REQUESTED** [review]
Couple of control-flow comments

File: src/ark/utils/deepcell_service_utils.py (hunk 7 above)
Comment (camisowers) [reply src/ark/utils/deepcell_service_utils.py:213]: Kinda hard to see with the code change structure, but I moved the if statement out of the loop so it only add the batch to `unprocessed_fovs` after the loop concludes (either bc of a successful output zip or timeout). Easier to see it here: https://github.com/angelolab/ark-analysis/blob/f40a37d11eb46453fdc753c2a8c8383ee279b76e/src/ark/utils/deepcell_service_utils.py#L207

**Review by srivarra — APPROVED** [review]
Looks super clean!

File: src/ark/utils/deepcell_service_utils.py [hunk 8]
```diff
@@ -99,85 +175,50 @@ def create_deepcell_output(deepcell_input_dir, deepcell_output_dir, fovs=None,
...
+                input_zip_path, deepcell_output_dir, host, job_type, scale, timeout
+            )
 
-        zip_write(zip_path)
+            # successful deepcell response
+            if status == 0:
+                # extract segmentation masks from deepcell output
+                extract_deepcell_response(deepcell_output_dir, fov_group, batch_num, wc_suffix,
+                                          nuc_suffix)
+                continue
```
alex-l-kong, line 204: `continue` works, since `output_zip_path` is created by this call. Is there a reason `break` doesn't work? I think that would be a bit easier to understand if the functionality is the same.

alex-l-kong (approved):
Looks good, just one side question.

src/ark/utils/deepcell_service_utils.py
camisowers: Oops `break` works fine. I thought it might mess with the nested loop but seems alright.

src/ark/utils/deepcell_service_utils.py
```diff
@@ -11,17 +10,97 @@
...
+            it will be renamed to this arg.
+        nuc_suffix (str):
+            Suffix for nuclear DeepCell output filename. e.g. for fovX, DeepCell output
+            should be `<fovX>+suffix.tif`.
+            Nuclear DeepCell files by default get suffixed with `'feature_1'`,
+            it will be renamed to this arg.
+    """
+
+    # extract the .tif output
+    print("Extracting tif files from DeepCell response.")
```
ngreenwald, line 69: Seems like there's a lot of print statements that will trigger on every batch, and they don't have the batch name in them. Did you run this notebook with 3-4 different batches? Is it just a string of the same print statements over and over again?

ngreenwald (changes requested):

src/ark/utils/deepcell_service_utils.py
camisowers: Yeah each batch gets a zip, upload, and extract print statement (plus a progress bar). I just kept it pretty consistent with the prior notebook output, but we can change it if you think it's a bit excessive. Or we could add batch numbers to the statements.

Previous output:
Screenshot 2023-07-25 at 10 22 33 AM

New output:
Screenshot 2023-07-25 at 10 26 36 AM

ngreenwald: I think it would be better to keep the first one that says # of batches, modify the segmentation_progress one so it says progress for batch x, and remove the rest.

ngreenwald (approved):

PR #1025, opened 2023-07-17
Maximize Available GitHub Storage Space

srivarra:

Allows the Ubuntu Test Runner to not run out of storage space.

Adds more `shutil.rmfree` function calls in the example dataset tests and adjusts those in the notebook tests. Adjusts the `Test_Cell_Neighbors` notebook to make use of the cached dataset on GHA. Added a decorator which can be used throughout the example dataset notebook tests to get info about storage space on the runner. For example:

```python
class Test_1_Segment_Image_Data:
    """
    Tests Notebook 1 - Segment Image Data for completion.
    NOTE: When modifying the tests, make sure the test are in the
    same order as the tagged cells in the notebook.
    """
	.
	.
	.
	@notebooks_test_utils.get_storage
    def test_ex_data_download(self):
        notebooks_test_utils._ex_dataset_download(dataset=self.dataset, save_dir=self.base_dir,
                                                  cache_dir=self.cache_dir)

	.
	.
	.

```

**Remaining issues**

None.


tests/utils/notebooks_test_utils.py
```diff
@@ -82,3 +83,28 @@ def _ex_dataset_download(dataset: str, save_dir: str, cache_dir: Union[str, None
     ex_dataset.download_example_dataset()
 
     ex_dataset.move_example_dataset(move_dir=save_dir)
+
+
+def get_storage(method):
```
camisowers, line 88: Where does this get used? Or is it just for future use if something comes up again?

camisowers (approved):
Looks good to me! Just a clarifying question about `get_storage()`.

tests/utils/notebooks_test_utils.py
srivarra: It isn't used anywhere right now, but I did use it to investigate how storage space changes over time through the example dataset tests and the notebook tests.

It should only be used for that purpose, and we don't need to decorate any function with it currently.

alex-l-kong (approved):
Looks good

PR #1050, opened 2023-08-28
Include `img_sub_folder` as parameter in Mantis calls

### LINKED ISSUE #1045: Add `img_sub_folder` to `create_mantis_dir` in Pixie

**Describe the bug**
In the Pixie notebooks, `img_sub_folder` is not in the function call for `create_mantis_dir`, so the single-channel TIFs are not copied over into the mantis directory (when there is an image subfolder).

**Expected behavior**
In both the pixel and cell clustering notebooks, add `img_sub_folder` to the function call for `create_mantis_dir` (it's already an input parameter so no need to change the function code).

The cell clustering notebook currently doesn't have the `img_sub_folder` variable - easiest solution is probably to save `img_sub_folder` to the json at the end of the pixel clustering notebook. Related to this, it looks like the user specifies `tiff_dir` again at the beginning of the cell clustering notebook although they already do so in the pixel clustering notebook - we can probably save `tiff_dir` to the json as well so users don't have to specify it twice.

alex-l-kong:

Closes #1045. Without this argument passed to `create_mantis_dir`, there isn't a way to copy the single-channel TIF files over.

Added this as a parameter to the function. Additionally, updated `cell_clustering_params.json` to take in both `tiff_dir` and `img_sub_folder`. The former is purely QOL and the latter is to support the `img_sub_folder` param for the Mantis function in the cell clustering notebook.

### DIFF
--- a/.github/scripts/get_example_dataset.py
+++ b/.github/scripts/get_example_dataset.py
@@ -23,6 +23,7 @@ def load_dataset(cache_dir: pathlib.Path, name: str):
         cache_dir=cache_dir,
         name=name,
         use_auth_token=False,
+        revision="c38309e19d7eaae9f67e1cf8e08a3fd2b8a6f299"
     )
 
 # Make the cache directory if it doesn't exist.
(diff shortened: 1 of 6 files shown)


cliu72 (approved):
Looks good!

PR #1058, opened 2023-09-06
Make sure overwrite functionality for pixel clustering propagates into helper functions

alex-l-kong:

The overwrite flag wasn't correctly dropping the `"pixel_som_cluster"` column, which will cause problems during assignment since it will attempt to use that column as an actual feature when mapping onto the SOM.

Pass the `overwrite` flag into `run_pixel_som_assignment`.

### DIFF
--- a/.github/scripts/get_example_dataset.py
+++ b/.github/scripts/get_example_dataset.py
@@ -23,7 +23,7 @@ def load_dataset(cache_dir: pathlib.Path, name: str):
         cache_dir=cache_dir,
         name=name,
         use_auth_token=False,
-        revision="c38309e19d7eaae9f67e1cf8e08a3fd2b8a6f299"
+        revision="main"
     )
 
 # Make the cache directory if it doesn't exist.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 dependencies = [
-    "alpineer>=0.1.8",
+    "alpineer==0.1.10",
     "Cython>=0.29,<1",
     "datasets>=2.6,<3.0",
     "dill>=0.3.5,<0.4",
--- a/src/ark/phenotyping/cluster_helpers.py
+++ b/src/ark/phenotyping/cluster_helpers.py
@@ -257,19 +257,22 @@ class PixelSOMCluster(PixieSOMCluster):
 
         super().train_som(self.train_data[self.columns])
 
-    def assign_som_clusters(self, external_data: pd.DataFrame) -> pd.DataFrame:
+    def assign_som_clusters(self, external_data: pd.DataFrame, normalize_data: bool = True) -> pd.DataFrame:
         """Assigns SOM clusters using `weights` to a dataset
 
         Args:
             external_data (pandas.DataFrame):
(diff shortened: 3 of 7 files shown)


cliu72 (approved):
Looks good.

Out of curiosity though, where was this causing an error? From a quick glance at the code, I couldn't figure out where the error would come up. 

In `assign_som_clusters` (https://github.com/angelolab/ark-analysis/blob/main/src/ark/phenotyping/cluster_helpers.py#L260-L278), it looks like only the column names that were in the normalization/weights file would be used. Like the functions `normalize_data` only uses columns from the normalization values, and `generate_som_clusters` only uses columns from the weights. In `generate_som_cluseters`:
```
            cluster_labels.append(map_data_to_nodes(
                self.weights.values.astype(np.float64),
                external_data.loc[
                    i:min(i + 99, external_data.shape[0]), **weights_cols**
                ].values.astype(np.float64)
```

Because of the `weight_cols`, it seems to me that even if the 'pixel_som_cluster' column was still in the data file, it would be ignored.

alex-l-kong:
> Looks good.
> 
> Out of curiosity though, where was this causing an error? From a quick glance at the code, I couldn't figure out where the error would come up.
> 
> In `assign_som_clusters` (https://github.com/angelolab/ark-analysis/blob/main/src/ark/phenotyping/cluster_helpers.py#L260-L278), it looks like only the column names that were in the normalization/weights file would be used. Like the functions `normalize_data` only uses columns from the normalization values, and `generate_som_clusters` only uses columns from the weights. In `generate_som_cluseters`:
> 
> ```
>             cluster_labels.append(map_data_to_nodes(
>                 self.weights.values.astype(np.float64),
>                 external_data.loc[
>                     i:min(i + 99, external_data.shape[0]), weights_cols
>                 ].values.astype(np.float64)
> ```
> 
> Because of the `weight_cols`, it seems to me that even if the 'pixel_som_cluster' column was still in the data file, it would be ignored.

You're right that leaving in the SOM cluster column shouldn't inherently cause problems. Still not not entirely sure why this was an issue with Patricia's run in particular, it's nice to have as an extra check.

After taking another look at the process it seems there's another problem: when the user needs to overwrite the data in `pixel_mat_data`, it shouldn't go through an additional round of normalization each time, which is what's actually happening at the moment. This makes the values extremely small. We'll need to propagate this further into the actual `PixelSOMCluster` and `PixieSOMCluster` classes so we don't renormalize if the `overwrite` flag is set to `True`.

cliu72 (changes requested):
Good catch! Perhaps I'm losing it, but I think that the logic is a bit off.

If `overwrite=True`, then we do not need to re-normalize, so `normalize_data` in `assign_som_clusters` should be False right? Right now, since we're passing `normalize_data=overwrite`, `normalize_data` would be True.

alex-l-kong:
> Good catch! Perhaps I'm losing it, but I think that the logic is a bit off.
> 
> If `overwrite=True`, then we do not need to re-normalize, so `normalize_data` in `assign_som_clusters` should be False right? Right now, since we're passing `normalize_data=overwrite`, `normalize_data` would be True.

Good catch, got a bit lost in the horrors of control flow logic.

cliu72 (approved):
Looks good to me!

PR #1060, opened 2023-09-11
Pin to scikit-image to v0.19.3

### LINKED ISSUE #1055: Bad filtering in fiber segmentation

**Describe the bug**
The current fiber segmentation notebook gives worse outputs than previously, although nothing has been changed in our code. 
Turns out by bumping scikit-image >=0.20 in ark, the frangi filtering step is now pretty different because of the changes made in this PR (scikit-image/scikit-image/pull/6446).

previous vs current ouput
Screenshot 2023-09-05 at 11 32 39 AM

**Expected behavior**
It should look better.

We'll probably need to explore a bit to see what's the best decision moving forward, but here's a few options to consider first: 
1. Pin scikit-image to 0.19.3 (a quick fix but probably not the best long-term plan)
2. See if we can get better results with the new version of `frangi()` by tweaking some parameters
3. Find a new filtering function that works better

**To Reproduce**
Run the fiber segmentation on the example data.

camisowers:

Closes #1055. Improves fiber segmentation results by pinning scikit-image to an older version.

Update pyproject.toml.

**Remaining issues**

Short term solution. Should be looked into for a more permanent fix in the future.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
     "pillow>=9,<10",
     "pyFlowSOM>=0.1.15",
     "requests>=2.20,<3",
-    "scikit-image>=0.21",
+    "scikit-image==0.19.3",
     "scikit-learn>=1.1,<2",
     "scipy>=1.7,<2",
     "seaborn>=0.12,<1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
     "pillow>=9,<10",
     "pyFlowSOM>=0.1.15",
     "requests>=2.20,<3",
-    "scikit-image==0.19.3",
+    "scikit-image<=0.19.3",
     "scikit-learn>=1.1,<2",
     "scipy>=1.7,<2",
     "seaborn>=0.12,<1",


alex-l-kong (commented):
I looked at the `frangi` implementation in `skimage.filters.ridges` and it looks like one of the key differences is in the handling of the `gamma` parameter. In 0.19.3, it's automatically set to 15, whereas in >=0.20.0, it's dynamically computed based on the eigenvalues. Maybe try manually setting the `gamma` param to 15 and see how that turns out?

camisowers:
> I looked at the `frangi` implementation in `skimage.filters.ridges` and it looks like one of the key differences is in the handling of the `gamma` parameter. In 0.19.3, it's automatically set to 15, whereas in >=0.20.0, it's dynamically computed based on the eigenvalues by default. Maybe as a step 0 try manually setting the `gamma` param to 15 and see how that turns out?

Sorry I mentioned this in the issue, but not the PR: I was able to get something comparable to the previous implementation by tweaking a few parameters (gamma for one, and also the fiber widths). It's much better than the current version but still not as good as using scikit-learn v0.19.3. I can further look into adjusting the code with v0.21 at the end of the month, but in the interim we should pin it.

srivarra (approved):

alex-l-kong (approved):

ngreenwald (changes requested):
Does it need to be exactly this version, or just less than or equal?

camisowers:
> Does it need to be exactly this version, or just less than or equal?

Less than or equal should be fine.

ngreenwald (approved):

PR #1061, opened 2023-09-11
Single mantis directory

### LINKED ISSUE #669: Have just one mantis directory
Right now, mantis directories are created at the end of the notebooks for pixel clustering, cell clustering, and neighborhood analysis. When generating a mantis directory, all the single-channel TIFs are copied, so right now, the data is being copied in all 3 of these places, which is inefficient.

Have just one mantis directory (probably in the base directory), that will hold single-channel TIFs, pixel overlay, cell overlay, and neighborhood overlay. In Mantis, the user can choose which of these overlays they want to look at.

Also have an option for the user to add a prefix (or suffix) to the pixel/cell/neighborhood mask overlays, so that they can compare across different runs. I think it is a good idea to have as default, no prefix/suffix and the mask overlays are overwritten if one already exists in the directory. If the user chooses, they can add a prefix/suffix to the files (such as "run1_population_pixel_mask.tif", "run2_population_pixel_mask.tif", etc.), so these are written as separate files. The user can then choose which of these to view in Mantis for easy comparison.

### LINKED ISSUE #1059: Single mantis directory

### Instructions

**Relevant background**

Addresses #669. 
- Currently, the pixie notebooks create separate Mantis directories for 2_pixel, 3_cell clustering, 4_post clustering, and generic clustering. This entails copying over all of the channel images and the segmentation mask for each FOV; it would be better to have a single mantis directory for sequential notebooks, rather than having duplicates of the data.

- For the prefix capable that I spoke with @cliu72 about, we can add a `mask_prefix` arg that defaults to None, so that any advanced users generating multiple pixel/cell masks can load those into the mantis directory. If the masks are named prefix_fov1_pixel_mask.tiff you can just specify the prefix and it will copy those over.

- Also add "mantis_" to all the copied masks, to find them easier in the mantis viewer.

**Design overview**
Current directory structure: (the masks are actually stored in fov subdirs in the mantis folder)
Screenshot 2023-09-07 at 1 35 31 PM

I think it would make more sense to have a higher level mantis directory that stores a copy of the segmentation masks, channel images, and then adds any masks that get generated.
Screenshot 2023-09-07 at 1 33 58 PM

We have two options with the generic cell notebook output:
1. Keep it within the pixie/cell_output_dir_generic subdirectory, since this will be the only mantis directory.
2. Also store the output in the top level mantis directory, since that’s the only mask that will be generated. (Unless the user runs both pixie and generic cell clustering which would be weird?)

**Code mockup**

The `create_mantis_dir()` function will need to be tweaked so that segmentation masks are not always copied over. The function already has a check to make sure channel images are no recopied, but this is not utilized since each notebook has it's own directory.

For the post_clusering output, we'll just need to change the name of the mask that's outputted from cell_mask.tiff to post_clustering_cell_mask.tiff. 
- Since `create_mantis_dir()` is called by `create_mantis_project()` in notebook 4, we can add a new arg to that function to specify the name of the mask to re-generate.

**Required inputs**

Channel images, segmentation masks, and output masks from whichever notebook is being run.

**Output files**

A single mantis directory that contains descriptive names for any masks added by the notebooks.

**Timeline**
- [x] A couple days
- [ ] A week
- [ ] Multiple weeks. For large projects, make sure to agree on a plan that isn't just a single monster PR at the end. 

Monday 9/11 

Wednesday 9/13

camisowers:

Closes #669 and closes #1059.  Creates a single top level mantis directory for notebooks 2, 3, and 4, to avoid duplicate files. 

- Set the mantis folder to be created as one subdirectory within the base_dir, rather than within each specific notebook folder.
- Change the output masks from notebook 4 to be named with `_post_clustering_cell_mask` instead of just `_cell_mask` to avoid overwriting notebook 3 cell cluster output files.
- Change the output masks from the generic_cell_clustering notebook to be named `_generic_cell_mask`.

**Remaining issues**

- [x] Run mantis generation in all clustering notebooks

### DIFF
--- a/src/ark/phenotyping/post_cluster_utils.py
+++ b/src/ark/phenotyping/post_cluster_utils.py
@@ -150,7 +150,7 @@ def create_mantis_project(
         )
         # save the cell mask for each FOV -- (saves with ".tiff" extension)
         data_utils.save_fov_mask(
-            fov, mask_dir, mask_data, sub_dir=None, name_suffix="_cell_mask"
+            fov, mask_dir, mask_data, sub_dir=None, name_suffix="_post_clustering_cell_mask"
         )
 
     # rename the columns of small_table
@@ -168,7 +168,7 @@ def create_mantis_project(
         mantis_project_path=mantis_dir,
         img_data_path=image_dir,
         mask_output_dir=mask_dir,
-        mask_suffix="_cell_mask",
+        mask_suffix="_post_clustering_cell_mask",
         mapping=mantis_df,
         seg_dir=seg_dir,
         cluster_type=cluster_type,
--- a/src/ark/utils/plot_utils.py
+++ b/src/ark/utils/plot_utils.py
@@ -508,7 +508,6 @@ def create_mantis_dir(fovs: List[str], mantis_project_path: Union[str, pathlib.P
                       seg_dir: Optional[Union[str, pathlib.Path]],
                       cluster_type='pixel',
                       mask_suffix: str = "_mask",
-                      mask_prefix: str = None,
                       seg_suffix_name: Optional[str] = "_whole_cell.tiff",
                       img_sub_folder: str = None,
                       new_mask_suffix: str = None):
@@ -555,8 +554,6 @@ def create_mantis_dir(fovs: List[str], mantis_project_path: Union[str, pathlib.P
             the type of clustering being done
         mask_suffix (str, optional):
             The suffix used to find the mask tiffs. Defaults to "_mask".
-        mask_prefix (str, optional):
-            The prefix used to find the mask tiffs. Defaults to None".
         seg_suffix_name (str, optional):
             The suffix of the segmentation file and it's file extension. If None, then
             the segmentation file will not be copied over.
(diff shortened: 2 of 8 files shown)


srivarra (approved):
Looks good!

cliu72 (changes requested):
Just a small comment - I think I confused you/I was also confused when I commented on the issue about the naming thing.

I don't think `mask_prefix` is actually needed. The masks are generated using `generate_and_save_pixel_cluster_masks`, which doesn't even have an option for a prefix, only a suffix. So the way the code is, there would be no situation where the masks have a prefix. I was originally thinking we could add the prefix when the masks are copied over into the mantis directory, however, after looking at this code, I don't think it's actually needed - I think `new_mask_suffix` is enough. So like users who want to can use the `new_mask_suffix`, can make it like `population_pixel_mask_whatevertheywant.tiff` in the mantis folder. So like I think we can jut remove `mask_prefix` and leave everything else as is.

alex-l-kong (commented):
Nothing to add on top of Candace's comments.

ngreenwald:
I think we should have a prefix so that all the masks are next to one another in alphabetical order when opening files in Mantis. Makes it easier to find the one you're looking for

cliu72:
> I think we should have a prefix so that all the masks are next to one another in alphabetical order when opening files in Mantis. Makes it easier to find the one you're looking for

They all start with "population_" then with the suffix after - which will make them all easy to find I think? Since you're opening one FOV in mantis at a time anyways, you can just look for all the files that start with "population"

camisowers:
> I think we should have a prefix so that all the masks are next to one another in alphabetical order when opening files in Mantis. Makes it easier to find the one you're looking for

Currently they all have a built in prefix of "population". So masks by default are `population_pixel_mask.tiff`, `population_cell_mask.tiff`, etc. We can change that to be something higher alphabetically if that would be more convenient.

ngreenwald:
Okay perfect

ngreenwald (approved):

PR #1069, opened 2023-09-28
Allow user to dynamically set batch size of pixel SOM assignments

alex-l-kong:

The current hard-coded batch size of 100 makes pixel SOM cluster assignment extremely slow. For a powerful machine (and even modern-day M1's and M2's), it can handle a much larger batch size and offer a 100x+ speedup.

Pass an argument through the PixelSOMCluster process called `num_parallel_pixels`, which controls how many SOM clusters get assigned at once.


src/ark/phenotyping/cluster_helpers.py
```diff
@@ -140,12 +147,12 @@ def generate_som_clusters(self, external_data: pd.DataFrame) -> np.ndarray:
 
         # work in batches of 100 to account to support large dataframe sizes
```
cliu72, line 148: A verrrrry small comment - but it'd be good to update this comment here so it doesn't say "batches of 100"

cliu72 (changes requested):
Sorry for the late review! Just some small requests.

cliu72 (changes requested):
Can we add the same functionality to `assign_som_clusters` in the `CellSOMCluster` class in cluster_helpers.py (and cell_som_clustering.py respectively)?

alex-l-kong:
> Can we add the same functionality to `assign_som_clusters` in the `CellSOMCluster` class in cluster_helpers.py (and cell_som_clustering.py respectively)?

Sure, same batch size work?

cliu72:
> > Can we add the same functionality to `assign_som_clusters` in the `CellSOMCluster` class in cluster_helpers.py (and cell_som_clustering.py respectively)?
> 
> Sure, same batch size work?

Yup, that works.

cliu72 (approved):
Looks good!

PR #1083, opened 2023-11-07
Fix cell cluster visualization in Mantis

### LINKED ISSUE #1075: Erosion and size of `plot_pixel_cell_cluster` output
In Pixie notebooks 2 and 3, the output of `plot_pixel_cell_cluster` is too large. It is annoying to manually shrink them, especially when looking at many FOVs. See screenshots for size.

Erosion should also be set to False for the pixel clustering notebook. It looks weird to have the edges eroded in the pixel cluster overlays. Erosion looks fine for the cell clustering notebook.

Smaller cluster overlays for both pixel and cell clustering. No erosion for pixel cluster overlays.

NA

**Additional context**
image
image

### LINKED ISSUE #1082: Incorrect cell cluster mapping in mantis

**Describe the bug**
Currently, the `cell_cluster_mask.csv` that is generated and loaded into mantis does not contain the correct integer IDs for each cell cluster. 

**Expected behavior**
When we updated the way we generate cell_masks to allow for colored masks, the integer IDs for the clusters changed.
If before they were [1,4,10] they are now [1,2,3]. The currently saved `cell_meta_cluster_mapping.csv` does not reflect this change.

**To Reproduce**
Run cell clustering and create new metaclusters.

camisowers:

1. Closes #1075.
Make minor tweaks to Pixie notebooks regarding plotting output. 

2. Closes #1082.
Ensure that the correct integer IDs are mapped to each cell cluster name in Mantis viewer.

1. The `erode` parameter in ` plot_utils.plot_pixel_cell_cluster()` is defaulted to True. Since it is also explicitly set to True in both the pixie cell clustering and generic cell clustering notebooks, we can just set the default to False to remove it from the pixel clustering output.

2. Adjust the `data_utils.generate_and_save_cell_cluster_masks()` function to update the cell_meta_cluster_mapping.csv at the same time it generates new integer IDs for the cell cluster masks. The new 'cluster_id' column is then read into `plot_utils.create_mantis_dir()` instead of the outdated 'cell_meta_cluster' values; this allows an accurate cell cluster mapping file to be generated and uploaded into mantis viewer.

**Remaining issues**

None. This bug had no effect on the cell table, since only the cluster name (not the integer ID) is stored there. Post-clustering is also not an issue, since a custom integer mapping is created in the `create_mantis_project()` function.


tests/utils/plot_utils_test.py
```diff
@@ -591,15 +591,18 @@ def test_create_mantis_dir(
...
 
-            # 3.a. Assert that metacluster col equals the region_id col
-            metacluster_col = original_mapping_df[[f"{cluster_type}_meta_cluster"]]
+            # 3.a. Assert that appropratiate cluster col equals the region_id col
+            cluster_col = "cluster_id" if cluster_type == "cell"\
+                else f"{cluster_type}_meta_cluster"
+            metacluster_col = original_mapping_df[[cluster_col]].drop_duplicates()
             region_id_col = new_mapping_df[["region_id"]]
-            metacluster_col.eq(region_id_col)
+            assert np.equal(metacluster_col.values, region_id_col.values).all()
```
srivarra, line 599: Got any preference on validating on the values stored in DataFrame vs the DataFrame as a whole (includes column names, dtypes, and indexes)?

srivarra (approved):
Looks good to me!

tests/utils/plot_utils_test.py
camisowers: These have different column names depending on whether it's the original mapping file or the new one generated for the mantis dir.

srivarra: oh okay, makes sense

templates/2_Pixie_Cluster_Pixels.ipynb
```diff
@@ -821,7 +821,8 @@
     "        pixel_cluster_mask,\n",
```
alex-l-kong, line 821: <p>Since you change the default <code>cluster_type</code> to  <code>'cell'</code>, you should explicitly pass <code>cluster_type='pixel'</code> here. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/ark-analysis/pull/1083/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='59'/>

alex-l-kong (approved):
Looks good

templates/2_Pixie_Cluster_Pixels.ipynb
camisowers: different function

src/ark/utils/plot_utils.py
```diff
@@ -740,15 +740,16 @@ def create_mantis_dir(fovs: List[str], mantis_project_path: Union[str, pathlib.P
     if not new_mask_suffix:
         new_mask_suffix = mask_suffix
 
-    map_df = map_df.loc[:, [f'{cluster_type}_meta_cluster', f'{cluster_type}_meta_cluster_rename']]
+    cluster_id_key = 'cluster_id' if cluster_type == 'cell' else f'{cluster_type}_meta_cluster'
```
cliu72, line 743: It's not a big deal at all, but `cluster_type` can only be `pixel` or `cell`. So I think it'd be a tad more clear if this was:
`cluster_id_key = 'cluster_id' if cluster_type == 'cell' else 'pixel_meta_cluster`

cliu72 (changes requested):
Just a few small things:
- The size of the actual plots look much better, but the font is now weirdly big. Are you seeing this on your machine too?
image
- Maybe this doesn't belong here (and could be it's own issue), but I noticed it while I was reviewing this. `post_cluster_utils.create_mantis_project` isn't ever used for pixel clustering (seems like it's only used in 4_Post_Clustering.ipynb to look at cell clusters). I think it's confusing to have `cluster_type` as a parameter (since it will never be set to `pixel`). Actually, I don't think it would work if `cluster_type` were ever set to pixel, because `"pop_vals"` would be changed to `cluster_id` (line 159), which is not the right nomenclature for the pixel clustering table. I think it'd be better to just remove `cluster_type` at all and assume this function will only be used for cells. Also in the `create_mantis_project` function, I don't understand why `small_table` is needed (line 119), since it's going to be fed into `ClusterMaskData`, which is able to be created directly from the cell table.

camisowers:
3x3 looks fine on my end, but will you try 4x4 and see if that's better? Could be a windows thing, but if it helps we can just change it.

Screenshot 2023-11-30 at 6 55 23 PM

I agree we can just get rid of the `cluster_type` arg from `create_mantis_project()` (I think it was just a copy over from `create_mantis_dir()`) an dis unnecessary.

Also, I think you're right about the small table being redundant, I'll change it and check it's everything is same.

cliu72 (commented):
Changes look good to me! I was actually using a Linux machine to look at the image before, but I just tried on my windows machine and it also looks weird. I wonder if it's a screen resolution thing.

Can you try if dpi=150 and figsize=(4,4) looks okay on your machine? If it looks like not too crazy, maybe we can set that as the default. That looks good for both Linux and Windows for me.

Also, I think it would be good to add dpi as a parameter in the notebook so people easily know they change it, and also add text telling them to change these values if it looks weird. So like to the pixel clustering notebook, after "Load a subset of the pixel cluster masks that you would like to preview." add "If the dimensions or text size of the plot need to be adjusted for optimal viewing, change the `dpi` and `figsize` parameters." And same thing to the cell clustering notebook.

camisowers:
@cliu72 5x5 with dpi=150 looks great on my end! Once you approve I can merge it in.

PR #1102, opened 2024-01-05
Fix issue with notebook test generators

alex-l-kong:

The generators don't properly get torn down for the notebook tests, which causes them to attempt to delete non-existent folders.

Investigate the way these tests are torn down and address accordingly.


alex-l-kong:
The issues in this PR were addressed by #1114.

Issue #1103, opened 2024-01-09
marker_quantification.process_lists() returning inappropriate mask_types


bryjcannon:
Describe the bug
![Screenshot 2024-01-08 at 4 58 41 PM](https://github.com/angelolab/ark-analysis/assets/18650480/bd332376-5a46-4786-83d9-73378d57be18)

When trying to run segmentation, an error propped up where a mask file that didn't exist was being loaded for csv writing. After troubleshooting, the problem occurs in `marker_quantification.process_lists` code. In this example, while only file names with 'R6C1' were desired, files with R6C10' were also being extracted due to the way `marker_quantification.process_lists` matches substrings in file names. Similar to list_files issue in alpineer.

Expected behavior
`marker_quantification.process_lists` currently finds files using the code below based on a matching file prefix. Instead of returning "whole_cell" and "nuclear", the returned list is "_whole_cell", "0_whole_cell", etc.

To Reproduce
Ping me for access to my mask files.

Proposed solution:
Similar to the list_files and list_folders fix in alpineer, token matching will be used instead of string subsetting to find matching files (code pic below)
![Screenshot 2024-01-08 at 4 59 03 PM](https://github.com/angelolab/ark-analysis/assets/18650480/fd57fa9d-3059-4456-9d56-752147b49854)

PR #1101, opened 2024-01-05
1100 empty composites masks in ez segmenter notebook

bryjcannon:
**If you haven't already, please read through our [contributing guidelines](https://ark-analysis.readthedocs.io/en/latest/_rtd/contributing.html) before opening your PR**

**What is the purpose of this PR?**
Fixes issue in #1100 where float images were not able to be used in composite, mask creation.

**How did you implement your changes**

Image loading was converted from np.int32 to np.float32, plus additional handling of composite array addition and subtraction functions.

**Remaining issues**

None as of now, may run into issues if non float or int img datatypes are used.

PR #1104, opened 2024-01-09
Changes process lists search for matching fov and mask names to use a…

bryjcannon:
Addresses Issue #1103


bryjcannon:
Same example_dataset fails as in #1101. All other tests pass.

alex-l-kong:
@bryjcannon added a simple test for `process_lists`, confirmed that it matches the correct FOVs to masks even in the case of `fov1` and `fov10`.

alex-l-kong (approved):
Looks good.

PR #1107, opened 2024-01-18
Only load whole cell masks if `nuclear_counts=False` in `generate_cell_table`

### LINKED ISSUE #1097: Even when setting `nuclear_counts = False`, nuclear masks are being quantified in `marker_quantification.generate_cell_table`
**Describe the bug**
After setting `nuclear_counts = False` in `marker_quantification.generate_cell_table`, the nuclear masks are still being quantified and added to the cell table. This leads to unexpected behavior in the pixie cell clustering notebook, since now in the cell table, there are two rows for every fov/label pair (one for whole cell, one for nuclear). Because there are two different cell sizes associated with each cell label, this leads to errors. (Of course, the easy option is to subset the cell table for rows only where mask_type = whole_cell before running cell clustering, but it's not obvious for users that this is necessary, since `nuclear_counts` was set to False in the segmentation notebook, so the expectation is that nuclear masks are not quantified in the cell table)

I think the problem is here (https://github.com/angelolab/ark-analysis/blob/main/src/ark/segmentation/marker_quantification.py#L532-L585):
```
       mask_files = io_utils.list_files(segmentation_dir, substrs=fov_name)
       mask_types = process_lists(fov_names=fovs, mask_names=mask_files)
```

`mask_types` here is `['nuclear','whole_cell']`. Therefore, both nuclear and whole cell segmentation masks are always being read (regardless of what `nuclear_counts` is set to).

**Expected behavior**
If `nuclear_counts=False`, do not read in nuclear masks in `generate_cell_table` (and do not have rows for nuclear masks in the final cell table). I think the logic of `generate_cell_table` needs to change slightly to not read in nuclear masks if `nuclear_counts=False`.

**To Reproduce**
As I noted in another issue (https://github.com/angelolab/ark-analysis/issues/1096), there is a problem with running the segmentation notebook using the example dataset, but if you manually delete `fov10` from the example dataset, you should be able to run-through the entire segmentation notebook. After doing this and running through notebook 1, you can open either cell table and see that the "mask_type" column has both 'nuclear' and 'whole_cell'. Then, you can reproduce the error in this issue by running notebooks 2 and 3 subsequently.

This is the error I get in notebook 3 after doing the above (length mismatch because there are extra rows in the cell table):
image

alex-l-kong:

Closes #1097. Updates `generate_cell_table` to be compatible with updated ezSeg pipeline when nuclear masks need to be discarded.

By default, the updated `generate_cell_table` loads in all the masks, including nuclear, and adds them to the cell table. For pixel clustering, only the whole cell data is needed, so if `nuclear_counts=False`, this mask type should not appear in the table.

**Remaining issues**

Integration testing with Pixie/Nimbus needed.

### DIFF
--- a/src/ark/segmentation/marker_quantification.py
+++ b/src/ark/segmentation/marker_quantification.py
@@ -529,9 +529,14 @@ def generate_cell_table(segmentation_dir, tiff_dir, img_sub_folder="TIFs",
         whole_cell_file = fov_name + '_whole_cell.tiff'
         nuclear_file = fov_name + '_nuclear.tiff'
 
-        # for each label given in the argument. read in that mask for the fov, and proceed with label and table appending
+        # for each label given in the argument, read in that mask for the fov, and proceed with
+        # label and table appending
         mask_files = io_utils.list_files(segmentation_dir, substrs=fov_name)
         mask_types = process_lists(fov_names=fovs, mask_names=mask_files)
+
+        # remove nuclear from mask_types if nuclear_counts False
+        if not nuclear_counts and "nuclear" in mask_types:
+            mask_types.remove("nuclear")
 
         for mask_type in mask_types:
             # load the segmentation labels in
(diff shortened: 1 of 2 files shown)


cliu72 (approved):
Looks good to me

PR #1110, opened 2024-01-22
Don't generate duplicate `cluster_id` columns if `generate_and_save_cluster_masks` run twice

### LINKED ISSUE #1099: Error in `create_mantis_dir` in cell clustering notebook if running `generate_and_save_cell_cluster_masks` multiple times
**Describe the bug**
This is a pretty niche problem, but hopefully this fix is straight forward. When I run the "4.4: Generate cell phenotype maps" section of the cell clustering notebook more than once, I get this error: `KeyError: "['cluster_id'] not in index"`. The use case for this is if a user wants to look at ['fov1','fov2'] first and runs the rest of the notebook with `subset_cell_fovs = ['fov1','fov2']`, then decides they want to look at other fovs, so changes `subset_cell_fovs` and tries to run the rest of the notebook again.

I think this error is because in `generate_and_save_cell_cluster_masks`, a `cluster_id` column is added to the mapping csv. However, if this function is run multiple times, because the `cluster_id` column already exists, the columns are not merged properly: 
image

**Expected behavior**
Should be allowed to change `subset_cell_fovs` and re-run the rest of the notebook without issue.

**To Reproduce**
Run the example dataset in notebook 3 all the way through. Go back to the start of section 4.4, change `subset_cell_fovs` to something else (ex. ['fov3','fov4']) and run the rest of the notebook. After doing this, I got this error in the last cell:
image

alex-l-kong:

Closes #1099. Addresses an issue in the merging of `gui_map` and `cluster_map` in `generate_and_save_cluster_masks` where if an existing `cluster_id` already exists in `gui_map`, multiple `cluster_id` columns with variable suffixes get created.

We use the `.drop` method in `pandas` with `errors="ignore"` to drop the `cluster_id` column, but only if it exists. This will prevent merging issues.


camisowers (approved):
Looks good.

cliu72 (approved):
Looks good to me

srivarra (approved):
Looks good!

PR #1126, opened 2024-03-12
Manual mask names

### LINKED ISSUE #1125: Duplicate FOV name issues with cell table generation

**Describe the bug**
If the user has a naming convention of `fov1` and `fov1_duplicate`, this causes an issue in `generate_cell_table()` due to some ezseg updates. Specifically, the function `get_existing_mask_types()` is run per FOV to detect any masks that may exist for that image, so when processing fov1 it returns the "mask types": ['whole_cell', 'nuclear', 'duplicate_whole_cell', 'duplicate_nuclear']. 

After chatting with @bryjcannon, we think it would be best for the user to just specify which mask names to look for (defaulted to ['whole_cell'] for the cell segmentation notebook and an argument in the ezseg notebook) rather than automating something so simple.

**Expected behavior**
Be able to generate a cell table no matter any odd FOV naming conventions.

**To Reproduce**
Try to run the segmentation notebook with the specified FOV names.

camisowers:

Closes #1125.

1. Prevents duplicate names from causing issues while generating the cell table.
2. Theres currently also a bug in the nuclear expression logic.
Currently if `nuclear_counts=True`, the nuclear counts are calculated in a loop for each mask (large/unnecessary computation burden for many ezseg masks).
The cell table creates it's own set of rows for the nuclear masks according to the ezseg logic, while also having a second set of columns suffixed with _nuclear using traditional cell table logic.  We want to remove the duplicated nuclear expression values, as well as saving the computational effort.

1. Adds `mask_types` as an optional argument to `generate_cell_table()` detailing which masks to extract expression data for; the default is set to ['whole_cell'].
2. Ensure nuclear counts are only done once, when the whole cell masks are being processed.

**Remaining issues**

Add manual mask name specification into the ez segmenter notebook. @bryjcannon

### DIFF
--- a/src/ark/segmentation/marker_quantification.py
+++ b/src/ark/segmentation/marker_quantification.py
@@ -476,10 +476,10 @@ def generate_cell_table(segmentation_dir, tiff_dir, img_sub_folder="TIFs",
         nuclear_counts (bool):
             boolean flag to determine whether nuclear counts are returned, note that if
             set to True, the compartments coordinate in segmentation_labels must contain 'nuclear'
-        mask_types (list):
-            if set, skips the custom regionprops and expensive base regionprops extraction steps
         fast_extraction (bool):
             if set, skips the custom regionprops and expensive base regionprops extraction steps
+        mask_types (list):
+            list of masks to extract data for, defaults to ['whole_cell']
         **kwargs:
             arbitrary keyword arguments for signal and regionprops extraction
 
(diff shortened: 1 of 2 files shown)


srivarra (approved):
Looks good!!

bryjcannon:
Code looks good with, just need to update the example_dataset and can then merge with main.

srivarra:
@camisowers Dataset is updated, this is good to go.

PR #1127, opened 2024-03-19
Fix color masking in Pixie

### LINKED ISSUE #1117: Error in `plot_utils.save_colored_masks` and `plot.utils.plot_pixel_cell_cluster`
**Describe the bug**
Both `plot_utils.save_colored_masks` and `plot_utils.plot_pixel_cell_cluster` are not mapping the right color to the right cell population when the cluster IDs are not continuous. See the example below. Essentially, when there are breaks in the metacluster IDs (for example, numbers are skipped), the color mapping becomes incorrect.

I think the reason for this is because the wrong column is being read in `MetaclusterColormap`. Due to some recent refactoring, we are now adding a "cluster_id" column to `cell_meta_cluster_mapping.csv` (https://github.com/angelolab/ark-analysis/blob/main/src/ark/utils/data_utils.py#L444-L451), and these numbers are used to create the cell mask. However, the "cluster_id" column is being ignored in the `MetaclusterColormap` functions (https://github.com/angelolab/ark-analysis/blob/main/src/ark/utils/plot_utils.py#L42), so the colors are being mapped to the wrong numbers.

We are currently feeding in `raw_cmap` into `plot_utils.save_colored_masks` and `plot_utils.plot_pixel_cell_cluster` . I think it would be safer to refactor that function and feed `renamed_cmap` into that function, since `renamed_cmap` has the actual text labels for each cluster. I think that would be the best way to ensure the right colors are mapped to the right cells.

**Expected behavior**
Generate accurate color overlays.

**To Reproduce**
Run the example dataset through notebook 3. Highlight all of metacluster 2 and click "new metacluster" to change metacluster 2 to metacluster 21. If you then look at the colored overlays, you'll see that cells that were previously in metacluster 2 are NOT properly assigned to metacluster 21 (as they should be). In the screenshot below, you can see that metacluster 2 no longer exists in the bottom (which is correct), but all cells that were previously metacluster 2 (light red in the top example) should now be green (color of metacluster 21 - it is correct in the legend), but they are not correctly colored (instead of green, they are a light teal color).

Original clustering:
image
image

New clustering (the only change I made was change cluster 2 to cluster 21):
image
image

camisowers:

Closes #1117. Fixes colored masking for both pixel and cell clustering.

The problem seemed to arise from non-sequential meta_cluster ids. If you change meta cluster 2 -> 21, then the IDs would be [1,3,4, ..., 21], which causes an issue when creating a colormap that expects only 20 clusters (and therefore values up to 20). However, upon further investigation, most of the cell cluster colors were being visualized incorrectly.

To ensure we are always matching the clusters correctly no matter the meta_cluster ID assignments, we
1. Reassign regular (non-colored) mask values to be sequential ints and save it to the *_meta_cluster_mapping.csv for reference. Use `sort_values(by=['cell_meta_cluster'])` to ensure the cluster_id are generated by ascending meta cluster number every time for consistency.
   - The cell masks already did this, but it was not using the saved `cluster_id` values in *_meta_cluster_mapping.csv and was instead doing it's own (second) reassignment, which was why the mapping was wrong.
2. When generating colored masks, simply go into the saved file so we can grab the cluster_id that corresponds to the meta_cluster number, and combine that with the raw_cmap (e.g. {meta_cluster: color}), so we can know for sure which mask integer ID connects to each color.

**Old (no good) colored mask output:**
image

**New (correct) colored mask output:**
Figure 3
Verified by me hand checking the `label` and the provided `cell_meta_cluster_rename` in the cell table for a bunch of random cells in the image. 😅

**Remaining issues**

- [x] Check pixel clustering mantis mapping file.

### DIFF
--- a/src/ark/utils/data_utils.py
+++ b/src/ark/utils/data_utils.py
@@ -477,7 +477,8 @@ def generate_and_save_cell_cluster_masks(
 
 
 def generate_pixel_cluster_mask(fov, base_dir, tiff_dir, chan_file_path,
-                                pixel_data_dir, pixel_cluster_col='pixel_meta_cluster'):
+                                pixel_data_dir, cluster_mapping,
+                                pixel_cluster_col='pixel_meta_cluster'):
     """For a fov, create a mask labeling each pixel with their SOM or meta cluster label
 
     Args:
@@ -493,6 +494,8 @@ def generate_pixel_cluster_mask(fov, base_dir, tiff_dir, chan_file_path,
         pixel_data_dir (str):
             The path to the data with full pixel data.
             This data should also have the SOM and meta cluster labels appended.
+        cluster_mapping (pd.DataFrame)
+            Dataframe detailing which meta_cluster IDs map to which cluster_id
         pixel_cluster_col (str):
             Whether to assign SOM or meta clusters
             needs to be `'pixel_som_cluster'` or `'pixel_meta_cluster'`
@@ -542,11 +545,9 @@ def generate_pixel_cluster_mask(fov, base_dir, tiff_dir, chan_file_path,
     # get the corresponding cluster labels for each pixel
     cluster_labels = list(fov_data[pixel_cluster_col])
 
-    # relabel clusters with sequential integers (cluster_id)
-    unique_clusters = list(np.unique(cluster_labels))  # returns sorted meta cluster numbers
-    cluster_ids = list(range(1, len(unique_clusters) + 1))
-    id_mapping = {meta_cluster: cluster_id
-                  for meta_cluster, cluster_id in zip(unique_clusters, cluster_ids)}
+    # relabel meta_cluster numbers with cluster_id
+    cluster_mapping = cluster_mapping.drop_duplicates()[[pixel_cluster_col, 'cluster_id']]
+    id_mapping = dict(zip(cluster_mapping[pixel_cluster_col], cluster_mapping['cluster_id']))
     cluster_labels = [id_mapping[label] for label in cluster_labels]
 
     # assign each coordinate in pixel_cluster_mask to its respective cluster label
@@ -626,7 +627,8 @@ def generate_and_save_pixel_cluster_masks(fovs: List[str],
                 generate_pixel_cluster_mask(fov=fov, base_dir=base_dir, tiff_dir=tiff_dir,
                                             chan_file_path=chan_file_path,
                                             pixel_data_dir=pixel_data_dir,
-                                            pixel_cluster_col=pixel_cluster_col)
+                                            pixel_cluster_col=pixel_cluster_col,
+                                            cluster_mapping=updated_cluster_map)
 
             # save the pixel mask generated
             save_fov_mask(fov, data_dir=save_dir, mask_data=pixel_mask, sub_dir=sub_dir,
--- a/src/ark/utils/data_utils.py
+++ b/src/ark/utils/data_utils.py
@@ -19,11 +19,10 @@ import xarray as xr
 from ark import settings
 from skimage.segmentation import find_boundaries
 import dask.dataframe as dd
-from dask import delayed
+from pandas.core.groupby.generic import DataFrameGroupBy
 from anndata import AnnData, read_zarr
 from anndata.experimental import AnnCollection
 from anndata.experimental.multi_files._anncollection import ConvertType
-from tqdm.dask import TqdmCallback
 from torchdata.datapipes.iter import IterDataPipe
 from typing import Iterator, Optional
 try:
@@ -849,14 +848,13 @@ def stitch_images_by_shape(data_dir, stitched_dir, img_sub_folder=None, channels
                                current_img)
 
 
-@delayed
-def _convert_ct_fov_to_adata(fov_dd: dd.DataFrame, var_names: list[str], obs_names: list[str], save_dir: os.PathLike) -> str:
+def _convert_ct_fov_to_adata(fov_group: DataFrameGroupBy, var_names: list[str], obs_names: list[str], save_dir: os.PathLike) -> str:
     """Converts the cell table for a single FOV to an `AnnData` object and saves it to disk as a
     `Zarr` store.
 
     Parameters
     ----------
-    fov_dd : dd.DataFrame
+    fov_group : DataFrameGroupBy
         The cell table subset on a single FOV.
     var_names: list[str]
         The marker names to extract from the cell table.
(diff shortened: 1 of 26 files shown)


src/ark/utils/plot_utils.py
```diff
@@ -147,14 +144,30 @@ def _metacluster_cmap_generator(self) -> None:
...
+        cluster_id_to_metacluster_map = cluster_id_to_name[
+            [f"{self.cluster_type}_meta_cluster", "cluster_id"]].drop_duplicates()
+        unassigned_cluster_id: int = int(cluster_id_to_name["cluster_id"].max() + 1)
+        cluster_id_to_metacluster_map = pd.concat(
+            [
+                cluster_id_to_metacluster_map,
+                pd.DataFrame(
+                    data={
+                        f"{self.cluster_type}_meta_cluster": [unassigned_cluster_id, 0],
+                        "cluster_id": [unassigned_id, 0],
```
cliu72, line 157: I think `unassgined_cluster_id` and `unassigned_id` are flipped? Should it be:
```
data={
                        f"{self.cluster_type}_meta_cluster": [unassigned_id, 0],
                        "cluster_id": [unassigned_cluster_id, 0],
}
```

Since above, `unassigned_cluster_id: int = int(cluster_id_to_name["cluster_id"].max() + 1)`, so `unassigned_cluster_id` shoud correspond to the `cluster_id` column.

(but per my comment below, this might not be necessary at all)

src/ark/utils/data_utils.py
```diff
@@ -537,9 +539,16 @@ def generate_pixel_cluster_mask(fov, base_dir, tiff_dir, chan_file_path,
...
+    # get the corresponding cluster labels for each pixel
     cluster_labels = list(fov_data[pixel_cluster_col])
 
+    # relabel clusters with sequential integers (cluster_id)
+    unique_clusters = list(np.unique(cluster_labels))  # returns sorted meta cluster numbers
+    cluster_ids = list(range(1, len(unique_clusters) + 1))
+    id_mapping = {meta_cluster: cluster_id
+                  for meta_cluster, cluster_id in zip(unique_clusters, cluster_ids)}
+    cluster_labels = [id_mapping[label] for label in cluster_labels]
+
```
cliu72, line 551: When I search for this function `generate_pixel_cluster_mask` in Github, I can't tell where in the pipeline it's used (search results only return the function definition and the test file), but I imagine that this implementation could cause some downstream problems if not every FOV contains every cluster. For example, if FOV1 contains clusters 1,3,5 - then those will be mapped to 1,2,3. However, if FOV2 contains clusters 1,2,5 - then those will also be mapped to 1,2,3. This could be confusing downstream because it seems like FOV1 and FOV2 contain the same clusters when they don't.

src/ark/utils/plot_utils.py
```diff
@@ -99,6 +98,7 @@ def _metacluster_cmap_generator(self) -> None:
                 f"{self.cluster_type}_som_cluster",
                 f"{self.cluster_type}_meta_cluster",
                 f"{self.cluster_type}_meta_cluster_rename",
+                f"cluster_id",
```
cliu72, line 101: Not important at all, but it'd be good to update the comment in line 95 with these updated column names.

src/ark/utils/plot_utils.py
```diff
@@ -147,14 +144,30 @@ def _metacluster_cmap_generator(self) -> None:
...
+        # sort by cluster_id ascending, so colors align with mask integers
+        cluster_id_to_color.sort_values(by="cluster_id", inplace=True)
+        cluster_id_to_color.reset_index(drop=True, inplace=True)
 
-        metacluster_to_index = {}
-        metacluster_id_to_name.reset_index(drop=True, inplace=True)
-        for index, row in metacluster_id_to_name.reset_index(drop=True).iterrows():
-            metacluster_to_index[row[f'{self.cluster_type}_meta_cluster']] = index
+        # Convert the list of tuples to a numpy array, each index is a color
+        mc_colors: np.ndarray = np.array(cluster_id_to_color['color'].to_list())
```
cliu72, line 170: I think everything above this is a little bit needlessly confusing. It seems like all of this code is to first map metacluster to color, and then metacluster to cluster_id, then cluster_id to color while adding an unassigned id twice. I think it can be simplified to add the unassigned id only once.

I think this is much simpler, starting from line 106 (sorry didn't have time to test it so might have issues, but this is the gist of what I'm thinking, let me know if there's a flaw my logic):

```
        # subset on metacluster, metacluster name, and cluster id
        metacluster_id_to_name = cluster_id_to_name[
            [f"{self.cluster_type}_meta_cluster", f"{self.cluster_type}_meta_cluster_rename","cluster_id"]
        ].copy()
        unassigned_meta_cluster: int = int(
            metacluster_id_to_name[f"{self.cluster_type}_meta_cluster"].max() + 1)
        unassigned_cluster_id: int = int(
            metacluster_id_to_name["cluster_id"].max() + 1)

        # Extract unique (metacluster-ID,  name, cluster_id)
        # Set the unassigned meta cluster to be the max ID + 1
        # Set 0 as the Empty value
        metacluster_id_to_name: pd.DataFrame = pd.concat(
            [
                metacluster_id_to_name.drop_duplicates(),
                pd.DataFrame(
                    data={
                        f"{self.cluster_type}_meta_cluster": [unassigned_meta_cluster, 0],
                        f"{self.cluster_type}_meta_cluster_rename": ["Unassigned", "Empty"],
                        "cluster_id":[unassigned_cluster_id, 0]
                    }
                )
            ]
        )

        # add the unassigned color to the metacluster_colors dict
        self.metacluster_colors.update({unassigned_meta_cluster: self.unassigned_color})

        # add the no cluster color to the metacluster_colors dict
        self.metacluster_colors.update({0: self.background_color})

        # assert the metacluster index in colors matches with the ids in metacluster_id_to_name
        misc_utils.verify_same_elements(
            metacluster_colors_ids=list(
                self.metacluster_colors.keys()),
            metacluster_mapping_ids=metacluster_id_to_name
            [f'{self.cluster_type}_meta_cluster'].values)
        # use metacluster_colors to add the colors to metacluster_id_to_name
        metacluster_id_to_name["color"] = metacluster_id_to_name[
            f"{self.cluster_type}_meta_cluster"
        ].map(self.metacluster_colors)

        # sort by cluster_id ascending, so colors align with mask integers
        metacluster_id_to_name.sort_values(by="cluster_id", inplace=True)
        metacluster_id_to_name.reset_index(drop=True, inplace=True)

        # Convert the list of tuples to a numpy array, each index is a color
        mc_colors: np.ndarray = np.array(metacluster_id_to_name['color'].to_list())
```

cliu72 (changes requested):
A few suggestions!

src/ark/utils/plot_utils.py
camisowers: Yeah this makes sense to me and should work! I was trying to avoid pd.merge() issues with the two dfs but just adding cluster_id to the original metacluster_id_to_name totally solves that!

src/ark/utils/data_utils.py
camisowers: It's used in the same data_utils files in `generate_and_save_pixel_cluster_masks()`.
But yes you're right this could definitely cause issues. I can instead pass the mapping saved to cluster_id_name_path in the wrapper function since that is generated directly from the mapping file which contains every cluster!

cliu72: Oh duh I'm dumb lol

cliu72 (approved):
Looks good to me!

alex-l-kong (approved):
Looks good

srivarra (approved):
Wild stuff!

PR #1114, opened 2024-01-29
Example Dataset valid datasets fix

srivarra:
**If you haven't already, please read through our [contributing guidelines](https://ark-analysis.readthedocs.io/en/latest/_rtd/contributing.html) before opening your PR**

**What is the purpose of this PR?**

Fixes the example dataset issue in CI.

**How did you implement your changes**

Added `"ez_seg_data"` to the list of valid datasets in `.github/scripts/get_example_dataset.py` script.

**Remaining issues**

N/A

PR #1152, opened 2024-08-29
Enforce consistent channel ordering in Pixie

alex-l-kong:

Users may not always specify channels in strictly alphabetical order when running pixel-level Pixie. Additionally, Python's definition of sorting differs significantly from how we normally think of sorting, especially with capitalization. To prevent confusion, we need a consistent, non-stochastic ordering of channel names.

Add a line to call Python's `sorted` function after defining the `channels` list. This ensures `channels` adheres to Python's sorting conventions.

**Remaining issues**

Not sure if cross-platform compatibility will be an issue, but `sorted` should be OS-agnostic.

### DIFF
--- a/templates/2_Pixie_Cluster_Pixels.ipynb
+++ b/templates/2_Pixie_Cluster_Pixels.ipynb
@@ -336,7 +336,10 @@
     "            \"CD163_nuc_exclude\", \"CK17\", \"Collagen1\", \"Fibronectin\", \n",
     "            \"ECAD_smoothed\", \"HLADR\", \"SMA\", \"Vim\"]\n",
     "blur_factor = 2\n",
-    "subset_proportion = 0.1"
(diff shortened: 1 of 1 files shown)


cliu72:
@alex-l-kong I thought this was already addressed in this PR: https://github.com/angelolab/ark-analysis/pull/1114 (See point that says "Adds natsort calls throughout the Pixie pipeline to avoid issues with channel ordering." under "Pixel clustering"). Does this PR address something different?

alex-l-kong:
> @alex-l-kong I thought this was already addressed in this PR: #1114 (See point that says "Adds natsort calls throughout the Pixie pipeline to avoid issues with channel ordering." under "Pixel clustering"). Does this PR address something different?

Let me check with Avery again. It looked like he had updated the repo but he was still running into this issue, so I added this as an additional safeguard.

Issue #1150, opened 2024-08-07
Can't run template notebooks "ModuleNotFoundError: No module named 'torch.utils._import_utils'"


givison:
Hi Angelo folks!

Trying to get started running the analysis notebooks. Followed the install directions in the readme, tried cloning & using conda and installing from pip and getting the same error when I try to load the libraries at the top of both notebook 1 & notebook 2:

---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], line 12
      8 from matplotlib import rc_file_defaults
     10 from ark.phenotyping import (pixel_cluster_utils, pixel_meta_clustering,
     11                              pixel_som_clustering, pixie_preprocessing)
---> 12 from ark.utils import data_utils, example_dataset, plot_utils
     13 from ark.utils.metacluster_remap_gui import (MetaClusterGui,
     14                                              colormap_helper,
     15                                              metaclusterdata_from_files)

File ~/miniconda3/envs/ark_env/lib/python3.10/site-packages/ark/utils/data_utils.py:25
     23 from anndata.experimental import AnnCollection
     24 from anndata.experimental.multi_files._anncollection import ConvertType
---> 25 from torchdata.datapipes.iter import IterDataPipe
     26 from typing import Iterator, Optional
     27 try:

File ~/miniconda3/envs/ark_env/lib/python3.10/site-packages/torchdata/datapipes/__init__.py:11
      7 from torch.utils.data import DataChunk, functional_datapipe
      9 from torchdata import _extension  # noqa: F401
---> 11 from . import iter, map, utils
     13 __all__ = ["DataChunk", "functional_datapipe", "iter", "map", "utils"]
     16 from torchdata import deprecation_warning

File ~/miniconda3/envs/ark_env/lib/python3.10/site-packages/torchdata/datapipes/iter/__init__.py:79
     68 from torchdata.datapipes.iter.transform.callable import (
     69     BatchAsyncMapperIterDataPipe as BatchAsyncMapper,
     70     BatchMapperIterDataPipe as BatchMapper,
   (...)
     76     ThreadPoolMapperIterDataPipe as ThreadPoolMapper,
     77 )
     78 from torchdata.datapipes.iter.util.bz2fileloader import Bz2FileLoaderIterDataPipe as Bz2FileLoader
---> 79 from torchdata.datapipes.iter.util.cacheholder import (
[...]

PR #1158, opened 2024-09-20
Remove anything Docker related from the README

### LINKED ISSUE #1145: Windows Problem With bash start_docker.sh Command
Hello!

I am attempting to install this package today to use Pixie on a Windows machine. I have made it through the installation tutorial up to the step "Setting up the repo". 

Everything seems to be working well until I execute the line `bash start_docker.sh` in the command prompt terminal. The outcome is the error message in the image below. I ran `wsl sudo apt-get install dos2unix` and then retried, but this did not work for me. 

Please let me know if there is a solution to this problem. I look forward to your response and thank you in advance!

image

alex-l-kong:

Closes #1145. Due to issues users have had trying the Docker setup, it's best for the general public not to use this setup.

Remove anything from the `README.md` about Docker setup. Instead, provide detailed instructions for using `conda` environments and `./start_jupyter.sh`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Full documentation for the project can be found [here](https://ark-analysis.read
     - [Download the Repo](#download-the-repo)
     - [Running on Windows](#running-on-windows)
     - [Using the Repository](#using-the-repository)
+    - [Updating the repo](#updating-the-repo)
 - [External Tools](#external-tools)
   - [Mantis Viewer](#mantis-viewer)
     - [Mantis directory structure](#mantis-directory-structure)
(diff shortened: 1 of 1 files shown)


alex-l-kong:
@cliu72 the failing tests are due to the ongoing pytorch issue (https://github.com/angelolab/ark-analysis/issues/1150)

camisowers (changes requested):
Since the conda env will be the way we support ark from now on, could use a little more info on how to maintain it like we have written up in toffy. For example, an [updating the repo](https://github.com/angelolab/toffy?tab=readme-ov-file#updating-the-repo) section is probably necessary.

srivarra (approved):

srivarra (changes requested):
@alex-l-kong I think we can also remove all docker functionality in this PR such as the [Dockerfile](https://github.com/angelolab/ark-analysis/blob/301e344f0ffbedafaad282c52c801e7713272f85/Dockerfile), and the [docker.yml](https://github.com/angelolab/ark-analysis/blob/301e344f0ffbedafaad282c52c801e7713272f85/.github/workflows/docker.yml) workflow. What do you think, or we can toss that into another issue?

alex-l-kong:
> @alex-l-kong I think we can also remove all docker functionality in this PR such as the [Dockerfile](https://github.com/angelolab/ark-analysis/blob/301e344f0ffbedafaad282c52c801e7713272f85/Dockerfile), and the [docker.yml](https://github.com/angelolab/ark-analysis/blob/301e344f0ffbedafaad282c52c801e7713272f85/.github/workflows/docker.yml) workflow. What do you think, or we can toss that into another issue?

@srivarra I can remove the Docker badge from this PR. The complete removal of Docker I wanted to save for a different PR. Mostly because I think it'll still be good to have around as legacy. Main goal is to keep this hidden from the instruction manual as much as possible so it doesn't confuse users.

srivarra:
@alex-l-kong Sounds good to me!

camisowers (approved):
Nice

srivarra (approved):

PR #1164, opened 2024-10-07
Adjust mask suffix input to generate_cell_table

camisowers:

Allow for users to generate a cell table for segmentation masks that have slightly different formatting for file names.  For example, simply 'fov1.tiff' or 'fov1-cell_mask.tiff' instead of always requiring _suffix (i.e. 'fov1_whole_cell.tiff').

Add in an extra argument `add_underscore` which is defaulted to True. Also allow None as a mask_type input. These adjustments will not be used internally in ark-analysis, but are intended to allow for more flexibility of segmentation masks generated with methods other than Mesmer and fed into [SpaceCat](https://github.com/angelolab/SpaceCat?tab=readme-ov-file).

**Remaining issues**

Need to make a new release so the changes can be used in SpaceCat.


srivarra (approved):
Looks good to me!

srivarra:
@camisowers the example dataset is fixed up.

alex-l-kong (approved):
Looks great, this will also be useful if we/DeepCell decides to ever change the suffix of the segmentation files for whole cell and nuclear.

PR #1183, opened 2025-03-11
Ensure `_temp` dir does not interfere with pixel SOM and consensus cluster assignment

alex-l-kong:

If a run is restarted from the beginning, or if a `_temp` dir still exists due to a crash, this creates a strange limbo state where the pipeline:

1. Will attempt to create an already existing `_temp` folder if an `overwrite` is forced (which is often the recommended solution in case the run is completely unsalvageable)
2. Cannot identify what all the seen SOM clusters are if all FOVs have already been assigned (because it simply "passes through" the assignment step, where this normally happens)

To address these issues:

1. If a `_temp` directory is seen on `overwrite=True`, delete it first
2. If the `pixel_pysom` object contains no SOM references in `som_clusters_seen`, invokes a new function, `pixel_cluster_utils.identify_seen_pixel_clusters`, to populate this. The function does extend support to meta clusters as well, though this case is not directly invoked in the pipeline.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,12 +62,15 @@ dependencies = [
     "zstandard>=0.19.0,<1",
     "zarr",
     "python-dateutil>=2.9.0",
-    "ark-analysis[colors]",
+    "palettable",
+    "cmasher",
+    "cmocean",
+    "colorcet",
+    "scienceplots",
 ]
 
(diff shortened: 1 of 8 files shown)


camisowers (approved):
Looks good! Just need to fix the builds.

